PR #3, opened 2019-07-27
Adjustable menu icons

Allcharles:
## Update Contents
Updated navigation panes to utilise a more standardised menu config.

## New Files
A new series of files was created to allow for this:

- `<component>.component.menus.ts`: Defines the component icon; and menu items if required. Icon is utilised by header and secondary menu links.
- `<component>/menus.service.ts`: Implements changes from `<component>.component.menus.ts` file.
- `<collection>.menus.ts`: Defines the collection action icon and title used by the action menu for each of the sub components.
- `<collection>.components.ts`: Defines the components inside the collection to reduce `app.module.ts` imports.
- `<collection>.routes.ts`: Defines the routes for the collections childrens to give better control over `app-routing.module.ts`.

### Notes
- component = Low level component
- collection = Collection of related low level components
- Changed menu.json files to menu.ts to support interface types and the use of functions. This move was required to support the loading of icons from the component.

## Future Work
Next update will involve changing the action and secondary menus to router-outlets. However as this may affect a large number of files, this intermediary step was taken.


src/app/app-routing.module.ts
```diff
@@ -3,58 +3,33 @@ import { Routes, RouterModule } from '@angular/router';
...
-    path: 'research',
-    children: [
-      { path: 'about', component: ResearchAboutComponent },
-      { path: 'articles', component: ResearchArticlesComponent },
-      { path: 'people', component: ResearchPeopleComponent },
-      { path: 'publications', component: ResearchPublicationsComponent },
-      { path: 'resources', component: ResearchResourcesComponent }
-    ]
-  }
+  { path: 'research', children: researchRoutes }
```
atruskie, line 32: This is looking compact 👍but I'd still like to see components define their own routes (or in your terminology a collection?)

src/app/app.module.ts
```diff
@@ -11,75 +11,43 @@ import { library } from '@fortawesome/fontawesome-svg-core';
...
 import { AudioAnalysisComponent } from './component/audio-analysis/audio-analysis.component';
 import { WebStatisticsComponent } from './component/web-statistics/web-statistics.component';
-import { LayoutComponent } from './component/shared/layout/layout.component';
 import { LayoutMenusService } from './services/layout-menus/layout-menus.service';
 
 import { BawApiService } from './services/baw-api/baw-api.service';
+import { aboutComponents } from './component/about/about.components';
+import { researchComponents } from './component/research/research.components';
+import { profileComponents } from './component/profile/profile.components';
+import { sharedComponents } from './component/shared/shared.components';
```
atruskie, line 32: Very nice, I like seeing this much red.

src/app/component/about/pages/contact/contact.component.menus.ts
```diff
@@ -0,0 +1 @@
+export const AboutContactIcon: [string, string] = ['fas', 'users'];
```
atruskie, line 1: This seems like it will produce a lot of files. Can we not export this route from the component file?

atruskie: Oh, I see you've removed it from contact.component.ts - why?

src/app/component/authentication/pages/login/login.component.menus.ts
```diff
@@ -0,0 +1,35 @@
...
+      icon: ActionMenuIcon,
+      label: ActionMenuTitle
+    },
+    links: [
+      {
+        route: 'https://www.ecosounds.org/my_account/password/new',
+        icon: ['fas', 'key'],
+        label: 'Reset password',
+        tooltip: 'Send an email to reset your password',
+        predicate: loggedin => !loggedin
```
atruskie, line 17: the predicate used on the old menu had the current user supplied as the first argument. `null` indicated a not logged in user...

src/app/component/authentication/pages/login/login.component.ts
```diff
@@ -5,28 +5,24 @@ import { FormGroup } from '@angular/forms';
...
+export class AuthenticationLoginComponent implements OnInit {
   private _jsonURL = 'assets/templates/login-form-template.json';
   form: FormGroup;
   model: {};
   fields: FormlyFieldConfig[];
   error: string;
 
-  secondaryLinks: SecondaryLink[];
+  secondaryLinks: Link[];
+  actionLinks: Link[];
```
atruskie, line 24: I thought the login component would have some actions (e.g. reset password). Not implemented yet?

src/app/component/authentication/pages/login/menus.service.ts
```diff
@@ -2,14 +2,13 @@ import { Injectable } from '@angular/core';
...
-import { action } from './menus.json';
+import { menus } from './login.component.menus';
 
 @Injectable({
   providedIn: 'root'
 })
 export class MenusService extends LayoutMenusService {
   constructor(private api: BawApiService) {
-    super(api, action);
+    super(api, menus);
```
atruskie, line 12: so I don't understand the point of these <component>.menu.service.ts files. What are they doing? Why is inheritance needed to register a menu? Will I see some use for these classes in a more complex page? If so can these boilerplate ones be omitted?

src/app/component/authentication/pages/register/register.component.menus.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const AuthenticationRegisterIcon: [string, string] = [
+  'fas',
+  'user-plus'
+];
+export const menus: Menus = {
+  action: {
+    title: {
+      icon: ActionMenuIcon,
+      label: ActionMenuTitle
```
atruskie, line 12: These names don't have much meaning. What is this a menu item for? I can't tell

src/app/component/authentication/pages/register/register.component.ts
```diff
@@ -21,9 +17,9 @@ export class RegisterComponent implements OnInit {
   model: {};
   fields: FormlyFieldConfig[];
 
-  secondaryLinks: SecondaryLink[];
+  secondaryLinks: Link[];
```
atruskie, line 20: I imagine `secondaryLinks` and `actionLinks` are part of a normal component definition? Are you just hitching a ride on `@component`?

I'd say it might we worth adding our own navigation and action decorators. That way the interface could be explicit and self-documenting.

src/app/component/profile/pages/home/home.component.menus.ts
```diff
@@ -0,0 +1,49 @@
...
+  action: {
+    title: {
+      icon: ActionMenuIcon,
+      label: ActionMenuTitle
+    },
+    links: [
+      {
+        route: 'https://www.ecosounds.org/my_account/edit',
+        icon: ['fas', 'edit'],
+        label: 'Edit my profile',
```
atruskie, line 15: generally, all of these strings should be pulled from translation files

src/app/component/profile/pages/home/home.component.menus.ts
```diff
@@ -0,0 +1,49 @@
+import { Menus } from 'src/app/services/layout-menus/menus';
+import { ActionMenuIcon, ActionMenuTitle } from '../../profile.menus';
+
+export const ProfileIcon: [string, string] = ['fas', 'user'];
+export const menus: Menus = {
```
atruskie, line 5: I would have imagined this menu defined elsewhere. It's more like the _common menu rather than the home page's menu

src/app/component/projects/pages/home/home.component.menu.ts
```diff
@@ -0,0 +1,28 @@
...
+export const ProjectsIcon: [string, string] = ['fas', 'globe-asia'];
+export const menus: Menus = {
+  action: {
+    title: {
+      icon: ActionMenuIcon,
+      label: ActionMenuTitle
+    },
+    links: [
+      {
+        route: 'https://www.ecosounds.org/projects/new',
```
atruskie, line 13: these hardcoded routes are a concern. I assume this is only an intermediate step?

src/app/component/research/research.routes.ts
```diff
@@ -0,0 +1,15 @@
+import { Routes } from '@angular/router';
```
atruskie, line 1: FYI everything in the _Research_ menu is part of a dynamic menu set in config.

Which mean all those components will probably be removed.

src/app/component/shared/footer/footer.component.html
```diff
@@ -5,7 +5,7 @@
         <p class="nav-link disabled m-0">&#169; QUT 2019</p>
       </li>
       <li class="nav-item">
-        <p class="nav-link disabled m-0">Production v3.0.0</p>
+        <p class="nav-link disabled m-0">Production v{{ version }}</p>
```
atruskie, line 8: I can't see where that is value is declared. We've been putting it in the config for the old site. Same here?

src/app/services/layout-menus/layout-menus.service.ts
```diff
@@ -1,93 +1,61 @@
...
-    @Inject('SECONDARY_LINKS_JSON') private _secondaryMenuJson?: any
-  ) {
-    // If no secondary menu items specified, use defaults
-    if (!_secondaryMenuJson) {
-      this._secondaryMenuJson = { ...secondary, _secondaryMenuJson };
-    } else {
-      this._secondaryMenuJson = secondary;
-    }
-  }
+    @Inject('MENU') private _menus?: Menus
```
atruskie, line 16: So I don't really understand this syntax. It looks like you're using Angular's DI framework to inject the MENU service into this private field `_menus`. Is that correct? Why? Why private?

src/app/services/layout-menus/layout-menus.service.ts
```diff
@@ -1,93 +1,61 @@
...
-        return link;
-      });
-    } else {
-      return this._secondaryMenuJson.no_auth.map(function(
-        link: SecondaryLink
-      ): SecondaryLink {
-        return link;
-      });
+  secondaryMenu(): Link[] {
+    const loggedIn = this._api.loggedIn;
```
atruskie, line 24: as per comment above, in the previous framework, this was just the User model of the logged in user

src/app/services/layout-menus/layout-menus.service.ts
```diff
@@ -1,93 +1,61 @@
...
+      }
+      return true;
+    });
   }
 
   /**
    * Returns the action title
    * @returns Action title and icon
    */
   actionTitle(): ActionTitle {
```
atruskie, line 46: Is `ActionTitle` the title that appears in the action bar? The name is a little vague, it sounds like it is the name of an action. The fact that is in the same bar is a design quirk; it  has no relation to actions.

Suggested alternative: `ActionListTitle`, ...

Also is it ever different from the title of the page? If not, can we just reuse the title?

src/app/services/layout-menus/layout-menus.service.ts
```diff
@@ -1,93 +1,61 @@
...
-    return {
-      label: this._actionLinksJson.title.label,
-      icon: [
-        this._actionLinksJson.title.icon.style,
-        this._actionLinksJson.title.icon.glyph
-      ]
-    };
+    return this._menus && this._menus.action.title
+      ? this._menus.action.title
+      : menus.action.title;
```
atruskie, line 49: see above comment.

this `actiontitle` if not the same as `title` has nothing to do with actions. Or menus. It's its own thing.

src/app/services/layout-menus/layout-menus.service.ts
```diff
@@ -1,93 +1,61 @@
...
-      return this._actionLinksJson.no_auth.map(function(
-        link: ActionLink
-      ): ActionLink {
-        return link;
-      });
-    }
+  actionLinks(): Link[] {
+    return this._menus && this._menus.action
+      ? this._menus.action.links
+      : menus.action.links;
```
atruskie, line 59: is this a default list? I missed its defintion

src/app/services/layout-menus/menus.ts
```diff
@@ -0,0 +1,125 @@
...
+import { AuthenticationLoginIcon } from 'src/app/component/authentication/pages/login/login.component.menus';
+import { AuthenticationRegisterIcon } from 'src/app/component/authentication/pages/register/register.component.menus';
+import { ProfileIcon } from 'src/app/component/profile/pages/home/home.component.menus';
+import { AudioAnalysisIcon } from 'src/app/component/audio-analysis/audio-analysis.component.menus';
+import { AboutReportIcon } from 'src/app/component/about/pages/report/report.component.menus';
+import { LibraryIcon } from 'src/app/component/library/pages/home/home.component.menu';
+import { DataRequestIcon } from 'src/app/component/data-request/data-request.component.menus';
+import { ProjectsIcon } from 'src/app/component/projects/pages/home/home.component.menu';
+import { SendAudioIcon } from 'src/app/component/send-audio/send-audio.component.menu';
+import { WebStatisticsIcon } from 'src/app/component/web-statistics/web-statistics.component.menu';
```
atruskie, line 14: k, so I'm not keen on this.

We should be importing not just the icon, but the route, label, tooltip, as well.

I know you modelled this global menu off of my suggestion (thanks) but I think we need it iterate on this further. It's just too spread out.

I think these secondary links should be registered next to the component definition, probably in the same file as the component.

That way each navigatable component will:
- define its routes
- define any navigation links
- define any actions
- define its icon
- define the component

src/app/services/layout-menus/menus.ts
```diff
@@ -0,0 +1,125 @@
...
+  title?: ActionTitle;
+  links: Link[];
+}
+export interface ActionTitle {
+  icon: [string, string];
+  label: string;
+}
+
+export interface Link {
+  route: string;
```
atruskie, line 120: add `href` option as an exclusive choice to route.

Or better yet: define a `uri` member of type `Href | Route` and strongly type the values.

src/app/services/layout-menus/menus.ts
```diff
@@ -0,0 +1,125 @@
...
+  icon: [string, string];
+  label: string;
+  tooltip: string;
+  predicate?: (user: boolean) => boolean;
```
atruskie, line 124: the argument name `user` is misleading for the current implementation (though correct for the original intent)

atruskie (approved):
This looks like quite a bit of work. It's moving in the right direction.

I'll approve changes because we're moving quickly, but I expect we'll iterate on this design a bit more.

On the whole, I'm looking for fewer files per page-component, more things located next to their relative components, and less hacks (probably a custom type decroator to store this navigation information).

src/app/component/about/pages/contact/contact.component.menus.ts
Allcharles: In my latest work I'm moving it back in. Just didn't know how to access the icon from the component however I've solved that issue.

src/app/component/authentication/pages/login/login.component.menus.ts
Allcharles: I was wondering about that. I know there are pages on the ecosound website for administrator accounts only. Do you use the current user to show those routes if the `user === administrative account`? Otherwise i would argue this is simpler to understand at a glance.

src/app/component/authentication/pages/login/login.component.ts
Allcharles: Not implemented yet, once I've got the menus working I'll probably have those options up and running pretty quick.

src/app/component/authentication/pages/login/menus.service.ts
Allcharles: They were included as i was unsure how to update the LayoutMenusService on construction from the component. I will have removed these in the next update as it'll be just the one service controlling all the menus.

src/app/component/authentication/pages/register/register.component.ts
Allcharles: So reverting this change and making the `SecondaryLink` and `ActionLink` interfaces again?

src/app/component/shared/footer/footer.component.html
Allcharles: It is read directly from the package.json file, I was looking online and this should also be safe so that web users wont be able to see the contents of the package.json file. Only the version number.

src/app/services/layout-menus/layout-menus.service.ts
Allcharles: I am a little unsure how injections function, mostly this is just here so that the project is able to build for production. What it is doing, is allowing the project to initialise the service on startup. For each constructor input which is not a module, or a service, you need to provide a default value which can then be overwritten later.

src/app/services/layout-menus/layout-menus.service.ts
Allcharles: It is different for a couple pages including all of the about pages. I will however update the variable name.

src/app/services/layout-menus/layout-menus.service.ts
Allcharles: Yes, the default action list containers the header "Home" with no action options. I may change it to just be empty space based on your suggestion during the meeting however.

src/app/component/authentication/pages/login/login.component.menus.ts
atruskie: the intention was that with a user account, we could test for any role-based authorization stuff as well. But that never happened. Adding admin links would be a good use case.

src/app/component/authentication/pages/login/menus.service.ts
atruskie: Great. See my decorator comments for how I think this could work.

src/app/component/authentication/pages/register/register.component.ts
atruskie: I'm talking here about using the component definition to store this stuff. That seems potentially leaky.

I'd suggest investigating a custom type decorator that would allow us to do:

```
@Page({
  secondaryLinks: ...,
  actionLinks: ...
})
// normal component defintition...
@Component({
```

src/app/services/layout-menus/layout-menus.service.ts
atruskie: So that's pretty standard for a DI framework. The question is more, why is not a "module or a service"?

Issue #5, opened 2019-08-03
Action links with menus


Allcharles:
Action link components cannot reference any action components in the action menu. A generic use case where this could impact is the confirm account component. Attempting to give it the same actions as the login component will result in the action menu crashing. This is because the confirm account component (and ResetPassword, UnlockAccount) will not have been generated when the action menu is built, thus it returns a null value.

While this is currently not an issue, it may become one in the future.

Edit:
A misunderstanding of the issue. A component can successfully place any component inside the action menu outside of itself, and its parent/s. This should not be an issue in the future, and it is likely there is no solution using the current model.

Allcharles:
Example of a broken Page system:

Paste the following in `reset-password.component.ts`

``` javascript
import { ConfirmPasswordComponent } from "../confirm-account/confirm-account.component";
import { LoginComponent } from "../login/login.component";
import { UnlockPasswordComponent } from "../unlock-account/unlock-account.component";

@Page({
  icon: ["fas", "key"],
  label: "Reset password",
  category: securityCategory,
  routeFragment: "reset_password",
  tooltip: () => "Send an email to reset your password",
  menus: {
    actions: List<AnyMenuItem>([
      LoginComponent.pageInfo,
      ResetPasswordComponent.pageInfo,
      ConfirmPasswordComponent.pageInfo,
      UnlockPasswordComponent.pageInfo
    ]),
    links: null
  },
  order: { priority: 2, indentation: 1 }
})
```

This should result in the following errors:

When component references parent:

```
ReferenceError: can't access lexical declaration `LoginComponent' before initialization

Circular dependency detected:

src/app/component/authentication/pages/login/login.component.ts -> src/app/component/authentication/pages/reset-password/reset-password.component.ts -> src/app/component/authentication/pages/login/login.component.ts
```

When component references itself:

```
ERROR TypeError: "link is null"
```

PR #7, opened 2019-08-12
Route experiments

atruskie:
Redesigned routing.

It's a generally better design that is more imperative and requires less searching when the app boots.

It also adds a so-called `StrongRoute` type that can output a single route as a string (`"/security/login"`) for as an array of commands via the `format` method (`["secruity", "login"]`) and will insert route parameters into the right spots.

This patch also enables the Ivy renderer and was checked to be working with the angular compiler aot option.

It, does however, require an update to ngx-formly: https://github.com/ngx-formly/ngx-formly/issues/1693


src/app/component/shared/menu/menu.component.html
```diff
@@ -24,15 +24,15 @@ <h6 class="nav-link disabled m-0">
...
+        <ng-container *ngIf="isInternalLink(link)">
           <app-menu-internal-link
             [id]="menuType + '-tooltip-' + i"
             [link]="link"
             [placement]="placement"
           >
           </app-menu-internal-link>
         </ng-container>
-        <ng-template #elseExternalLink>
+        <ng-template *ngIf="isExternalLink(link)">
```
Allcharles: This should be an `ng-container`. Reason why is here: https://stackoverflow.com/questions/40529537/ng2-difference-between-ng-container-and-ng-template-tags

angular.json
```diff
@@ -22,8 +22,8 @@
             "main": "src/main.ts",
             "polyfills": "src/polyfills.ts",
             "tsConfig": "tsconfig.app.json",
-            "aot": false,
-            "assets": ["src/assets"],
+            "aot": true,
```
Allcharles, line 25: I can see from the tsconfig that you are using ivy which requires AOT. However, we should have a separate build process for developing if possible as AOT disables the debugging tools such as Augury.

src/app/component/error/errorModule.ts
```diff
@@ -0,0 +1,17 @@
...
+      {
+        path: "**",
+        component: PageNotFoundComponent
+      }
+    ])
+  ],
+  declarations: [PageNotFoundComponent],
+  exports: [PageNotFoundComponent, RouterModule]
+})
+export class ErrorModule {}
```
Allcharles: I am assuming you wish to put more components into the error module?
  - Page not found
    - For this, are we going to attempt to use it in a similar manner to ecosounds or create a different screen?
  - Unauthorised access
  - Resource has moved?

src/app/interfaces/Routing.ts
```diff
@@ -0,0 +1,116 @@
...
+                } else {
+                    throw new Error(`Parameter named ${x.name} was not supplied a value and a default value was not given`);
+                }
+            } else {
+                return x.name;
+            }
+        };
+
+        return this.full.map(prepareParam);
+    }
```
Allcharles: `format` function is never used

src/app/interfaces/Routing.ts
```diff
@@ -0,0 +1,116 @@
...
+            // generate it.
+            callback(current.pageComponent, current.config);
+            const thisRoute = current.routeConfig;
+            const childRoutes = current.children.map(recursiveAdd);
+            // TODO: the order of these arrays may well need to be changed...
+            // depends on how nested routes work
+            // we want "" to match current
+            // but e.g. "sites" to go down a level
+            // and not for both to match at same time.
+            thisRoute.children = [...(thisRoute.children || []), ...childRoutes];
```
Allcharles: Would this not just be a simple sort function? eg. `If string.length === 0 return higher priority, else sort by alphabet`?

Allcharles (approved):
Changing the names of default files (eg. app-routing.module.ts) may cause some confusion if a new angular developer works on this project.
There are a few bugs stopping the project from building but I've already worked out most of them (mostly just broken imports).

angular.json
atruskie: Yeah, I saw that augury didn't work with AOT :-( 

The [Ivy guide](https://angular.io/guide/ivy) suggests always using AOT:

>AOT compilation with Ivy is faster and should be used by default

I'll look into adding another build.

src/app/component/error/errorModule.ts
atruskie: Those are good ideas.

Mainly I wanted 404s to render a page on the faulty URL and not redirect to `/`. I just tried to follow the convention we had set up for other components/modules.

src/app/interfaces/Routing.ts
atruskie: Never used yet. It's designed to be used in templates. It is the most useful for parameterized routes, of which we did not have any yet in this branch. I'll update some of the templates...

src/app/interfaces/Routing.ts
atruskie: Well, there's no need to sort when you can place the items in the order you want.

What I'm unsure of is the order they should be. I suspect we'll find out quickly as soon as we add a nested route.

atruskie:
> Changing the names of default files (eg. app-routing.module.ts) may cause some confusion if a new angular developer works on this project.

Ah, I didn't realise this was a default name.

Standard naming of JS files should be pascalCase. I get angular components being kebab-case because that is the same name as their html tag. However, since app-routing was not a component...

Allcharles:
Do you know if this will address the issue #5? The biggest issue with the old system was that a component could not reference its parent or itself in the action menu (and partially the secondary menu which would break the ability for breadcrumbs).

atruskie:
> Do you know if this will address the issue #5? The biggest issue with the old system was that a component could not reference its parent or itself in the action menu (and partially the secondary menu which would break the ability for breadcrumbs).

I intended it to solve #5, but didn't check. Will follow up. 

In #5 can you add an example configuration that did not work?

Allcharles (approved):

Allcharles:
Currently receiving the following error when I try to npm start:
`ERROR in Cannot read property 'loadChildren' of undefined`

atruskie:
yup, still working on it. None of the solutions I've found so far are adequate. This problem does not happen with the Ivy renderer...

Don't merge yet

PR #17, opened 2019-10-15
Google Maps

### LINKED ISSUE #10: Google Maps
AGM does not support ivy, thus it crashes in the current version. However this has been fixed in their next release 1.0.0-beta.8 which is yet to be released.
Relates to https://github.com/SebastianM/angular-google-maps/issues/1647

Allcharles:
Added google maps to project and site pages. Zoom level on the map may need some tweaking later 

New dependencies:
- [@agm/core](https://github.com/SebastianM/angular-google-maps)
- [snazzy-info-window](https://github.com/SebastianM/angular-google-maps)

Closes #10 

Edit # 1: Updated snazzy-info-window link
Edit # 2: Example to show off update
image


atruskie:
I'm dubious of the utility of the snazzy-info-window plugin, essentially because the repository hasn't been updated for two years.

Can you show me a screenshot?

src/app/component/projects/pages/details/details.component.html
```diff
@@ -56,18 +57,46 @@ <h5>
...
 
+<!-- Unauthorized Access Display -->
 <ng-container *ngIf="state === 'unauthorized'">
   <app-unauthorized></app-unauthorized>
 </ng-container>
 
+<!-- Project Not Found Display -->
 <ng-container *ngIf="state === 'notFound'">
   <app-item-not-found></app-item-not-found>
 </ng-container>
```
atruskie, line 50: It seems strange these are part of the details view. Do they have to be specified for every view? Is there a more general way of doing this?

src/app/component/projects/pages/details/details.component.ts
```diff
@@ -54,6 +54,8 @@ export class DetailsComponent extends PageComponent implements OnInit {
   sites: Site[];
   errorCodes = this.sitesApi.apiReturnCodes;
   state = "loading";
+  avgLat = 0;
```
atruskie: don't abbreviate names

src/app/component/projects/pages/details/details.component.ts
```diff
@@ -82,6 +84,13 @@ export class DetailsComponent extends PageComponent implements OnInit {
 
         this.sitesApi.getProjectSites(params.projectId).subscribe({
           next: sites => {
+            sites.map(site => {
+              this.avgLat += site.customLatitude;
+              this.avgLong += site.customLongitude;
+            });
+            this.avgLat /= sites.length;
+            this.avgLong /= sites.length;
+
```
atruskie: Averaging of lat/long doesn't work like this.

But more so, why are we averaging?

atruskie (commented):
I'd like to see if what the snazzy dep offers, and understand what the averaging is about.

But otherwise, nice, and clean ✔

src/app/component/projects/pages/details/details.component.html
Allcharles: I would almost prefer a more general way of performing this. Perhaps a component which takes a state variable as an input?

src/app/component/projects/pages/details/details.component.ts
Allcharles: We need the average latitude and longitude for the `<agm-map>` component as it does not have an inbuilt averaging feature (as far as I tested, will look over again). I will look into a better implementation of this calculation.

Allcharles:
> I'm dubious of the utility of the snazzy-info-window plugin, essentially because the repository hasn't been updated for two years.
> 
> Can you show me a screenshot?

Snazzy info window is required to customize the google maps plugin. I noticed that the original website uses this to set a background on the location pin labels for readability. Currently agm core recommends the snazzy-info-window plugin as they maintain it and even give a guide to install it on the google maps plugin page.
As for the last update, it appears I linked to the wrong repo (the official guides also link to it in my defense). I suspect that the npm package is also provided by the angular-google-maps repo.

src/app/component/projects/pages/details/details.component.ts
atruskie: No, please do no implement such an algorithm.

But what is being averaged here? Are we averaging the site locations? Why?

Previous maps just showed all sites in the project details page - why not again?

src/app/component/projects/pages/details/details.component.html
atruskie: So when errors like this happen, the routing system should render an error component in place of the component that was meant to be rendered.

E.g. 404 on projects/123 should render <not-found /> in place of <project-details />

This should be a separate issue, let's not fix it here. Create an issue, please.

atruskie:
K, let's leave snazzy in.

Just got to address the averaging (see [comment](https://github.com/QutEcoacoustics/workbench-client/pull/17#discussion_r335782742)) and the build needs to pass.

src/app/component/projects/pages/details/details.component.html
Allcharles: Issue created here #30.

src/app/component/projects/pages/details/details.component.ts
Allcharles: This is to average the site locations, its needed because you need to supply the `<agm-map>` component with a longitude and latitude to focus the map on. Adding pins has no affect on the zoom level or position of the map. Below is an image of the map that is generated. The pins are added using a for loop in the html which iterates over the sites and adds pins, snazzy-info-window creates the white background for the pin label, and the average latitude and longitude focus the map to where the pins are located.
image

Allcharles: Resolved: Found how to use the agm/core library to auto focus on the pins

src/app/component/projects/pages/details/details.component.ts
Allcharles: Resolved: Removed variables

src/app/component/projects/pages/details/details.component.ts
atruskie: Fantastic 👍

atruskie:
Since the CI is broken, can you let me know when the relevant tests are working?

Allcharles:
Currently the google maps api key is being loaded from the environment file. I am hoping to move that to the external config file in another pull request from the branch [extract-environment-values](https://github.com/QutEcoacoustics/workbench-client/tree/extract-environment-values).

PR #67, opened 2020-01-08
Standardized API Service Access

### LINKED ISSUE #64: Standardize api get request function body
Standardize function inputs for api GET requests to be the same as the other HTTP Requests.

### LINKED ISSUE #65: Standardize api service functions
Currently there is a lot of overlapping code where a models service has its own custom update, new, list all, list, delete functions. This could be changed into one generic service and inherited from.

Allcharles:
Abstracted the logic from model services (sites, projects, user) into one model service file. Sites/Projects/User services now extend that file which allows for more in-depth testing and far better support for filtered requests.

Things to review

- `model.service.ts` api logic
  - Check correct HTTP request method (GET/POST/PATCH)
  - Check filter logic on details and show functions are correct
- `model.service.spec.ts` unit tests
- `sites.service.ts` api logic
  - Do the functions use the correct model service function?
- `projects.service.ts` api logic
  - Do the functions use the correct model service function?
- `user.service.ts` api logic
  - Do the functions use the correct model service function?
- baw-api/mock
  - Check mock services have the same public functions as their originals

Closes: #65 
Closes: #64


src/app/models/Project.ts
```diff
@@ -86,3 +86,11 @@ export class Project implements ProjectInterface {
...
   }
 }
+
+export const mockProject = new Project({
+  id: 1,
+  name: "name",
+  description: "description",
+  creatorId: 1,
+  siteIds: new Set([1])
+});
```
Allcharles, line 96: Creating a mock project may be a bad practice for the unit tests which implement it

src/app/models/Site.ts
```diff
@@ -82,3 +82,12 @@ export class Site implements SiteInterface {
...
 }
+
+export const mockSite = new Site({
+  id: 1,
+  name: "name",
+  description: "description",
+  creatorId: 1,
+  projectIds: new Set([0]),
+  imageUrl: "/assets/images/site/site_span4.png"
+});
```
Allcharles, line 93: Creating a mock site may be a bad practice for the unit tests which implement it

src/app/services/baw-api/mock/baseApiMockService.ts
```diff
@@ -1,12 +1,17 @@
...
 
-  public isLoggedIn() {}
+  public isLoggedIn(): boolean {
+    return false;
+  }
 
-  public getSessionUser() {}
+  public getSessionUser(): SessionUser | null {
+    return null;
+  }
```
Allcharles: Don't need function return typing in mock services

src/app/services/baw-api/model.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      throw new Error(
+        `Expected ${count} arguments to satisfy url parameters for path '${path}'. Instead received ${args.length}.`
+      );
+    }
+
+    return path;
+  }
+}
+
+export type Args = (string | number)[];
```
Allcharles: Missing type XML comment

src/app/models/Project.ts
atruskie: Generally faking frameworks can help us generate better test data: https://github.com/marak/faker.js

src/app/services/baw-api/model.service.spec.ts
```diff
@@ -0,0 +1,2565 @@
...
+  it("update should handle error", done => {
+    const service: ModelService<MockModel> = TestBed.get(ModelService);
+    service["update"]("/broken_path", null).subscribe(
+      () => {
+        expect(true).toBeFalsy("Service should not generate data response");
+      },
+      (err: APIErrorDetails) => {
+        expect(err).toBeTruthy();
+        expect(err).toEqual({
+          stat
```
atruskie: I've managed to understand a little of this file, but:

- it is way too long
- has way too much repetition in it

Can you find some ways to break it up?

src/app/services/baw-api/model.service.ts
```diff
@@ -0,0 +1,143 @@
+import { HttpClient } from "@angular/common/http";
+import { Inject, Injectable, InjectionToken } from "@angular/core";
+import { Subject } from "rxjs";
+import { AppConfigService } from "../app-config/app-config.service";
+import { APIErrorDetails } from "./api.interceptor";
+import { BawApiService, Filters } from "./base-api.service";
+
+export let MOCK_CLASS_BUILDER = new InjectionToken("test.class.builder");
```
atruskie: This seems like a conflation of concerns. Why is a test token defined in a non-test file? Is there a better spot for it?

src/app/services/baw-api/model.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    subject.complete();
+  }
+  private subjectError(subject: Subject<T | T[]>, err: APIErrorDetails) {
+    subject.error(err);
+  }
+
+  constructor(
+    http: HttpClient,
+    config: AppConfigService,
+    @Inject(MOCK_CLASS_BUILDER) private classBuilder: (object: any) => T
```
atruskie: Why does this need to be injected? Why is it a 'mock' builder?

src/app/services/baw-api/model.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    return subject;
+  }
+
+  /**
+   * Replace URL parameters with argument values
+   * @param path URL path
+   * @param args Arguments to insert
+   */
+  private refineUrl(path: string, args: Args): string {
```
atruskie: Is it possible to reuse angular's URL templating functions? We should do so

src/app/services/baw-api/projects.service.ts
```diff
@@ -9,71 +9,49 @@ import {
...
 })
-export class ProjectsService extends BawApiService {
+export class ProjectsService extends ModelService<Project> {
+  private paths: {
+    [key: string]: string;
+  };
+
   constructor(http: HttpClient, config: AppConfigService) {
-    super(http, config);
+    const classBuilder = (project: ProjectInterface) => new Project(project);
```
atruskie: is this needed for every model?

Let's make it optional. Doing a `new T()` should be possible inside of `ModelService`

atruskie (changes requested):
There seems like a lot of repetition in the `xxxMockService` classes. I think this can be simplified but it's really a smell of another issue.

I like the common functionality abstraction. However, why don't you use common names as well? Why `getProjects`, `getProject`, etc over just `get` and `getAll`? Oh I see why you're doing this. I'd suggest trying overriding methods via inheritance rather than doing a bunch of plumming in every child class.

We use the common names `list` (GET /x), `show` (GET /x/123), `destroy`, `create`, and `update` in rails - it seems like it would be a good idea to reuse those names. I'd say that using "details" as the name for "list" is confusing.

This has all been done before in `$resource` in Angular 1.x. I'd suggest reading the docs https://docs.angularjs.org/api/ngResource/service/$resource for some history.

Your idea of abstracting service commonality into one place is great but needs some work.

- I'd avoid the name `ModelService` - its mixing two very general names and is not clear by itself. What about `WorkbenchApiService`, or `ApiCommon`, or `ApiService`?
- I'd customize each child class with route parameters when they're instantiated and using that customization for storing path and host data, rather than passing the path/URL into each service method as a parameter

src/app/services/baw-api/model.service.ts
Allcharles: `MOCK_CLASS_BUILDER` is used by the model.service unit tests. The reason it is a mock is because in the actual application `ModelService` is never actually used, only extended from. Any non service inputs in a service must have an injector so that the app.module file can instantiate default values. Thus to test the file, i need a mock injector which is only ever used by the model service unit tests.

src/app/services/baw-api/model.service.ts
Allcharles: The injection token is used by the model.service.spec.ts file so moving this variable outside of the model.service file would be a bit awkward. Potentially i can move it to the injection token definition to the unit testing file and import it into the service file?

src/app/services/baw-api/model.service.spec.ts
Allcharles: I have two immediate methods I could use to split up the unit tests. Either I create multiple `describe` functions inside the file, or I can split the file into multiple smaller files.

The first method will maintain a large file, however I can group unit tests by the method it tests, and the unit test output will separate the tests into different groups similar to separate unit test files. 

The second method of splitting the file up would have a similar effect to the first method in terms of output from the unit testing framework, however it would add unnecessary files to the directory. I would likely split the files by method name resulting in the following file name format: `model.<method>.spec.ts`.

Which method do you prefer.

As for the repetition, some of the no filter vs filter tests could be removed with regards to error handling (as that is independent of filter), however this would make the assumption that someone wont break that independence in the future which I'm not a huge fan of. While I understand there is a lot of repetition, I'm unsure how I would reduce it without reducing the coverage of the testing.

atruskie: I'll let you decide.

If you can factor out enough common code, then keep it one file.

Normally I'd baulk at more files, but 2.5K lines in a file are unwieldy. Breaking out into methods is fine if you can't factor out helper methods/common code etc

src/app/services/baw-api/model.service.ts
atruskie: I think my biggest issue is one of naming then. "Mocks" shouldn't exist in non-test code.

src/app/services/baw-api/model.service.ts
atruskie: See [comment](https://github.com/QutEcoacoustics/workbench-client/pull/67#discussion_r368373182) on naming. What about "stub" or "abstract"?

src/app/services/baw-api/projects.service.ts
Allcharles: Seems to be possible, although not as clean as you are thinking: https://stackoverflow.com/questions/17382143/create-a-new-object-from-type-parameter-in-generic-class

src/app/services/baw-api/model.service.ts
Allcharles: As far as I can tell, there is no exposed method of handling URL attributes. HttpClient only exposes a method of adding HTTPParams which are the attributes placed after the `?` in a url. Eg. `https://brokenlink?http_params_go_here`.

Allcharles:
Okay I've fixed as many problems as I was able to. The major change you want to make which will remove the "plumbing" will have to wait for another PR. I can do it, its just that that much of a change will be very difficult to pull through the other 3 PRs. I have created an issue for it here: #78

src/app/services/baw-api/projects.service.ts
atruskie: Yep, you're right.

I'd suggest following this pattern:

```typescript
//model
class foo {
    constructor() {
    }

    name: string;
}

abstract class CommonApi<T> {
    abstract modelFactory(): T;
    

    get() {
        return this.modelFactory();
    }
}

class FooService extends CommonApi<foo> {
    modelFactory() { return new foo() };

}
```

src/app/services/baw-api/model.service.ts
atruskie: https://angular.io/api/router/Router#parseurl

src/app/services/baw-api/projects.service.ts
Allcharles: I found what I consider to be a cleaner solution in this commit: https://github.com/QutEcoacoustics/workbench-client/pull/67/commits/fe59ca29ed64861a5096f2aa1226580fbf0a1341.

Basically it allows me to extend the `ApiCommon` class like this:

```javascript
class ApiCommon<T> {
  constructor(
    http: HttpClient,
    config: AppConfigService,
    private type: new (object: any) => T
  ) {
    super(http, config);
  }
}

class ProjectsService extends ApiCommon<Project> {
  constructor(http: HttpClient, config: AppConfigService) {
    super(http, config, Project);
  }
}
```

atruskie (approved):

PR #68, opened 2020-01-09
Project and Site Deletion

Allcharles:
Added the ability to delete projects and sites. The implementation is a bit different to the original website, however I believe its far more modern and user friendly.

Things to check:
- Try delete a project
- Try delete a project you don't have access to
- Try delete a project you aren't the owner of
- Try delete a project when you are not logged in
- Try delete a site
- Try delete a site you don't have access to
- Try delete a site you aren't the owner of
- Try delete a site when you are not logged in

Blocked by PR #67 

Closes #63 

## Visual Changes

> Added delete project action link:

image

> Delete project page:

image

> Failure to delete project:

image

> Added delete site action link

image

> Delete site page:

image

> Failure to delete site:

image

PR #70, opened 2020-01-09
Notification system

### LINKED ISSUE #69: Create notification system
Currently notifications are restricted to a single page. This causes issues for forms which wish to redirect the user on a successful submission, however are unable to display information to the user about the success.

Create a notification service which runs in the app component so that any notifications are shown at the top of the current component (not sure how this should look visually on fullscreen components as we only have the home component right now).

- Components using this notification should have 3 options: danger, success, warning. These relate to the bootstrap colours.
- Alerts should not be overridden, instead they should pile on top of each other ordered by newest to oldest
- Service should use an observable trigger to cause updates to the app component

Allcharles:
Created a notification system for the website. This will allows forms to redirect users to another page whilst also allowing notifications to follow them. Eg. Create a new site -> get sent to site page with notification at the top of the screen for success **this is not implemented yet**.

Blocked by PR #68 
Closes #69 

## Visual Changes

image


atruskie (changes requested):
I'm on a campaign to have less code. This is great work, but if we're going to have toast notifications, then please use a library. https://github.com/scttcper/ngx-toastr looks good.

atruskie (approved):

PR #75, opened 2020-01-10
Create contributing guildeline

Allcharles:


atruskie:
There are some not-relevant sections to that document. They should be removed (e.g. mentions of `StyleCop`)

Allcharles:
Outside of StyleCop I'm unsure what other non-relevant sections remain. The only section I can see is the best practices and how it relates to the un-prefixed SI units however that should become relevant once we add the /listen pages.

Issue #78, opened 2020-01-20
Remove plumbing from api services


Allcharles:
Abstract API Services further such that  standardized method names are used (eg. `get`, `getAll`, ...) by all services (Project/Site/User/...). This will further reduce the amount of unit tests and code debt.

Possible implementation is discussed in the following comment: https://github.com/QutEcoacoustics/workbench-client/pull/67#pullrequestreview-345064046. However a workaround will need to be developed for edge case Services such as UserService which will have two `/get` routes (could just split the service).

atruskie:
So the structure I've previously established, which I think will work well again is:

Standard methods:

name | method  | url format | returns collection?
--- | --- | --- | ---
list | GET |  /resource | ✅
show | GET | /resource/:id | ❌
create | POST | /resource | ❌
update | PATCH | /resource/:id | ❌
destroy | DELETE | /resource/:id | ❌
filter | POST |  /resource | ✅

Previously everything that implemented the abstract API methods had to implement ALL the API methods. I'd like it if we could omit the standard methods when appropriate.

Additionally, special case methods should be addable to child service classes when needed.

Example:

```typescript
abstract class AbstractApi<T> {
    authority: string;
    path: string;
}

abstract class ApiList<T> extends AbstractApi<T> {
    list(): T[] {
        console.log(this.path);
        return [];
    }
}
abstract class ApiShow<T> extends AbstractApi<T> {
    show(id: bigint): T {
        return null;
    }
}

// helper types

abstract class StandardApi<T> {
}
interface StandardApi<T> extends ApiShow<T>, ApiList<T> {

}

abstract class ApiWithoutShow<T> {
}
interface ApiWithoutShow<T> extends ApiList<T> {}


 // models

class Project {
    id: bigint;
    name: string;
}

class Site {
    id: bigint;
    name: string;
}

// services 

class ProjectService extends StandardApi<Project> {

    // example special custom method
    requestAccess(projects: Project[]) {
        // blah blah blah
        this.path
    }


}

class SiteService extends ApiWithoutShow<Site> {
    // example resource that does support `show`

    // and a special custom method
    updateOrphan(orphan: Site, newParent: Project) {
        // blah blah blah
        this.path
    }
}

class SiteService2 extends ApiWithoutShow<Site> {
    // example of overriding standard method
    show(id: bigint, parentId: bigint): Site {
        return null;
    }
}

// syntax check

function test() {

    let service = new ProjectService();

[...]

PR #80, opened 2020-01-23
String templating in API Services

### LINKED ISSUE #79: Add string templating and replace http url construction
- [x] Add code from the following: https://gist.github.com/atruskie/12abf8cc7979eec14fbee956f13e943c
- [x] replace the implentation of `refineUrl`  (workbench-client/src/app/services/baw-api/api-common.ts) with invocations of the template function

Allcharles:
- Replaced the implementation of `refineUrl` with `makeTemplate`
- Improved http request handling
  - Remove separate http type functions (get, post, patch, delete)
  - Created apiRequest function (list, filter, show, create, update, delete)

Closes: #79


src/app/services/baw-api/base-api.service.spec.ts
```diff
@@ -65,4 +65,58 @@ describe("BawApiService", () => {
     expect(service.getSessionUser().authToken).toBe("aaaaaaaaaaaaaaaaaaaaaa");
     expect(service.getSessionUser().userName).toBe("username");
   });
+
```
atruskie: these tests can be moved to a `stringTemplateTests.ts` class

src/app/services/baw-api/base-api.service.ts
```diff
@@ -62,126 +64,72 @@ export abstract class BawApiService {
...
+          .subscribe(this.handleResponse(next, error));
+      case "UPDATE":
+        return this.http
+          .patch<APIResponse>(path, body)
+          .subscribe(this.handleResponse(next, error));
+      case "DELETE":
+        return this.http
+          .delete<APIResponse>(path)
+          .subscribe(this.handleResponse(next, error, true));
+    }
```
atruskie: revert this change, please. 

Why not use methods?
Stringly type invocation is not fantastic  - especially since you didn't handle any other case in the switch. (I do soee the string-type restriction in the parameter type, but types get erased at compile time).

src/app/services/baw-api/base-api.service.ts
```diff
@@ -62,126 +64,72 @@ export abstract class BawApiService {
...
+   * parameter type compatibility with the placeholders from the interpolated string.
+   * @param strings The strings around the tokens to template
+   * @param placeholders Placeholders that are substituted for token when
+   *  templating is done. Note these are transform functions.
    */
-  protected apiDelete(
-    next: (data: any) => void,
-    error: (err: any) => void,
-    path: string
+  protected makeTemplate<T extends ((any: any) => any)[]>(
```
atruskie: makeTemplate is general enough to use anywhere. Put it in its own file, perhaps `stringTemplating.ts`

src/app/services/baw-api/security.service.ts
```diff
@@ -23,35 +23,34 @@ export class SecurityService extends BawApiService {
 
     this.loggedInTrigger.next(this.isLoggedIn());
     this.paths = {
-      register: "/security",
-      signIn: "/security",
-      signOut: "/security"
+      register: this.makeTemplate`/security`,
```
atruskie: you probably don't _need_ to use make template around a string with no parameters

atruskie (approved):
Some minor changes required. Factor out `makeTemplate` mainly.

I think you should tackle #78 next.

src/app/services/baw-api/security.service.ts
Allcharles: This is mostly just for standardising the services before i tackle #78 .

src/app/services/baw-api/base-api.service.ts
Allcharles: My bad, misread https://docs.angularjs.org/api/ngResource/service/$resource and thought it was building a single function which handles all the requests instead of a class.

PR #81, opened 2020-01-23
Seconday Menu Indentation System

### LINKED ISSUE #58: Improve indentation system on secondary links
Currently, links added to the secondary links menu have to specify how many indentations they require. This can however be determined by counting its number of parents.

This information can be gathered from here:
secondary-menu.component.ts
```javascript
const parentMenuRoutes: MenuRoute[] = [];
let menuRoute = current;
while (menuRoute.parent) {
  menuRoute = menuRoute.parent;
  parentMenuRoutes.push(menuRoute);
}
```

This would reduce the amount of information that must be know before compile time in the `PageComponents`, as well as potentially allowing the same component to have two direct parents.

Allcharles:
Improved secondary menu indentation system so that indentation does not need to be defined by page components. Components specify their parent, which is then used to calculate its indentation level and organise sub-links.

Edit 1:
Order and indentation can now be calculated for all sub-links automatically. Order for top level links must still be set manually.

Closes #58

### DIFF
--- a/src/app/component/shared/secondary-menu/secondary-menu.component.ts
+++ b/src/app/component/shared/secondary-menu/secondary-menu.component.ts
@@ -54,13 +54,21 @@ export class SecondaryMenuComponent implements OnInit, OnDestroy {
           indentation++;
         }
 
+        // Update indentation values
         if (current.order) {
           current.order.indentation = indentation;
+        } else {
+          current.order = { indentation };
         }
 
         parentMenuRoutes.forEach(parent => {
           indentation--;
-          parent.order.indentation = indentation;
+
+          if (parent.order) {
+            parent.order.indentation = indentation;
+          } else {
+            parent.order = { indentation };
+          }
         });
 
         // with any links from route
--- a/src/app/component/shared/secondary-menu/secondary-menu.component.ts
+++ b/src/app/component/shared/secondary-menu/secondary-menu.component.ts
@@ -42,7 +42,6 @@ export class SecondaryMenuComponent implements OnInit, OnDestroy {
         const defaultLinks = DefaultMenu.contextLinks;
 
         // and current page
-        let indentation = 0;
         const current = page.self;
 
         // and parent pages
@@ -51,25 +50,7 @@ export class SecondaryMenuComponent implements OnInit, OnDestroy {
         while (menuRoute.parent) {
           menuRoute = menuRoute.parent;
           parentMenuRoutes.push(menuRoute);
-          indentation++;
         }
-
-        // Update indentation values
-        if (current.order) {
-          current.order.indentation = indentation;
-        } else {
-          current.order = { indentation };
-        }
-
-        parentMenuRoutes.forEach(parent => {
-          indentation--;
-
-          if (parent.order) {
-            parent.order.indentation = indentation;
-          } else {
-            parent.order = { indentation };
-          }
-        });
 
         // with any links from route
         const links = page.menus?.links
--- a/src/app/component/shared/secondary-menu/secondary-menu.component.ts
+++ b/src/app/component/shared/secondary-menu/secondary-menu.component.ts
@@ -42,6 +42,7 @@ export class SecondaryMenuComponent implements OnInit, OnDestroy {
         const defaultLinks = DefaultMenu.contextLinks;
 
         // and current page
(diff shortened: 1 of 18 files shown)


src/app/interfaces/menusInterfaces.ts
```diff
@@ -53,12 +53,12 @@ export interface Order {
...
    */
-  priority: number;
+  priority?: number;
 
   /**
-   * Indentation of link
+   * Indentation of link. Value is calculated by SecondaryMenuComponent
    */
-  indentation: number;
+  indentation?: number;
```
atruskie: if this option is no longer user-configurable  then let's remove it from the interface

src/app/component/shared/secondary-menu/secondary-menu.component.ts
```diff
@@ -50,8 +51,26 @@ export class SecondaryMenuComponent implements OnInit, OnDestroy {
         while (menuRoute.parent) {
           menuRoute = menuRoute.parent;
           parentMenuRoutes.push(menuRoute);
+          indentation++;
```
atruskie: Hmmmm... recursive loop

src/app/component/shared/secondary-menu/secondary-menu.component.ts
```diff
@@ -50,8 +51,26 @@ export class SecondaryMenuComponent implements OnInit, OnDestroy {
...
         }
 
+        // Update indentation values
+        if (current.order) {
+          current.order.indentation = indentation;
+        } else {
+          current.order = { indentation };
+        }
+
+        parentMenuRoutes.forEach(parent => {
```
atruskie: it isn't readily obvious why you have to set indentation on parent routes?

This loop has the potential to run many times...

atruskie (changes requested):
Okay. So the original reason that I very intentionally set the indentation to a user-controlled integer is that it is simple. The menu indentation is something that very rarely changes, and as an integer is super easy to fix. More than all of that, it is fast!

Making the indentation value dynamic means there are fewer chances for the value to be incorrect, so I'm willing to entertain this idea.

However, the current implementation has worse theoretical performance; it went from O(1) to O(2N).

So here's the solution I'd like you to try and get to:

When a menu item is created, it directly keeps a reference to its parent. When that happens it gets the parent's indentation level, adds one and stores the indentation value on itself. 
In this way, lookup of indentation values is O(1) (which is good because lookup happens many times).
The flaw here is if we need to change indentation value while the app is running - but that is not a requirement we have yet, and if we do need to we can recompute the chain once, rather than on every render.

src/app/interfaces/menusInterfaces.ts
Allcharles: I am unsure how to remove this, there is no way i can pass the indentation value of the menu item outside of the values specified in the interface. The only option i have is to turn all of the menu item types into classes which would be a bit of a rewrite.

Allcharles:
I have updated the system to try follow your ideas. Currently the major change i have made is that i have hijacked the functions you created to automatically set the kind value of the interfaces. Those functions now calculate the order and indentation of the item depending on a couple conditions. Ive highlighted the important bits below:

```javascript
export function MenuLink<T extends Omit<MenuLink, "kind">>(item: T): MenuLink {
  return Object.assign(item, {
    kind: "MenuLink" as "MenuLink",
    indentation: 0
  });
}
export function MenuRoute<T extends Omit<MenuRoute, "kind">>(
  item: T
): MenuRoute {
  return Object.assign(item, {
    kind: "MenuRoute" as "MenuRoute",
    indentation: item.parent ? item.parent.indentation + 1 : 0,
    order: item.parent ? item.parent.order : item.order
  });
}
export function MenuAction<T extends Omit<MenuAction, "kind">>(
  item: T
): MenuAction {
  return Object.assign(item, {
    kind: "MenuAction" as "MenuAction",
    indentation: 0
  });
}
```

The only concern i have about this method is that it relies on the programmer using the correct value when creating menus. There is no check currently which throws an error if a menu item is not using the function instead of the interface. A potential solution is to check if the `kind`, `order`, and `indentation` keys have been set inside the menu component. Anything without all three defined must be a mistake where someone has instantiated an interface instead of calling the function.

atruskie (approved):

PR #82, opened 2020-01-24
Initial Api Plumbing

### LINKED ISSUE #78: Remove plumbing from api services

Allcharles:
This is just an initial outlook on the solution to #78, DO NOT MERGE. Unit tests will fail spectacularly until I do a full rewrite.

Potential improvements:

  - Replace `allPath` and `idPath` with `CommonApiPaths` from previous solution. This would make it possible to define custom paths in the constructor instead of the special case method


Allcharles:
Need to update this so that collections return both a model and its paging details. This could be done either by created a custom model which stores both, or output them as an array/object. Creating a custom model would potentially allow me to not have to update all pre-existed code which would be affected by the changed output.

atruskie:
@Allcharles I've "reviewed" this in d3d9bc5. I had to look at it locally to play with types/APIs etc... and this happened. 

In d3d9bc5 there's a list of "canonical" types that are mostly correct. Most of the other files haven't been updated (so anywhere there is an error).

Also, the notion of a double rxjs observable chain has been removed. Unless we can demonstrate a need to transform an observable into a subject, then I have a feeling it is unnecessary. If it is necessary, then we can should be `pipe`ing and `map`ing these things together, not binding them together in a custom fashion.

package.json
```diff
@@ -102,7 +102,7 @@
     "ts-node": "^8.4.1",
     "tslint": "^5.20.1",
     "tslint-angular": "^3.0.2",
-    "typescript": "^3.7.4",
+    "typescript": "^3.8.0-dev.20200201",
```
Allcharles: Not a huge fan of using dev typescript versions. Currently 3.7.4 isn't even officially supported and is currently stopping us from upgrading the build-angular dependency. I'm also seeing a large list of dependencies complaining that they do not support this version of typescript.

src/app/helpers/stringTemplate/stringTemplate.ts
```diff
@@ -1,3 +1,5 @@
+import { type } from "os";
```
Allcharles: Unused import

src/app/services/baw-api/base-api.service.ts
```diff
@@ -35,12 +35,56 @@ export abstract class BawApiService {
...
+      return new _new(response.data);
+    };
+
+    this.handleResponse = (response: ApiResponse<T>): T[] => {
+      if (response.data instanceof Array) {
+        return response.data.map(x => new _new(x));
+      } else {
+        return [new _new(response.data)];
+      }
+    };
```
Allcharles: Not sure why these are implemented in the constructor

package.json
atruskie: Reverting to stable is fine

src/app/services/baw-api/base-api.service.ts
atruskie: Ah, that was a tricky bug. 

Reactive requires pure functions. Just pulling a function off of the service class (i.e. `this.handleResponse`) risks rebinding `this`. We need `this` though so we can get access to that `_new` property.

By assigning arrow functions to the values in the constructor, we close over `_new` instead and do not need to rely on `this`.

src/app/services/baw-api/api-common.ts
```diff
@@ -83,30 +91,46 @@ export interface ApiUpdate<M extends AbstractModel, P extends any[]> {
...
+    model: PartialWith<M, "id">,
+    ...urlParameters: P
+  ): Observable<M>;
+  abstract destroy(model: IdOr<M>, ...urlParameters: P): Observable<null>;
+}
+
+/**
+ * Api class with only the show route enabled
+ */
+export abstract class ShowApi<M extends AbstractModel, P extends any[]>
```
atruskie: So you probably don't need to create a child abstract class for every combination of API methods that may be implemented in a service. It was my intention that one would just implement the interface they needed.

The abstract class here is there really to just combine a whole wack of type signatures into a nice bundle.

Still, if you find this construct useful - go for it.

src/app/models/AudioRecording.ts
```diff
@@ -29,40 +29,32 @@ export interface AudioRecordingInterface {
...
+  public readonly siteId?: Id;
+  public readonly durationSeconds?: number;
+  public readonly sampleRateHertz?: number;
+  public readonly channels?: number;
+  public readonly bitRateBps?: number;
+  public readonly mediaType?: string;
+  public readonly dataLengthBytes?: number;
+  public readonly status?: "ready" | "uploading" | "corrupt";
+  public readonly createdAt?: DateTimeTimezone;
+  public readonly updatedAt?: DateTimeTimezone;
```
atruskie, line 47: This is a good change. 

I think generally we'll have more property getters and setters defined though.

Hopefully we can use those to convert types (like strings to dates) during serialization

src/app/models/Project.ts
```diff
@@ -15,68 +15,56 @@ import { AbstractModel } from "./AbstractModel";
...
-    this.id = project.id;
-    this.name = project.name;
     this.imageUrl =
       project.imageUrl || "/assets/images/project/project_span4.png";
-    this.creatorId = project.creatorId;
     this.createdAt = project.createdAt
       ? DateTime.fromISO(project.createdAt as string, {
           setZone: true
         })
       : defaultDateTimeTimezone;
```
atruskie, line 56: What purpose does `setZone` have?

src/app/services/baw-api/account.service.spec.ts
```diff
@@ -8,214 +8,378 @@ import { Subject } from "rxjs";
...
+  }
+
+  const shouldNotSucceed = () => {
+    fail("Service should not produce a data output");
+  };
+
+  const shouldNotFail = () => {
+    fail("Service should not produce an error");
+  };
+
```
atruskie: these seem sufficiently general enough to be moved to a common test helpers class

src/app/services/baw-api/account.service.spec.ts
```diff
@@ -8,214 +8,378 @@ import { Subject } from "rxjs";
...
-
-  it("should handle response with random id", fakeAsync(() => {
-    spyOn<any>(service, "show").and.callFake(
-      (path: string, filters: Filters) => {
-        expect(path).toBe("/user_accounts/5");
-        expect(filters).toEqual(null);
+      });
+    }
+
+    it("should handle empty response", fakeAsync(() => {
```
atruskie, line 81: why `fakeAsync`? why not just use the `async` keyword?

src/app/services/baw-api/account.service.spec.ts
```diff
@@ -8,214 +8,378 @@ import { Subject } from "rxjs";
...
+
+      createSuccess("/user_accounts/5", model);
+
+      service.show(5).subscribe((_model: User) => {
+        expect(_model).toBeTruthy();
+        expect(_model).toEqual(model);
+      }, shouldNotFail);
+
+      tick(100);
+    }));
```
atruskie, line 341: these tests seem redundant - what are they testing?

src/app/services/baw-api/account.service.spec.ts
```diff
@@ -8,214 +8,378 @@ import { Subject } from "rxjs";
...
-    );
-
-    tick(100);
-  }));
+        expect(err).toEqual(errorInfoResponse);
+      });
+
+      tick(100);
+    }));
+  });
```
atruskie, line 364: Generally, this test file is a lot simpler. Well done.

I'm not sure we need to test every combination you're testing here, especially for every service. But for now, good as is.

src/app/services/baw-api/base-api.service.spec.ts
```diff
@@ -1,68 +1,855 @@
...
+      {
+        functionName: "httpDelete",
+        method: "DELETE",
+        hasBody: false
+      }
+    ];
+
+    httpMethods.forEach(httpMethod => {
+      describe(httpMethod.functionName, () => {
+        it("should create patch request", () => {
```
atruskie:
```suggestion
        it(`should create ${ httpMethod } request`, () => {
```

src/app/services/baw-api/base-api.service.spec.ts
```diff
@@ -1,68 +1,855 @@
...
+          );
+
+          const req = catchRequest("/broken_link", httpMethod.method);
+          flushResponse<MockModel[]>(req, response);
+        });
+
+        it("should handle error response", () => {
+          const response = { meta: errorMeta, data: null } as ApiResponse<
+            MockModel
+          >;
```
atruskie: you're line length linting rules seem a bit aggressive

src/app/services/baw-api/base-api.service.ts
```diff
@@ -36,56 +37,66 @@ export abstract class BawApiService<T extends AbstractModel> {
...
+  }
 
   constructor(
     protected http: HttpClient,
-    private apiRoot: string,
-    _new: new (_: object) => T) {
+    config: AppConfigService,
+    @Inject(STUB_CLASS_BUILDER) classBuilder: new (_: object) => T
+  ) {
+    this.apiRoot = config.getConfig().environment.apiRoot;
```
atruskie: TODO: remove later

src/app/services/baw-api/base-api.service.ts
```diff
@@ -36,56 +37,66 @@ export abstract class BawApiService<T extends AbstractModel> {
...
+    // Create pure functions to prevent rebinding of 'this'
+
+    this.handleCollectionResponse = (response: ApiResponse<T>): T[] => {
       if (response.data instanceof Array) {
-        throw new Error("Received an array of API results when only a single result was expected");
+        return response.data.map(
+          model => new classBuilder(this.addMetadata(model, response.meta))
+        );
+      } else {
+        return [this.handleSingleResponse(response)];
```
atruskie: inline this. This is so simple we don't need to abstract into handle single response

src/app/services/baw-api/base-api.service.ts
```diff
@@ -36,56 +37,66 @@ export abstract class BawApiService<T extends AbstractModel> {
...
+    this.apiRoot = config.getConfig().environment.apiRoot;
 
-    this.handleSingleResponse = (response: ApiResponse<T>) => {
+    // Create pure functions to prevent rebinding of 'this'
+
+    this.handleCollectionResponse = (response: ApiResponse<T>): T[] => {
       if (response.data instanceof Array) {
-        throw new Error("Received an array of API results when only a single result was expected");
+        return response.data.map(
+          model => new classBuilder(this.addMetadata(model, response.meta))
```
atruskie: I would tag with metadata after the new class has been constructed.

Object.assign semantics may make this method work, but it is clearer if the final type get tagged with metadata.

Have you considered adding a method on `AbstractModel` that deals with metadata?

src/app/services/baw-api/base-api.service.ts
```diff
@@ -36,56 +37,66 @@ export abstract class BawApiService<T extends AbstractModel> {
...
   }
 
+  /**
+   * Retrieve metadata from model
+   * @param model Model
+   */
+  public retrieveMetadata(model: AbstractModel): Meta {
+    return model[this.metaKey];
+  }
+
```
atruskie: move implementation to AbstractModel

src/app/services/baw-api/base-api.service.ts
```diff
@@ -197,18 +239,26 @@ export abstract class BawApiService<T extends AbstractModel> {
...
   }
 
+  /**
+   * Append metadata to model data
+   * @param model Model
+   * @param meta Metadata
+   */
+  private addMetadata(model: object, meta: Meta): object {
+    return { ...model, [this.metaKey]: meta };
+  }
```
atruskie: move this to AbstractModel

src/app/services/baw-api/mock/projectsMockService.ts
```diff
@@ -1,29 +1,39 @@
...
+import { AppConfigService } from "../../app-config/app-config.service";
+import { StandardApi } from "../api-common";
 
 @Injectable({
   providedIn: "root"
 })
-export class MockProjectsService extends MockApiCommon<Project> {
-  public getProject() {
-    return new Subject();
+export class MockProjectsService extends StandardApi<Project, []> {
```
atruskie: I wonder if there is a more general way to create mock services this?

It seems like a lot of work to create a mock service for every class.

src/app/services/baw-api/projects.service.spec.ts
```diff
@@ -7,24 +7,70 @@ import { RouterTestingModule } from "@angular/router/testing";
...
+  }
+
+  const shouldNotSucceed = () => {
+    fail("Service should not produce a data output");
+  };
+
+  const shouldNotFail = () => {
+    fail("Service should not produce an error");
+  };
+
```
atruskie: see the previous comment about abstracting out common test helpers

atruskie (approved):
Generally: the destroy action may or may not return the object that was just destroyed. So change all `destroy` return types from `null` to `M | null`.

A very good set of changes. A lot of the unit test code is easier to read and understand. 💯

We're at the point where you have enough direction, so go ahead and merge in when you've made changes and continue the bulk work on master.

src/app/models/Project.ts
Allcharles: setZone makes the luxon date object accept accept timezone data. By default, luxon and javascript dates ignore timezones

src/app/services/baw-api/account.service.spec.ts
Allcharles: I prefer `fakeAsync` because it has all of the functionality of `async` with additional control. `async` actually waits for the timer to finish, whereas `fakeAsync` allows me to kill the timer making tests run faster.

src/app/services/baw-api/account.service.spec.ts
Allcharles: The difference is that "should handle response with random user model id" uses the User model to get the id, whereas "should handle response with random user id" passes the id directly. Its mostly a check that the id is being read correctly from both types of inputs using an id value which is not 1 (which i use as a standard placeholder id)

src/app/models/Project.ts
atruskie: Ah goodo

src/app/services/baw-api/account.service.spec.ts
atruskie: Ah fair enough.

src/app/services/baw-api/base-api.service.spec.ts
Allcharles: Took a crack at fixing it, Prettier (my formatter) has options to expand the line length but its ignoring whatever i set to ill have to fix it another time

src/app/services/baw-api/base-api.service.ts
Allcharles: I believe you abstracted this to prevent issues with the value of `this` being overwritten

atruskie: No, I mean literally that line:
```suggestion
        return [new classBuilder(this.addMetadata(model, response.meta)];
```

PR #86, opened 2020-01-28
Login redirect

### LINKED ISSUE #85: Login page redirection
Login page should handle page redirection on successful login. If the url contains a redirect path (within in the ecosounds website domain), it should use the redirect instead of the default link back to the home page.

Allcharles:
Added login page redirection.

## Version One

This allows other components to direct the user to the login page with an optional redirect parameter that will send the user back on successful completion. External routes which do not use the ecosounds.org domain (staging.ecosounds.org during staging) will be ignored to prevent attacks.

Example use with internal route:

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  [queryParams]="{ redirect: '/link' }"
  >
    Login
</a>
```

Example use with external route:

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  [queryParams]="{ redirect: 'https://staging.ecosounds.org/projects' }"
  >
    Login
</a>
```

Example use with go back:

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  [queryParams]="{ redirect: true }"
  >
    Login
</a>
```

 ## Version 2

The login page now redirects to the previous page by default (within the angular app). If it is unable to determine a redirection path, it defaults to the home page. It also maintains the ability to redirect both to an external or internal path depending on the queryParam.

Default (redirects to previous page on submission):

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  >
    Login
</a>
```

Example use with internal route:

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  [queryParams]="{ redirect: '/link' }"
  >
    Login
</a>
```

Example use with external route:

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  [queryParams]="{ redirect: 'https://staging.ecosounds.org/projects' }"
  >
    Login
</a>
```

Example use without redirect path:

```html
<a
  class="nav-link"
  [routerLink]="routes.login.route.toString()"
  [queryParams]="{ redirect: false }"
  >
    Login
</a>
```

Closes #85


atruskie:
Can we rethink this?

Here is the general flow:

1. a user is navigated to a new route
2. The route, tied to a component, and a server resource attempt to load but fails because of a HTTP 401
3. the router redirects us to a login page
4. the user signs in
5. the user is redirected back to the original resource

We agree on the top part.

Your solution requires devs to opt-in on each (every?) link whether they want redirect behaviour?
Your current solution is good for external links for which we cannot control logins/routing/redirects.

Does the following behaviour also exist?:

I'd prefer redirect was on by default. Unless `redirect=false` the redirect link should be set automatically to whichever route failed loading its data (the `401`). The router, when redirecting to the login page should append the `redirect` URL as a QSP to the login page route.

You maY ask: why don't we trust the fact that we have a user token in local storage (which should mean we are logged in)? That token can expire (lol: should) or be revoked at any time on the server.  We may not be logged in after all.

atruskie (changes requested):
See previous comment

atruskie (approved):

PR #87, opened 2020-01-29
Some minor tweaks to menu system

atruskie:
Just removed a few functions that seemed redundant from menu.component.ts

### DIFF
--- a/src/app/component/shared/menu/menu.component.ts
+++ b/src/app/component/shared/menu/menu.component.ts
@@ -99,6 +99,19 @@ export class MenuComponent implements OnInit, OnDestroy {
   }
 
   /**
+   * Calculate the indentation of a secondary link item
+   * @param link Link to calculate padding for
+   */
+  calculateIndentation(link: AnyMenuItem) {
+      // Only the secondary menu implements this option
+      if (this.menuType !== "secondary" || !link.indentation) {
+        return 0;
+      }
+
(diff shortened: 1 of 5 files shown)


src/app/component/shared/menu/menu.component.ts
```diff
@@ -63,13 +63,10 @@ export class MenuComponent implements OnInit, OnDestroy {
...
-      this.links
-        ? this.menuType === "secondary"
-          ? this.links.filter(link => this.filter(link)).sort(this.compare)
-          : this.links.filter(link => this.filter(link))
-        : List<AnyMenuItem>([])
-    );
+    this.filteredLinks = new Set(this
+      .links
+      .filter(this.filter)
+      .sort(this.compare));
```
Allcharles: Converting the list to a set does not remove duplicates, do we not care about them? Although I do agree that the filtering method could be simplified and improved. Currently it only matches based on label, however this should instead match based on if the label, order, and indentation match.

Allcharles: Sorting the action menu may be a little wonky. Could be improved by implementing the ordering system from the secondary menu.

atruskie: Since we export and reuse all menu items, they are unique in terms of object identity. In this case, the set uniqueness does work properly.

PR #92, opened 2020-02-03
Project edit permissions

Allcharles:
# Project Edit Permissions Page

This will in future allow users to update the user permissions for a project they are the owner of. The next steps for this component is link it to a service and have the permission buttons update permissions. However creating the service functions required is blocked by PR #82 so this will be done in a separate PR.

## Changes

- Created project edit permissions page (`/projects/:id/permissions`)
- Created selectable-item component
- Updated page layout padding
- Updated project/site pages to contain PermissionsShield Widget displaying user permissions

## Issues:

- Couldn't add tooltips to table headers
- Did not add unit tests to EditPermissionsComponent (because no real logic to test)
- Did not add unit tests to SelectabledItemComponent (not sure if the layout is generic enough yet)

## Visual Changes

image


src/app/component/projects/pages/permissions/permissions.component.ts
```diff
@@ -0,0 +1,173 @@
...
+    users: "None" | "Reader" | "Owner";
+    overall: "None" | "Reader" | "Owner";
+  }[];
+  public rows: {
+    user: string;
+    individual: string;
+    visitors: "None" | "Reader" | "Owner";
+    users: "None" | "Reader" | "Owner";
+    overall: "None" | "Reader" | "Owner";
+  }[];
```
atruskie: I'm not sure what the purpose of these properties are but the permission levels seem to be omitting writer?

src/styles.scss
```diff
@@ -38,6 +38,12 @@ section {
 }
 
 /* BACKWARDS COMPATIBILITY */
+.btn-default {
```
atruskie, line 44: I'd like if anything in backwards compat was given a description of the effect it has and the reason it was added - that we know when we can remove it later

atruskie (approved):
Looks good.

We'll leave any client-side logic nice and thin until we get the associated route going.

PR #93, opened 2020-02-03
Project assign sites

Allcharles:
# Project Assign Sites Page

This will, in future, allow admin users to update the sites associated with a project. The next steps for the component is to move it to the admin dashboard, and link the actions to a service. This PR is non-functional.

Blocked by PR #92 

## Changes

- Created project assign sites page (`/projects/:id/assign`)

## Issues

- Did not add unit tests to AssignSitesComponent (because no real logic to test)

## Visual Changes

image


src/app/component/projects/pages/assign/assign.component.html
```diff
@@ -0,0 +1,49 @@
...
+      [headerHeight]="50"
+      [limit]="25"
+      [reorderable]="false"
+      [rowHeight]="'auto'"
+      [rows]="rows"
+      [scrollbarH]="true"
+      [selectAllRowsOnPage]="false"
+      [selected]="selected"
+      [selectionType]="SelectionType.checkbox"
+      (select)="onSelect($event)"
```
atruskie, line 29: Is it possible to move these attributes to the controller code? For so many options it would seem neater to just `...options` rather than do this...

src/app/component/projects/pages/assign/assign.component.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  ngOnInit() {
+    this.ready = false;
+
+    this.route.params
+      .pipe(
+        flatMap(params => {
+          return this.projectsApi.getProject(params.projectId);
+        }),
+        takeUntil(this.unsubscribe)
```
atruskie, line 66: I've noticed this pattern previously where you inject the service into the component controller and not the object.

e.g. rather than

```
  constructor(
  )
```

we should be doing:

```
  constructor(
    private project: Project
  )
```

The angular router has the ability to load resources for routes and inject them straight into the component. We should definitely take advantage of this because the app behaves nicer (concurrent loading of resources mean the page renders quicker) and there's less boilerplate code in the components.

Of course, any additional API resources needed on the page will still need services injected. That's a different matter though.

This is not a blocking comment for this PR.

atruskie (approved):
This will need more work in the future but it is designed as well as it can be until you're given a service to play with.

src/app/component/projects/pages/assign/assign.component.ts
Allcharles: Created an issue: https://github.com/QutEcoacoustics/workbench-client/issues/99

src/app/component/projects/pages/assign/assign.component.html
Allcharles: The most angular method for this would be to create a wrapper component which replicates the inputs from `ngx-datatable` and sets defaults for the inputs not provided. I intend to do this once I've built enough tables to know a) what the defaults are, and b) link up the tables to changing data so I can test updating the tables. For now I've created an abstract class which defines a list of defaults to reduce the amount of boilerplate and magic numbers.

atruskie: This is basically what I'm referring too: https://github.com/angular/angular/issues/14545
Looks like it is not a thing 🤷‍♂️

Don't over nest components - it just creates more work for browser and angular.

Maybe a custom directive would be better, like `ngx-standard-options` that does the binding that way.

PR #96, opened 2020-02-13
Extract Environment

### LINKED ISSUE #90: Create external environment file
Create an external environment file which can be placed next to the compiled production file containing the following information:

- `appConfig`

This should be designed in such a way that the environment file can be inserted by the CI, and/or changed by an admin. The use case for this is switching the system between staging and production modes. Relates to #83.

### LINKED ISSUE #91: Remove testing files from production build
Remove environment test files (`assets/tests`) are being kept in the production bundle. These should be removed either by changing their stored location, making the bundler ignore them, or removed by the CI.

Allcharles:
Extracted the environment variables from the application for the inevitable push to deployment.

This should allow a build environment to easily select an environment to push with the build to switch it between development, staging, and production. Currently this environment file only stores the external configs location on the BAW Server (currently redirects to workbench-client as route does not exist).

With this update there are three types of environments:

- Build environment
  - This is baked into the application and cannot be changed after build
- Deployment environment
  - This is the environment which can be changed after build time. This stores information about whether the system is accessing the staging or production server
- Config
  - This is the downloaded config file from the BAW Server. It stores all values that should change between staging and production versions of the application.

Unit tests are a little skimp as they relate to Issue #97 

Closes #90 
Closes #91


src/app/app.helper.ts
```diff
@@ -3,20 +3,15 @@ import { APP_INITIALIZER } from "@angular/core";
...
 import {
-  APP_CONFIG,
   AppConfigService,
   appInitializerFn
 } from "./services/app-config/app-config.service";
-import {
-  APP_CONFIG as MOCK_APP_CONFIG,
-  MockAppConfigService
-} from "./services/app-config/appConfigMockService";
+import { MockAppConfigService } from "./services/app-config/appConfigMock.service";
```
atruskie: I guess I'll find out what `MockAppConfig` is below... but the name is suspicious. Mocks should only be used in tests...

src/app/services/app-config/app-config.service.ts
```diff
@@ -13,53 +14,41 @@ export class AppConfigService {
...
-      },
-      () => {
-        this.appConfig = null;
-        return null;
+
+    const handleData = (data: Configuration) => {
+      if (data.environment.environment !== environment.environment) {
+        throw new Error(
+          `AppConfigService: Deployment environment (${environment.environment}) and config environment (${data.environment.environment}) do not match.`
+        );
```
atruskie: I'm not sure in which case this condition would occur. Even if it does, why is it an error?

src/app/services/app-config/app-config.service.ts
```diff
@@ -93,11 +82,17 @@ export class AppConfigService {
...
 ) {
-  return await fetch(config)
+  return await fetch(environmentUrl)
+    .then(response => response.json())
+    .then(options => {
+      environment.configUrl = options.configUrl;
+      environment.environment = options.environment;
+      return options.configUrl;
+    })
+    .then((configUrl: string) => fetch(configUrl))
```
atruskie: Ok officially confused. Why are we fetching two things from the server?

src/app/app.helper.ts
atruskie: Yeah, it is a test feature. Code in test files, even if just a named constant, should never bleed back into source code.

Try inverting it:
- Define an APP_CONFIG symbol in app-config and override the injection of that symbol in the tests only.

atruskie (changes requested):
Okay.

Delete the following files:

- src/assets/config/development.json
- src/assets/config/production.json
- src/assets/config/staging.json
- src/assets/config/
- src/assets/environment/environment.json
- src/assets/environment/environment.production.json
- src/assets/environment/environment.staging.json

Now add: 

- src/assets/environment.json

The contents of which will have the structure of the  `Configuration` interface.

Then, simplify the rest of this code.

src/app/app.helper.ts
Allcharles: `APP_CONFIG` is no longer used, i have however taking all of the testing code out of `app.helper.ts` and into `test.helper.ts`.

atruskie (approved):

PR #101, opened 2020-02-14
Angular9

Allcharles:
Updated workbench-client to Angular 9.0.0

Contains the following:

- Minor rewrite of action menu component unit tests
- Minor rewrite of secondary menu component unit tests
- Updated all dependencies to latest version
- Fixed component imports causing test suite to crash
  - Some components imported `pageComponent` and `pageDecorator` with a `.js` causing the test suite to be unable to find the file


tsconfig.json
```diff
@@ -21,7 +21,6 @@
   },
   "angularCompilerOptions": {
     "fullTemplateTypeCheck": true,
-    "strictInjectionParameters": true,
-    "enableIvy": true
+    "strictInjectionParameters": true
```
atruskie, line 24: Is Ivy on by default now?

atruskie (approved):

tsconfig.json
Allcharles: Yes

Issue #104, opened 2020-02-18
User submission handling


Allcharles:
Currently, if a user submits a request to the server and immediately leaves or navigates from the page before the request is complete, it will fail silently. This can be solved through the use of CanDeativate guards: https://angular.io/guide/router#candeactivate-handling-unsaved-changes. Commit https://github.com/QutEcoacoustics/workbench-client/commit/b7197c71cba481aa084809b46ae4a0001afd5abf provides a rudimentary method of solving this issue, however a more specific solution using the dialog service should be implemented to give users more information about why they cannot leave/navigate away from the page.

Allcharles:
Relevant link: https://hackernoon.com/harnessing-the-power-of-mixins-in-angular-faox230fk

PR #105, opened 2020-02-18
Model Injection

### LINKED ISSUE #94: Inject api root into ApiServices
Explore the possibility of restricting the injected config values for the services.
Currently the inputs for the BawApiService are the following:

```javascript
constructor(
  protected http: HttpClient,
  config: AppConfigService,
  @Inject(STUB_CLASS_BUILDER) classBuilder: new (_: object) => T
) {
```

This should instead be the following:

```javascript
constructor(
  protected http: HttpClient,
  @Inject(STUB_API_ROOT) apiRoot: string,
  @Inject(STUB_CLASS_BUILDER) classBuilder: new (_: object) => T
) {
```

May be possible through the use of lazy loading the apiRoot.

### LINKED ISSUE #99: Inject api models into components
The angular router has the capability to load resources for routes, and inject them into components. This should enable the component to set the api model it requires as an input to the constructor instead of the service.

Example:

Current implementation:

```typescript
  constructor(
  )
```

Optimal result:

```typescript
  constructor(
    private project: Project
  )
```

Relates to this comment: https://github.com/QutEcoacoustics/workbench-client/pull/93#discussion_r376673658
May relate to this issue: #94

Allcharles:
# Component/Service Model Injection

This PR deals with two concepts: injecting a model into a component, and replace `appConfigService` with apiRoot for service.

Currently this is mid-development, just looking for comments on whether the patterns used are good. DO NOT MERGE.

## Changes

- Replaced `config: AppConfigService` with `apiRoot: string` in baw api services [MISSING BAW-API-INTERCEPTOR]
- Replaced references to api services in components which only needed to require requesting a model/s [NOT IMPLEMENTED YET]
- Added `canDeativate` guard to prevent users from redirecting away from page until a request to the server completes (relates to #104)

## Closes

Closes #99 
Closes #94


src/app/app.helper.ts
```diff
@@ -73,5 +97,35 @@ export const providers = [
...
+  },
+  {
+    provide: ShallowSitesService,
+    useFactory: shallowSitesServiceFactory,
+    deps: bawApiDeps
+  },
+  {
+    provide: UserService,
+    useFactory: userServiceFactory,
+    deps: bawApiDeps
```
atruskie, line 129: So this is heading the wrong way: we don't want a custom factory per service. We want to `@Inject` with a custom token in the services

src/app/component/projects/pages/details/details.component.ts
```diff
@@ -49,58 +56,39 @@ export const projectMenuItemActions = [
...
-          this.project = project;
-          this.ready = true;
-        },
-        (err: ApiErrorDetails) => {
-          this.error = err;
+    this.route.data.pipe(takeUntil(this.unsubscribe)).subscribe(
+      (data: {
+        project: Project | ApiErrorDetails;
+        sites: Site[] | ApiErrorDetails;
+      }) => {
```
atruskie, line 76: I think you can use generics to type this

atruskie, line 76: We should never reach this code if the API returned an error. Right? Doesn't the route fail to load in this case?

src/app/component/projects/pages/details/details.component.ts
```diff
@@ -49,58 +56,39 @@ export const projectMenuItemActions = [
...
+        if (isApiErrorDetails(data.project)) {
+          this.error = data.project;
+          return;
```
atruskie, line 79: I'm not fond of the duality here between `ApiErrorDetails` and `TModel`.

Isn't it clearer to return `ApiResponse<Project>`? and have an `isApiError` type guard for errors?

src/app/component/projects/pages/list/list.component.ts
```diff
@@ -42,41 +49,35 @@ export const projectsMenuItemActions = [
...
   error: ApiErrorDetails;
 
-  constructor(private api: ProjectsService) {
+  constructor(private route: ActivatedRoute) {
     super();
   }
 
   ngOnInit() {
-    this.loading = true;
+    this.ready = false;
```
atruskie, line 67: What is this used for?

If we're not ready then `this.cardList` will be empty/null. Why do we need a separate ready flag?

src/app/component/projects/pages/list/list.component.ts
```diff
@@ -42,41 +49,35 @@ export const projectsMenuItemActions = [
...
+        if (isApiErrorDetails(data.projects)) {
+          this.error = data.projects;
+          return;
         }
-      );
+
+        this.cardList = List(data.projects.map(project => project.getCard()));
+        this.ready = true;
+      },
+      err => {}
```
atruskie, line 79: silent error

src/app/component/security/pages/confirm-account/confirm-account.component.ts
```diff
@@ -23,6 +23,7 @@ import data from "./confirm-account.json";
     ]),
     links: List()
   },
+  canDeactivate: true,
```
atruskie, line 26: Why does confirm account need a canDeactivate guard?

It looks like you've stuck these in almost all components. Maybe we should give a better name to the property. What does it do actually? Prevent navigation on unsaved data?

`verifyFormSaved`?

src/app/component/security/pages/unlock-account/unlock-account.component.ts
```diff
@@ -23,6 +23,7 @@ import data from "./unlock-account.json";
     ]),
     links: List()
   },
+  canDeactivate: true,
```
atruskie, line 26: If we really want this on every component maybe it should be set as a default true and allow any component that doesn't need it to opt out

src/app/guards/can-deactivate/can-deactivate.guard.ts
```diff
@@ -0,0 +1,17 @@
...
+  canDeactivate: () => Observable<boolean> | Promise<boolean> | boolean;
+}
+
+@Injectable({
+  providedIn: "root"
+})
+export class CanDeactivateGuard
+  implements CanDeactivate<CanComponentDeactivate> {
+  canDeactivate(component: CanComponentDeactivate) {
+    return component.canDeactivate ? component.canDeactivate() : true;
```
atruskie, line 15: oh, I see, it is a callback. Definitely define as a default then in page component.

src/app/helpers/page/pageInfo.ts
```diff
@@ -26,6 +28,8 @@ export class PageInfo implements PageInfoInterface {
   category: Category;
   menus: Menus;
   fullscreen: boolean;
+  canDeactivate: boolean;
+  resolvers: { [key: string]: any };
```
atruskie, line 32: What do resolvers do?

atruskie: Ah I see, they are passed to route resolvers. Add some docs then?

src/app/services/app-config/app-config.service.ts
```diff
@@ -25,18 +26,19 @@ export class AppConfigService {
...
-      console.error("AppConfigService: ", err);
-      throw new Error("AppConfigService: Failed to load configuration file");
-    };
-    return retrieveAppConfig(handleData, handleError);
+    return await fetch(environmentUrl)
+      .then(response => response.json())
+      .then((data: Configuration) => {
+        this.appConfig = data;
+        this.titleService.setTitle(data.values.brand.name);
+        environment.googleMapsKey = this.appConfig.values.keys.googleMaps;
```
atruskie, line 34: why is the googleMapsKey special?

src/app/services/baw-api/account.service.ts
```diff
@@ -13,17 +13,22 @@ import {
...
+  config: AppConfigService
+) => {
+  return new AccountService(http, config.getConfig().environment.apiRoot);
+};
+
+@Injectable()
 export class AccountService extends ReadonlyApi<User, []> {
-  constructor(http: HttpClient, config: AppConfigService) {
-    super(http, config, User);
+  constructor(http: HttpClient, @Inject(STUB_API_ROOT) apiRoot: string) {
```
atruskie, line 30: Inject token: perfect

src/app/services/baw-api/account.service.ts
```diff
@@ -13,17 +13,22 @@ import {
...
-@Injectable({
-  providedIn: "root"
-})
+export const accountServiceFactory = (
+  http: HttpClient,
+  config: AppConfigService
+) => {
+  return new AccountService(http, config.getConfig().environment.apiRoot);
+};
+
```
atruskie, line 27: factory: not great. remove factory. Same for all following services

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -19,6 +18,7 @@ export const apiReturnCodes = {
   internalServerFailure: 500
 };
 
+export let STUB_API_ROOT = new InjectionToken("test.api.root");
 export let STUB_CLASS_BUILDER = new InjectionToken("test.class.builder");
```
atruskie, line 22: remove "test" from these injection tokens. They can be used for more than just testing cases - in fact, API root token should be used for prod code cases

maybe: "baw.api.root" and "baw.models.builder"

atruskie: Also, it seems like API root token should be in another file. Probably app config.

src/app/services/baw-api/projects.service.ts
```diff
@@ -13,20 +13,22 @@ import {
...
-@Injectable({
-  providedIn: "root"
-})
+export const projectsServiceFactory = (
+  http: HttpClient,
+  config: AppConfigService
+) => {
+  return new ProjectsService(http, config.getConfig().environment.apiRoot);
+};
+
```
atruskie, line 27: as mentioned before, let's avoid the factories. And so on for all services...

src/app/services/baw-api/resolvers/resolver-common.ts
```diff
@@ -0,0 +1,44 @@
...
+} from "@angular/router";
+import { Observable, of } from "rxjs";
+import { catchError, take } from "rxjs/operators";
+import { Id } from "src/app/interfaces/apiInterfaces";
+import { AbstractModel } from "src/app/models/AbstractModel";
+import { ApiErrorDetails } from "../api.interceptor.service";
+
+export class ListResolver<T extends AbstractModel>
+  implements Resolve<T[] | ApiErrorDetails> {
+  constructor(private api: any, private getIds: (params: ParamMap) => Id[]) {}
```
atruskie, line 16: We could be more specific about the type of `api` here

src/app/services/baw-api/resolvers/projects-resolver.service.ts
```diff
@@ -0,0 +1,24 @@
...
+@Injectable({
+  providedIn: "root"
+})
+export class ProjectResolverService extends ShowResolver<Project> {
+  constructor(api: ProjectsService) {
+    super(api, params => {
+      return [parseInt(params.get("projectId"), 10)];
+    });
+  }
+}
```
atruskie, line 24: this seems like a lot of ceremony. I wonder if we can get rid of these derived resolver lasses. All that is really needed is a 

`ListResolver` or `ShowResolver` (both perfectly generic) and optionally a callback for customisation.

src/environments/environment.prod.ts
```diff
@@ -2,5 +2,8 @@ import { version } from "../../package.json";
 
 export const environment = {
   production: true,
-  version
+  version,
+
+  // This will be updated by the appConfig service
+  googleMapsKey: ""
```
atruskie, line 8: why? Why not just inject AppConfigService where needed?

atruskie (changes requested):
canDeactivate worked example: https://hackernoon.com/harnessing-the-power-of-mixins-in-angular-faox230fk

I wonder if we can simplify baw-api with a mixin model for injecting services ala: https://github.com/angular/components/blob/3acd6c17197d6526870ea97dba09073b8c86f101/src/material/core/common-behaviors/color.ts#L30

https://codeburst.io/having-fun-with-mixins-in-angular-719f1ec83aeb

A fair few architectural changes needed. We should meet to discuss if anything is unclear.

PR #107, opened 2020-02-19
Strong route parameter failure

### LINKED ISSUE #12: StrongRoute failing on mixed normal and parameter children for same parent
StrongRoute bug affecting multiple components which append to the same parent StrongRoute. If multiple routes are attached to a StrongRoute, the parameter (eg. /:siteId) route must be last. Otherwise any routes appended after it will not be added to the created routes, although its route will exist and will redirect to the parameter route component.

Currently I am unsure if this is an issue for multiple parameter routes connected to the same parent as this scenario currently does not exist.

Example of a failing `menus.ts` file:

``` javascript
export const siteMenuItem = MenuRoute({
  icon: ["fas", "map-marker-alt"],
  label: "Site",
  route: sitesRoute.add(":siteId"),
  tooltip: () => "The current site",
  parent: projectMenuItem,
  order: {
    priority: projectMenuItem.order.priority,
    indentation: projectMenuItem.order.indentation + 1
  }
});

export const newSiteMenuItem = MenuRoute({
  icon: ["fas", "plus"],
  label: "New site",
  route: sitesRoute.add("new"),
  tooltip: () => "Create a new site",
  predicate: user => !!user,
  parent: projectMenuItem,
  order: {
    priority: projectMenuItem.order.priority,
    indentation: projectMenuItem.order.indentation + 1
  }
});
```

And a functional `menus.ts` file:

``` javascript
export const newSiteMenuItem = MenuRoute({
  icon: ["fas", "plus"],
  label: "New site",
  route: sitesRoute.add("new"),
  tooltip: () => "Create a new site",
  predicate: user => !!user,
  parent: projectMenuItem,
  order: {
    priority: projectMenuItem.order.priority,
    indentation: projectMenuItem.order.indentation + 1
  }
});

export const siteMenuItem = MenuRoute({
  icon: ["fas", "map-marker-alt"],
  label: "Site",
  route: sitesRoute.add(":siteId"),
  tooltip: () => "The current site",
  parent: projectMenuItem,
  order: {
    priority: projectMenuItem.order.priority,
    indentation: projectMenuItem.order.indentation + 1
  }
});
```

Allcharles:
# Fixing Bug in Strong Route

PR fixes bug where the order in which routes are added to the StrongRoute class can cause them to fail. Specifically any routes declared after a parameter route (eg `/:id`) would fail to load. 

## Changes

- Added StrongRoute unit tests
- Added sorting to StrongRoute paths

## Issues

- `routeConfig` function unit tests not written as the function is not used
- Noticed that StrongRoute children will also add their parent routes to the router. This may cause an issue later in development, however for now it has no adverse affects

## Closes

Closes #12

### DIFF
--- a/src/app/interfaces/strongRoute.spec.ts
+++ b/src/app/interfaces/strongRoute.spec.ts
@@ -1,0 +1,58 @@
+describe("StrongRoute", () => {
+  describe("Base StrongRoute", () => {
+    it("should create base route", () => {});
+    it("should add route", () => {});
+    it("should add multiple routes", () => {});
+    it("should add parameter route", () => {});
+    it("should add mixed routes", () => {});
+  });
+
+  describe("Child StrongRoute", () => {
+    it("should add route", () => {});
+    it("should add multiple routes", () => {});
+    it("should add parameter route", () => {});
+    it("should add mixed routes", () => {});
+  });
+
+  describe("Sorting", () => {
+    it("should order '\"\"' routes first", () => {});
+    it("should order parameter route last", () => {});
+  });
+
+  describe("Formatting", () => {
+    it("should format base route", () => {});
+    it("should format child route", () => {});
+    it("should format parameter route", () => {});
+    it("should format child parameter route", () => {});
+  });
+
+  describe("Compile Routes", () => {
+    it("should compile base route with route", () => {});
+    it("should compile base route with parameter route", () => {});
+    it("should compile base route with mixed routes", () => {});
+    it("should compile child StrongRoute with route", () => {});
+    it("should compile child StrongRoute with parameter route", () => {});
+    it("should compile child StrongRoute with mixed routes", () => {});
+  });
+
+  describe("toString", () => {
+    it("should handle base StrongRoute", () => {});
+    it("should handle child StrongRoute", () => {});
+    it("should handle grandchild StrongRoute", () => {});
+    it("should handle parameter StrongRoute", () => {});
+  });
+
+  describe("toRoute", () => {
+    it("should handle base StrongRoute", () => {});
+    it("should handle child StrongRoute", () => {});
+    it("should handle grandchild StrongRoute", () => {});
+    it("should handle parameter StrongRoute", () => {});
+  });
+
+  describe("routeConfig", () => {
+    it("should handle base StrongRoute", () => {});
+    it("should handle child StrongRoute", () => {});
+    it("should handle grandchild StrongRoute", () => {});
+    it("should handle parameter StrongRoute", () => {});
+  });
+});
--- a/src/app/interfaces/strongRoute.spec.ts
+++ b/src/app/interfaces/strongRoute.spec.ts
@@ -1,3 +1,5 @@
+import { Type } from "@angular/core";
+import { Route, Routes } from "@angular/router";
 import { StrongRoute } from "./strongRoute";
 
 describe("StrongRoute", () => {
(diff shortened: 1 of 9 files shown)


src/app/interfaces/strongRoute.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+describe("StrongRoute", () => {
+  describe("Base StrongRoute", () => {
+    let strongRoute: StrongRoute;
+
+    beforeEach(() => {
+      strongRoute = StrongRoute.Base;
+    });
+
+    it("should create base route", () => {
+      expect(strongRoute).toBeTruthy();
```
atruskie: OK, all the `toBeTruthy` tests here are a code smell. Is there nothing better worth testing?

src/app/interfaces/strongRoute.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+
+    it("should add route", () => {
+      const homeRoute = strongRoute.add("home");
+      expect(homeRoute).toBeTruthy();
+    });
+
+    it("should add multiple routes", () => {
+      const homeRoute = strongRoute.add("home");
+      const houseRoute = strongRoute.add("house");
+      expect(homeRoute).toBeTruthy();
```
atruskie: are we just testing an exception is not thrown?

If you only want to test nothing then use `expect().nothing()`

I would test that the route is actually added to the strong route base

src/app/interfaces/strongRoute.ts
```diff
@@ -113,22 +116,36 @@ export class StrongRoute {
    */
   compileRoutes(callback: RouteConfigCallback): Routes {
     const rootRoute = this.root;
+
+    const sortRoutes = (a: Route, b: Route): -1 | 0 | 1 => {
+      // If one of the routes is a parameter route
+      if (a.path.charAt(0) === ":" || b.path.startsWith(":")) {
```
atruskie: why use two different methods to test first char of string? Pick one

src/app/interfaces/strongRoute.ts
```diff
@@ -113,22 +116,36 @@ export class StrongRoute {
...
+        // If both are parameter routes, they are equal
+        if (a.path.charAt(0) === ":" && b.path.startsWith(":")) {
+          return 0;
+        }
+        // Else give priority to the non-parameter route
+        return a.path.charAt(0) === ":" ? 1 : -1;
```
atruskie: you pull the first char out of each string so many times, might as well just

```
aParamRoute = a.path.charAt(0) === ":";
bParamRoute = b.path.charAt(0) === ":";
```
at the start of the method

src/app/interfaces/strongRoute.ts
```diff
@@ -113,22 +116,36 @@ export class StrongRoute {
...
-      // and not for both to match at same time.
-      thisRoute.children = [...(thisRoute.children || []), ...childRoutes];
+      thisRoute.children = [...(thisRoute.children || []), ...childRoutes].sort(
+        sortRoutes
+      );
       return thisRoute;
     };
 
-    return rootRoute.children.map(recursiveAdd);
+    const output = rootRoute.children.map(recursiveAdd).sort(sortRoutes);
```
atruskie, line 149: Why have two sorts? in line 140 and here?

src/app/interfaces/strongRoute.ts
```diff
@@ -142,7 +159,8 @@ export class StrongRoute {
    * Router representation of the route
    */
   toRoute(): string[] {
-    return this.full.map(x => x.name).filter(x => !!x);
+    const output = this.full.map(x => x.name);
+    return output.length > 1 ? output.filter(x => !!x) : [""];
```
atruskie: what does this condition guard against? why do we need to return an array with an empty string over an empty array? I'd say an empty array is more semantically correct.

atruskie (changes requested):

src/app/interfaces/strongRoute.ts
Allcharles: The sort function is not recursive. The inner sort handles the children routes, while the outer sort handles the base routes.

src/app/interfaces/strongRoute.ts
Allcharles: `[""]` is the base route. I remove it during the filter because otherwise you would have `['""', 'projects']` which would not be a valid route. As a counter-point,  i sometimes need `[""]` as a route (eg the HomeMenuItem), so i need `[""]` as an output. The only time its the output is if it is the only route in the list.

atruskie: Where does a case like this `['""', 'projects']` happen? As I read the code, it shouldn't.

The first component should always be `null` in  `StrongRoute.Base.add("projects")`

So why not just slice?

And for home? Rather than `null` the first component should be `""`. E.g.   `StrongRoute.Base.add("")`

Then there's always a null and code simply becomes: `this.slice(1).map(x => x.name)`

atruskie (approved):

PR #110, opened 2020-02-19
Send audio

Allcharles:
# Send Audio Page

Created the send audio page.

## Issues

- Could not properly implement the CMS logic for the page because the CMS cannot receive data from the component. Specifically this affects the Shipping details which should be hidden if the user is not logged in presumably to prevent bots from reading the postal address.
- Table in CMS is not mobile compatible

## Visual Changes

image

### DIFF
--- a/src/app/component/send-audio/send-audio.component.ts
+++ b/src/app/component/send-audio/send-audio.component.ts
@@ -1,0 +1,24 @@
+import { ChangeDetectionStrategy, Component, OnInit } from "@angular/core";
+import { List } from "immutable";
+import { Page } from "src/app/helpers/page/pageDecorator";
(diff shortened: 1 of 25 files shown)


src/app/component/send-audio/send-audio.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  menus: {
+    actions: List(),
+    links: List()
+  },
+  self: sendAudioMenuItem
+})
+@Component({
+  selector: "app-send-audio",
+  template: `
+    <app-cms page="sendAudio.html"></app-cms>
```
atruskie: pull the URL out from somewhere in code, please. The component instance is fine for now

atruskie (approved):

PR #111, opened 2020-02-19
Data request CMS

Allcharles:
# Updated Data Request Page

Extracted boilerplate descriptions from Data Request page into CMS files.

## Changes

- Extracted download annotations description to CMS
- Added download annotations form

## Visual Changes

image


src/app/component/data-request/data-request.component.ts
```diff
@@ -16,53 +17,62 @@ import { dataRequestCategory, dataRequestMenuItem } from "./data-request.menus";
...
   ngOnInit() {
-    this.loading = false;
+    this.requestLoading = false;
+  }
+
+  /**
+   * Form submission
+   * @param $event Form response
+   */
+  submitAnnotation($event: any) {
```
atruskie: why is this named annotation?

atruskie: Oh I see. Maybe "submitDownloadAnnotations"

atruskie (approved):
Why are the forms so skinny?

Allcharles:
All forms currently maintain the same size, we can create a toggle between large and small forms if you want them larger.
@atruskie

atruskie:
> All forms currently maintain the same size, we can create a toggle between large and small forms if you want them larger.
> @atruskie

I think they should fill to 100% of their container by default. We can use CSS to override container width where needed afterwards.

atruskie (approved):

PR #116, opened 2020-02-24
API Model Injection

### LINKED ISSUE #99: Inject api models into components

Allcharles:
# API Model Injection

This PR relates to retrieving models from the API before a component loads.

## Changes

- Created `FilterResolver`, `ListResolver`, and `ShowResolver` template classes
- Created resolvers for each api service
- Created error subroute to perform automatic error handling
  - This is similar to the secondary and action subroutes which are not directly affected by the main page
- Updated forms to newer notification system (`toastr`)
- Optimised api requests using new resolver system
  - Previously each component which needed access to a model had to either request it or be passed the model by its parent
  - New system allows requesting of models to flow through the children allowing all subroutes to access the models in use
- Updated menu system to include Category hierarchy
  - This allowed for the categories to define the resolvers for the child pages
  - Child pages can still define page specific resolvers

## Issues
 
- Pages still need to handle models failing to be retrieved
- Disabled the following unit test files (due to PR size):
  - component\about\pages\credits\credits.component.spec.ts
  - component\about\pages\disclaimers\disclaimers.component.spec.ts
  - component\about\pages\ethics\ethics.component.spec.ts
  - component\data-request\data-request.component.spec.ts
  - component\profile\pages\my-edit\my-edit.component.spec.ts
  - component\profile\pages\profile\my-profile.component.spec.ts
  - component\profile\pages\profile\their-profile.component.spec.ts
  - component\profile\pages\their-edit\their-edit.component.spec.ts
  - component\projects\pages\delete\delete.component.spec.ts
  - component\projects\pages\edit\edit.component.spec.ts
  - component\projects\pages\list\list.component.spec.ts
  - component\projects\pages\new\new.component.spec.ts
  - component\projects\pages\permissions\permissions.component.spec.ts
  - component\projects\pages\request\request.component.spec.ts
  - component\security\pages\login\login.component.spec.ts
  - component\send-audio\send-audio.component.spec.ts
  - component\shared\cms\cms.component.spec.ts
  - component\shared\error-handler\error-handler.component.spec.ts
  - component\shared\permissions-shield\permissions-shield.component.spec.ts
  - component\shared\user-badges\user-badge\user-badge.component.spec.ts
  - component\sites\pages\delete\delete.component.spec.ts
  - component\sites\pages\details\details.component.spec.ts
  - component\sites\pages\edit\edit.component.spec.ts
  - component\sites\pages\edit\edit.component.ts
  - component\sites\pages\harvest\harvest.component.spec.ts
  - component\sites\pages\new\new.component.spec.ts

## Closes

Closes #99


src/app/services/baw-api/account.service.ts
```diff
@@ -36,3 +37,25 @@ export class AccountService extends ReadonlyApi<User, []> {
...
+  providedIn: "root"
+})
+export class ShallowSiteResolver extends ShowResolver<User> {
+  constructor(api: AccountService) {
+    super(
+      api,
+      params => parseInt(params.get("accountId"), 10),
+      () => []
+    );
+  }
```
atruskie: Okay, that's a lot of ceremony still.

I'd like to see this reduced to a function. On usage something like `ListResolver<User>(AccountsService)`

atruskie (changes requested):
I'd like to see the resolver functions pretty much user defaults rather than defining a new XXXResolver class for every case. They have almost the same functionality - the core goal is to cut down on boilerplate.

src/app/services/baw-api/resolver-common.ts
```diff
@@ -1,53 +1,124 @@
...
+          catchError((error: ApiErrorDetails) => {
+            return of({ error });
+          })
+        );
+      }
+    }
+
+    return [
+      {
+        provide: name + "sResolver",
```
atruskie: I wonder if you could use `typeof` here to get typescript to put the name is for us?

Also, don't bother with the pluralization... it's not worth it in the end. Why not simpler: `XXXListResolver`

atruskie (approved):
Much cleaner implementation.

I'd suggest adding a few comments around the resolvers-common file to remind us of choices made and general workflow.

src/app/services/baw-api/resolver-common.ts
Allcharles: Unfortunately `typeof` does not work on custom classes, only `instanceof`

Allcharles:
@atruskie Can you re-review the logic handling resolver generation. It's changed since the initial design so that there isn't a doubled up list of resolvers accessible to a page.

atruskie (approved):

PR #118, opened 2020-02-24
Extract cms links

### LINKED ISSUE #113: Extract CMS links to deployment environment
CMS routes to access the pages from the BAW Server should not be hard coded into the application. Instead they should be extracted into the deployment environment to simplify the process of changing content between staging/development, and synchronising changes between server and client.

Allcharles:
# Environment CMS Paths

Extracted CMS links to environment to simplify the process of changing CMS file locations.

## Changes

- Updated `CMSComponent` to read urls from deployment environment
- Updated all components calling `<app-cms>` to the new path

## Issues

- Some duplication required in the mock app config. This should be fixed in another branch which removes the `AppConfigService`

## Closes

Closes #113


src/app/component/shared/cms/cms.component.ts
```diff
@@ -38,16 +37,24 @@ export class CmsComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit() {
+    this.loading = true;
+    const config = this.config.getConfig();
+    const cms = config.values.cms.find(cmsPage => cmsPage.title === this.page);
```
atruskie: this should be an O(1) not an O(n). Refactor to use keyed values

src/app/services/app-config/appConfigMock.service.ts
```diff
@@ -62,6 +62,44 @@ export class MockAppConfigService {
               }
             ]
           }
+        ],
+        cms: [
```
atruskie: turn into hash/map so something like `cms.credits.url` is possible

atruskie (changes requested):
So a couple of things:
 A previous commit removed `AppConfigService` - I'd suggest reinstating it. Sometimes you just need all of a config rather than a singular setting.

I'd like to see the property declarations (i.e. `public page = "xxx"`) turned into typed values or better yet they can lookup directly into the app config. The type checker should enforce retrieving a known constant. something like `public page = appConfig.pages.xxx`.

Allcharles:
~`AppConfigService` was removed because its unneeded. The deployment environment is stored inside the global environment so there was no point having a service which only returns the global environment variable.~

src/app/services/app-config/appConfigMock.service.ts
Allcharles: I'm not sure if creating a hash/map is possible as currently the deployment environment where these values come from is a `.json` file. We could however change it fairly easily into a `.ts` file.

atruskie: sure it is. Not a literal Hash type, but dictionary/hash/map are general terms for key-value based collection... aka an object in JS

in json:

```
"cms" : {
  "credits": {
    "url": "...
  }
}
```

atruskie (approved):

PR #121, opened 2020-02-25
Form guard

### LINKED ISSUE #104: User submission handling

Allcharles:
# Form Submission Guard

Added a guard to the website to protect users from leaving a half finished form.

## Changes

- Updated the following pages to enable guard:
  - Contact Us
  - Data Request
  - Edit My Account
  - Edit Their Account
  - Delete Project
  - Edit Project
  - New Project
  - Request Project Access
  - Report Problem
  - Confirm Password
  - Reset Password
  - Unlock Account
  - Delete Site
  - Edit Site
  - New Site
- Created `FormGuard`
- Created `WithFormCheck` mixin
- Fixed bug in forms where forms did not reset after user leaves page

## Issues

- Does not prevent users from closing the browser or leaving the site

## Closes

Closes #104 

## Visual Changes

User Prompt:
image


src/app/guards/form/form.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Constructor } from "src/app/helpers/advancedTypes";
+
+/**
+ * Add form checking to the component
+ * @param Base Class to extend
+ */
+export function WithFormCheck<T extends Constructor<{}>>(
+  Base: T = class {} as any
+) {
+  class Temporary extends Base {
```
atruskie: The name `Temporary` can show up in type annotations. Best to name it something better. Maybe `FormCheckingPageComponent1?

src/app/guards/form/form.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Form checking guard.
+ * This stops the user from leaving a page where a form has been
+ * modified by the user in any way.
+ */
+@Injectable()
+export class FormTouchedGuard
+  implements CanDeactivate<{ form: AbstractControl }> {
+  canDeactivate(component: any) {
+    // If component doesn't have a form, ignore it
+    if (!component.isFormTouched) {
```
atruskie: perhaps we want to do a type check here. `instanceof` and improve on the `component` type annotation - it should be T where T extends PageComponent or PageComponent & xxx` where xxx is an interface yet to be defined in this ile

src/app/helpers/advancedTypes.ts
```diff
@@ -1,9 +1,16 @@
...
+> &
+  {
+    [K in Keys]: T[K];
+  };
+
+/**
+ * Utilized by mixins. Refers to the constructor function of a
+ * class.
+ */
+export declare type Constructor<T> = new (...args: any[]) => T;
```
atruskie: angular or typescript already defines this. why not reuse their definition?

atruskie (changes requested):
Some very nice work. Very neat and integrates well with a minimum amount of boilerplate.

A few minor type annotation and naming tweaks required.

atruskie (approved):

PR #139, opened 2020-03-02
NGX Datatable Defaults

### LINKED ISSUE #100: Ngx-datatable custom directive
This is basically what I'm referring too: https://github.com/angular/angular/issues/14545
Looks like it is not a thing 🤷‍♂️

Don't over nest components - it just creates more work for browser and angular.

Maybe a custom directive would be better, like `ngx-standard-options` that does the binding that way.

_Originally posted by @atruskie in https://github.com/QutEcoacoustics/workbench-client/timeline_

Allcharles:
# Datatable Defaults

Created a directive for `<ngx-datatable>` components which loads the table with set defaults.

## Changes

- Created `src/app/directives/datatable/datatable.directive.ts
- To enable defaults the following must be added to a `<ngx-datatable>`

```typescript
<ngx-datatable
      #table
      appDatatable>
</ngx-datatable>
```

- To override defaults, you can do the following:

```typescript
<ngx-datatable
      #table
      appDatatable
      [defaults]=" { footerHeight: 150 } ">
</ngx-datatable>
```

- Updated Dependencies

## Closes

Closes #100


src/app/directives/datatable/datatable.directive.ts
```diff
@@ -0,0 +1,41 @@
...
+  AfterContentInit,
+  Directive,
+  ElementRef,
+  Input,
+  OnInit
+} from "@angular/core";
+import { DatatableComponent } from "@swimlane/ngx-datatable";
+
+@Directive({
+  selector: "[appDatatable]"
```
atruskie: suggested rename: `baw-datatable-defaults`

src/app/directives/datatable/datatable.directive.ts
```diff
@@ -0,0 +1,41 @@
...
+})
+export class DatatableDirective implements OnInit, AfterContentInit {
+  @Input() datatable: DatatableComponent;
```
atruskie: I believe you can use `@Host` here rather than `@Input` to get angular to inject whatever host component the directive is attached to

src/app/directives/datatable/datatable.directive.ts
```diff
@@ -0,0 +1,41 @@
...
+        scrollbarH: true,
+        reorderable: false
+      },
+      this.defaults
+    );
+  }
+
+  ngAfterContentInit() {
+    for (const key of Object.keys(this.datatableConfig)) {
+      this.datatable[key] = this.datatableConfig[key];
```
atruskie: This is problematic. In your PR description, you said: "To override defaults". Defaults don't need to be overridden (do they?).

I'd suggest this is implemented like this:

1. get host component
1. `for property in this.datatableConfig`
    1. if the host's element has attribute of the same name, continue
    1. else, set the property on host

atruskie (changes requested):
Looks pretty nice.

I'd suggest:
1. automatically getting host component (either via injection or `@Host`)
1. rename selector to `baw-datatable-defaults`
1. change the way overriding a default works

src/app/directives/datatable/datatable.directive.ts
Allcharles: Ngx-datatable comes with its own predefined defaults. Checking if a key exists before overwriting it would be hit and miss resulting in obscure issues.

src/app/directives/datatable/datatable.directive.ts
Allcharles: Angular convention for directives is to prepend app so that it doesn't conflict with a library. My counter offer would be `appBawDatatableDefaults`

atruskie: As per https://angular.io/guide/styleguide#directive-custom-prefix

Should be: `bawDatatableDefaults`

src/app/directives/datatable/datatable.directive.ts
atruskie: I think you misunderstand.

Order of priority for inputs are (highest wins):

1. ngx defaults
2. our defaults
3. special-case override of our defaults <-- I don't think we need this
3. a value set by an attribute in the template

The rules are (highest wins):

1. if no default set by us, use theirs
2. if our default set, use that
3. if our default are overridden,  use those (<-- still don't think we need this)
4. if an attribute is set use it

With your current method, you could well see that an attribute is ignored because you always assign all of our defaults to the host component.

Rather, to respect rule 4, you should only assign one of our defaults to the host component IFF an attribute of the same name is not already set in the template.

Allcharles: The only issue is that I have no method of determining what attributes are set, I can only see the values assigned to the component after it has been initialized. `<ngx-datatable>` default values, and attributes set by the parent component.

For example, the default headerHeight set by `ngx-datable` is 75. If I execute the following code in the parent component:

```html
<ngx-datatable
      #table
      bawDatatableDefaults
      [headerHeight]=150">
```

The `DatatableDirective` will only see headerHeight=150 when it checks the `ngx-datatable` component. Unless I maintain a list of default values set for the `DatatableComponent`, I have no method of knowing what changed. Furthermore, if the parent component sets a value to a `DatatableComponent` how can I tell that it has been changed?

Allcharles: The workaround I can think of is to set `@Input()`'s for all of the default config values the user can change. This would allow me to do what you want, however it would be fairly repetitive.

Eg:

```typescript
export class DatatableDirective implements OnInit, AfterContentInit {
  @Input() footerHeight: number;

  private datatableConfig: Partial<DatatableComponent>;

  constructor(
    @Host() private datatable: DatatableComponent,
    private datatableRef: ElementRef
  ) {}

  ngOnInit(): void {
    this.datatableRef.nativeElement.classList.add("bootstrap");
    this.datatableConfig.footerHeight = this.footerHeight || 50;
  }

  ngAfterContentInit() {
    for (const key of Object.keys(this.datatableConfig)) {
      this.datatable[key] = this.datatableConfig[key];
    }
  }
}

```

src/app/directives/datatable/datatable.directive.ts
Allcharles: I suspect that because angular ships with the following lint:
```
"directive-selector": [true, "attribute", "app", "camelCase"],
"component-selector": [true, "element", "app", "kebab-case"],
```
they want you to preface the directive and component with "app".
I've updated the rule to the following:

```
"directive-selector": [true, "attribute", ["app", "baw"], "camelCase"],
```

atruskie (changes requested):
This appears to have regressed from the last feedback.

I'd suggest deleting the commit and trying again

atruskie:
Okay, I've researched it further and I don't believe what I want is possible.

Leave as is, but check if that `ngAfterContentInit` hook is actually needed

atruskie (approved):

Allcharles:
Ill rebase and merge once I've fixed up all the missing unit tests from model injection

PR #140, opened 2020-03-02
Simplify CI

Allcharles:
# Simplify CI

Simplify and update CI to reduce load time.

## Changes

- Added linux, windows, and mac testing
- Linux OS builds project
- Removed analysis of build files
  - This can be done on a dev computer if its ever needed
- Updated node version requirements (10.x -> 12.x)
- Adding caching of node_modules to reduce load time


Allcharles:
Just realised upgrading the node version added a 6min wait to the CI, ill try figure out what is the newest supported version on the VM to remove the wait.

atruskie:
We should test on Windows and Linux, not just one.

atruskie (approved):

PR #152, opened 2020-03-12
Enhance category resolvers

Allcharles:
# Category Resolvers

Bandaide solution to applying resolvers to categories which do not have a root page (page with the same route as the category).

## Changes

- Enabled resolvers for pages where the category has no root page

## Issues

- Resolvers are not attached to the category route, instead they are applied to all category children when no root page is given
  - This means that some unnecessary api requests may occur (currently this only really affects the admin pages)
  - A more optimal solution would be to apply the resolvers to the category route instead
- This appears to be a flaw in the architecture, and may require a significant change to how routes are calculated

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -18,6 +18,7 @@ import { AppComponent } from "./app.component";
 import { appLibraryImports } from "./app.module";
 import { homeMenuItem } from "./component/home/home.menus";
 import { projectsMenuItem } from "./component/projects/projects.menus";
+import { SharedModule } from "./component/shared/shared.module";
 import { Project } from "./models/Project";
 import { AppConfigService } from "./services/app-config/app-config.service";
 import { ProjectsService } from "./services/baw-api/projects.service";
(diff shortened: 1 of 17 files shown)


atruskie:
Ok, I'm not sure I understand the problem.

Categories, as far as my mental model is aware, are just icon plus labels that can be placed to give a user navigational context.

Why would categories need resolvers at all?

Resolvers should be attached to pages.

Now, the category interface shape should overlap with a page interface, thus automatically making any page also a category, but that is just a useful by-product. I.e. there is no need for a "projects" category because the project page can be used as the category for all project subpages.

Allcharles:
Technically correct, however I need to add the resolver to the category instead of the page. Categories wrap a group of pages under a single group, usually we have a single page which acts like the category and all the pages stem from there. This allows me to add the categories to the one page which then filters down to all the children. There are, however, some scenarios where there is no single page for a category, but all of the pages require access to the same resolvers. I could create a dummy component to hold the resolvers and make all the other pages children of the component, however I would then have to create a redirect from that page to the PageNotFound component as otherwise it would be a valid route with no content.

image

 In the abnormal scenario, it would be more efficient to apply the resolvers to the category level instead as it would reduce additional traffic/wait time.

PR #153, opened 2020-03-12
Form templates

Allcharles:
# Form Templates

Reducing code duplication by creating form templates for the regular cases of New, Edit, and Delete forms.

## Changes

- Created `NewFormTemplate`, `EditFormTemplate`, `DeleteFormTemplate`

## Issues

- Haven't updated unit tests (awaiting feedback first)
- Could potentially create directive to create defaults for form submit button label and colour
  - Would be even more efficient if the directive could access the parent component and add the schema, and onSubmit inputs as well
  - Unsure if that is a possibility

### DIFF
--- a/src/app/component/projects/pages/delete/delete.component.ts
+++ b/src/app/component/projects/pages/delete/delete.component.ts
@@ -1,17 +1,14 @@
-import { Component, OnInit } from "@angular/core";
+import { Component } from "@angular/core";
 import { ActivatedRoute, Router } from "@angular/router";
 import { List } from "immutable";
 import { ToastrService } from "ngx-toastr";
 import { PermissionsShieldComponent } from "src/app/component/shared/permissions-shield/permissions-shield.component";
 import { WidgetMenuItem } from "src/app/component/shared/widget/widgetItem";
-import { WithFormCheck } from "src/app/guards/form/form.guard";
-import { PageComponent } from "src/app/helpers/page/pageComponent";
+import { DeleteFormTemplate } from "src/app/helpers/formTemplate/deleteTemplate";
 import { Page } from "src/app/helpers/page/pageDecorator";
 import { AnyMenuItem } from "src/app/interfaces/menusInterfaces";
 import { Project } from "src/app/models/Project";
-import { ApiErrorDetails } from "src/app/services/baw-api/api.interceptor.service";
 import { ProjectsService } from "src/app/services/baw-api/projects.service";
-import { ResolvedModel } from "src/app/services/baw-api/resolver-common";
 import {
   deleteProjectMenuItem,
   projectCategory,
@@ -33,9 +30,9 @@ import { projectMenuItemActions } from "../details/details.component";
   selector: "app-projects-delete",
   template: `
     <app-form
-      *ngIf="project"
+      *ngIf="models"
       [schema]="{ model: {}, fields: [] }"
-      [title]="'Are you certain you wish to delete ' + project.name + '?'"
+      [title]="'Are you certain you wish to delete ' + getProject().name + '?'"
       [btnColor]="'btn-danger'"
       [submitLabel]="'Delete'"
       [submitLoading]="loading"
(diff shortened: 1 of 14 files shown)


src/app/component/projects/pages/delete/delete.component.ts
```diff
@@ -33,61 +30,40 @@ import { projectMenuItemActions } from "../details/details.component";
   selector: "app-projects-delete",
   template: `
     <app-form
-      *ngIf="project"
+      *ngIf="models"
       [schema]="{ model: {}, fields: [] }"
-      [title]="'Are you certain you wish to delete ' + project.name + '?'"
+      [title]="'Are you certain you wish to delete ' + getProject().name + '?'"
```
atruskie, line 35: why `getProject()` here? It looks an awful lot like a property called with method syntax

src/app/component/projects/pages/delete/delete.component.ts
```diff
@@ -33,61 +30,40 @@ import { projectMenuItemActions } from "../details/details.component";
...
-    private router: Router,
-    private route: ActivatedRoute,
     private api: ProjectsService,
-    private notification: ToastrService
+    notifications: ToastrService,
+    route: ActivatedRoute,
+    router: Router
   ) {
-    super();
+    super(["project"], [], notifications, route, router);
```
atruskie, line 50: magic string?

src/app/component/projects/pages/delete/delete.component.ts
```diff
@@ -33,61 +30,40 @@ import { projectMenuItemActions } from "../details/details.component";
...
-      .project;
+  public getProject(): Project {
+    return this.models.project as Project;
+  }
 
-    if (projectModel.error) {
-      return;
-    }
+  apiDestroy() {
+    const project = this.getProject();
```
atruskie, line 58: this seems overly specific

src/app/component/projects/pages/edit/edit.component.ts
```diff
@@ -32,75 +28,55 @@ import { fields } from "./edit.json";
...
-        () => {
-          this.resetForms();
-          this.notification.success("Project was successfully updated.");
-          this.router.navigateByUrl(project.redirectPath());
-        },
-        (err: ApiErrorDetails) => {
-          let errMsg: string;
+  apiUpdate(event: FormEvent) {
+    const project = this.getProject();
+    const updatedProject = new Project({ id: project.id, ...event });
```
atruskie, line 65: why are we recreating the model here?

Can't we just push to model that was edited by the form? i.e. `this.models.project`

src/app/component/projects/pages/edit/edit.component.ts
```diff
@@ -32,75 +28,55 @@ import { fields } from "./edit.json";
...
-
-    if (projectModel.error) {
-      return;
-    }
-
-    this.project = projectModel.model;
-    this.schema.model["name"] = this.project.name;
-    this.schema.model["description"] = this.project.description;
+  public getProject(): Project {
+    return this.models.project as Project;
```
atruskie, line 53: why are we nesting models here?

src/app/helpers/formTemplate/deleteTemplate.ts
```diff
@@ -0,0 +1,18 @@
...
+import { FormTemplate } from "./formTemplate";
+
+export abstract class DeleteFormTemplate<
+  M extends AbstractModel
+> extends FormTemplate<M, {}> {
+  /**
+   * Create delete request
+   * @param event Form submission
+   */
+  protected abstract apiDestroy(): Observable<M | void>;
```
atruskie, line 12: Why do we need this?

The API has a standard destroy method, as in the interface.

Is this a before/after hook?

src/app/helpers/formTemplate/formTemplate.ts
```diff
@@ -0,0 +1,118 @@
...
+  ngOnInit() {
+    // Grab models from route data
+    const data = this.route.snapshot.data;
+    const resolvedModels: ResolvedModel<AbstractModel>[] = this.modelKeys.map(
+      key => data[key]
+    );
+
+    // If any models failed, return
+    if (resolvedModels.some(resolvedModel => resolvedModel.error)) {
+      return;
```
atruskie, line 45: what is the effect of returning early here? Is this component broken? Disabled?

src/app/helpers/formTemplate/formTemplate.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // Add models to object
+    this.models = {};
```
atruskie, line 49: why not initiate this at field declaration?

src/app/helpers/formTemplate/formTemplate.ts
```diff
@@ -0,0 +1,118 @@
...
+  public submit(event: F) {
+    this.loading = true;
+
+    this.apiAction(event)
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe(this.apiSuccess, this.apiError);
+  }
+}
+
+export interface FormlySchema {
```
atruskie, line 115: I don't understand why the schema contains a copy of the model?

src/app/helpers/formTemplate/formTemplate.ts
```diff
@@ -0,0 +1,118 @@
...
+    protected route: ActivatedRoute,
+    protected router: Router
+  ) {
+    super();
+  }
+
+  ngOnInit() {
+    // Grab models from route data
+    const data = this.route.snapshot.data;
+    const resolvedModels: ResolvedModel<AbstractModel>[] = this.modelKeys.map(
```
atruskie, line 39: does data ever contain non-models?

If it doesn't, then there's no need to specify keys?

atruskie (changes requested):
I appreciate what you've done with formTemplate.ts, but the downrange effects are overly-complex. 

I think there's a far simpler core to this idea, a collection of boilerplate, that can and should be extracted. However, I don't think this form is appropriate. 

I think we need to do a call to explain this all to each other.

src/app/component/projects/pages/edit/edit.component.ts
Allcharles: Models are readonly to preserve immutability

src/app/helpers/formTemplate/formTemplate.ts
Allcharles: Yes, this returns before any of the variables are initialised so that the parent component knows not to render anything. I can make it set a `error` variable or something instead to make this behaviour more obvious.

src/app/helpers/formTemplate/formTemplate.ts
Allcharles: Data does contain non-models (specifically it contains all of the page data as well), but ill move to your suggestion because the page data shouldn't affect the behaviour

atruskie:
Superseded by #161

PR #155, opened 2020-03-16
Owner Predicate

### LINKED ISSUE #144: Menu Predicates
Update menu predicates to handle the following scenarios:

- [x] isAdmin
  - Call `my_account` on login and save to session to determine this value
- [x] isOwner
  - This will need to somehow leave room for a route to provide a custom checking value for the user such as a callback

Allcharles:
# Link Project Owner Predicate

Updated link predicate for detecting project owners so that it now checks the `user.id` against the `project.ownerId`.

## Issues

- Currently the BAW Server does not provide the ownerId in its output for a project so this predicate only functions for the admin user group
- Unit tests failing on edge browser (will fix in separate PR)
  - AdminUserListComponent > api > should request final page from api
  - AdminUserListComponent > api > should request the second page from api
  - AdminUserListComponent > api > should request next page from api

## Closes

Closes #144

### DIFF
--- a/src/app/app.menus.spec.ts
+++ b/src/app/app.menus.spec.ts
@@ -1,0 +1,121 @@
+import {
+  isAdminPredicate,
+  isGuestPredicate,
+  isLoggedInPredicate,
+  isProjectOwnerPredicate
+} from "./app.menus";
+import { Project } from "./models/Project";
+import { SessionUser } from "./models/User";
+import { ApiErrorDetails } from "./services/baw-api/api.interceptor.service";
+import { ResolvedModel } from "./services/baw-api/resolver-common";
+
+describe("Predicates", () => {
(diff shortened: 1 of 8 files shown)


atruskie (approved):
I'm going to approve this because the change is fine, however, do note that we'll use capabilities to determine if someone can do something.

We won't be using the predicate system to hide or show links in for authorisation cases. There is some overlap in the functionality, and that is because capabilities supersede the predicate system. 

It'll be something more like this:

The API will look like:
`project.capabilities.edit` 

We might have a directive like: `baw-auth-hide="edit"` or `baw-auth-disable="edit"` that we apply to relevant elements, that knows how to access capabilities and conventionally can access the relevant model (or we can supply the model by reusing the model directive).

PR #160, opened 2020-03-23
Route resolvers

### LINKED ISSUE #158: Route resolvers
Redo implementation of route resolvers. Currently, resolvers are treated as a hereditary system where the parents resolvers are also applied to the children. This needs to change so that all children define their own resolvers individually, and resolvers are not passed to any children.

### LINKED ISSUE #159: StrongRoute child category
Child categories (categories which are not a root `StrongRoute`) create duplicate routes. Currently this does not causes issues, however if there becomes a difference between the children route and parent route versions, this could cause intermittent bugs dependent on the load order of modules. The cause of this issue is that the root of the `StrongRoute` belongs to another category. Because of this, the routes for the parent categories are recalculated and added to the applications route. The solution to this is to allow a `StrongRoute` to reference another `StrongRoute` to inherit from, but not calculate from.

Allcharles:
# Route Resolvers 2.0

Re-implemented resolvers to attach them to a page instead of the category. This removed inheritance of resolvers from parent pages as well.

## Changes

- Fixed child categories redefining parent routes
- Flattened routes instead of previous nested route solution
- Removed category resolvers
- Updated app dependencies

## Closes

Closes #159 
Closes #158


src/app/interfaces/strongRoute.ts
```diff
@@ -66,9 +76,10 @@ export class StrongRoute {
    * Add a child route
    * @param name Route name
    * @param config Additional router configurations
+   * @param isRoot Is this a root category
    */
-  add(name: string, config: Partial<Route> = {}) {
-    return new StrongRoute(this, name, config);
+  add(name: string, config: Partial<Route> = {}, isRoot?: boolean) {
```
atruskie: why are we exposing `isRoot` publically?

atruskie (approved):
It's a little worrying that all the resolvers are referred to by strings.

src/app/interfaces/strongRoute.ts
Allcharles: Maybe the variable name is a little misleading, it is designed for child categories such as the Sites Category which allows it to set itself as the root, even though the Projects Category is the actual root. Basically it lets me stop the child category redefining the parent category routes and vice versa. This is part of the solution for #159

Allcharles:
Ill look into referring resolvers by variable instead of string, it should be simple enough to do. It'll just fatten up the resolver declaration a little.

atruskie:
If we could use the angular injector here that would be cool (so we'd be using type signatures to do the injection)

src/app/interfaces/strongRoute.ts
atruskie: Responded to #159

Allcharles:
Re-review requested to look at the new resolver key system (extracted to constant variable), and resolver name system (`projectResolvers.show`)

src/app/interfaces/strongRoute.ts
```diff
@@ -76,10 +76,18 @@ export class StrongRoute {
...
+  add(name: string, config: Partial<Route> = {}) {
+    return new StrongRoute(this, name, config);
+  }
+
+  /**
+   * Add a child route
+   * @param name Route name
+   * @param config Additional router configurations
+   */
+  addFeatureModule(name: string, config: Partial<Route> = {}) {
```
atruskie, line 89: As per https://github.com/QutEcoacoustics/workbench-client/issues/159#issuecomment-603528912 I'd prefer if mature models used a different flag and did not reuse the `isRoot` property.

This can be fixed in a later commit, however.

atruskie (approved):

PR #161, opened 2020-03-23
Formly redesign

### LINKED ISSUE #157: Re-implement Formly
Issues with current implementation:

- Created schema interface. This should instead be split back into model and fields
- Form Wrapper should accept `AbstractModel` which it provides to the formly model input
  - This should allow the user to directly edit the model instead of only during submission
- Cleanup unit tests as part of #146 
  - Specifically, dont test the forms. Test the config file and enhance the testing of the formly wrapper
- Rewrite schema grab in form wrapper, it should be a service
- Remove new/edit/delete templates, just extend the base class
- Remove all of the abstract functions
  - Allow overriding in child class through super call
- Potentially completely remove `.json` files for forms, just read from the AbstractModel

Allcharles:
# Formly Redesign (Do Not Merge)

Re-implementation/redesign of formly wrapper component and form template. The intention is for this PR to change the formly implementation to a sustainable solution. Other PRs will come later (and merge into this) to update the rest of the pages and unit tests.

## Changes

- Created `formTemplate`
- Changed `<app-form>` inputs to accept `model` and `fields`

## Issues

- Not completed, will create additional PR's to merge into this before accepting the final changes

## Closes 

Closes #157


src/app/component/projects/pages/edit/edit.component.ts
```diff
@@ -35,75 +32,58 @@ import { fields } from "./edit.json";
...
+        let errMsg: string;
+
+        // Handle additional error details
+        if (err.info && err.info.name && err.info.name.length === 1) {
+          errMsg = err.message + ": name " + err.info.name[0];
+        } else {
+          errMsg = err.message;
+        }
+        return errMsg;
+      }
```
atruskie: these both look fairly standard - refactor opportunity?

src/app/component/projects/pages/edit/edit.component.ts
```diff
@@ -35,75 +32,58 @@ import { fields } from "./edit.json";
...
+    notifications: ToastrService,
+    route: ActivatedRoute,
+    router: Router
   ) {
-    super();
+    super(
+      notifications,
+      route,
+      router,
+      ["project"],
```
atruskie: string token here is concerning

src/app/component/shared/form/form.component.ts
```diff
@@ -26,69 +23,38 @@ import { ApiResponse } from "src/app/services/baw-api/baw-api.service";
...
-  @Input() btnColor:
-    | "btn-danger"
-    | "btn-success"
-    | "btn-warning"
-    | "btn-primary"
-    | "btn-secondary"
-    | "btn-info" = "btn-success";
   /**
-   * Deprecated
+   * DEPRECATED
```
atruskie: the diff is a little hard to read - is this entire component deprecated?

src/app/helpers/formTemplate/formTemplate.ts
```diff
@@ -0,0 +1,118 @@
...
+   * @param router Router service
+   * @param modelKeys Router data model keys (first key is default model)
+   * @param successMsg Success message
+   * @param errorMsg Error message
+   */
+  constructor(
+    protected notifications: ToastrService,
+    protected route: ActivatedRoute,
+    protected router: Router,
+    private modelKeys: string[],
```
atruskie: so this relates to my previous string focused comment. 

I wonder, can we attach to models (maybe to AbsrtactModel) a method that returns the name of the model?
Then all we'd need to pass here is any model or model constructor ...

src/app/helpers/formTemplate/formTemplate.ts
```diff
@@ -0,0 +1,118 @@
...
+    private successMsg: (model: Partial<M>) => string,
+    private errorMsg: (err: ApiErrorDetails) => string = err => err.message
```
atruskie: these could probably have defaults set

atruskie (changes requested):
This looks much cleaner than #153 - well done.

A little bit more work to clean-up those stringly-typed data keys and we'll be set

src/app/component/shared/form/form.component.ts
Allcharles: No, just `@Input() schema: any`. I haven't removed it so that I can still compile the app while i make changes to the new inputs: `model`, and `fields`

src/app/helpers/formTemplate/formTemplate.ts
Allcharles: So i already thought about that, and the `kind` key will tell us what type of `AbstractModel` we are dealing with. However i couldnt figure out the next step of typing the variable.

atruskie (approved):

atruskie (approved):

PR #182, opened 2020-04-09
Harvest Pages

Allcharles:
# Harvest Pages

Prototype harvest pages. This PR is just so I don't delete the branch until it's no longer needed.


src/app/component/projects/pages/harvest/harvest.component.scss
```diff
@@ -0,0 +1,4 @@
+// Set width of prime ng steps so that they fill screen
+.ui-steps .ui-steps-item {
+  width: 16.66%; // 100 / 6
```
atruskie: this seems to be hacky.

Can we get the items inside to scale to their box? Flexbocx should do this

src/app/component/projects/pages/harvest/harvest.component.ts
```diff
@@ -0,0 +1,177 @@
...
+        this.progress = 100;
+        clearInterval(this.interval);
+
+        if (callback) {
+          callback();
+        }
+      }
+    }, this.intervalSpeed);
+  }
+
```
atruskie: Could probably use rxjs http://reactivex.io/documentation/operators/interval.html to emit event periodically.

While mocking, we'd generate fake progress.
When stubbing to a real service we'd then do an api call, but the same observable mechanism can be used

src/app/component/projects/pages/harvest/harvest.component.ts
```diff
@@ -0,0 +1,177 @@
...
+      case Harvest.Summary: {
+        this.buttons = {
+          previous: {},
+          next: {},
+        };
+        break;
+      }
+    }
+  }
+}
```
atruskie: As a style preference, I'd prefer to see the stage data in a single object/lookup, rather than as a series of object literals.

e.g.

```
const stages = {
    check: {
        buttons: { previous: { text: "back" }, next: { text: "next" } }
    },
   review: { ... }
}
```

src/app/component/shared/indicator/indicator.component.ts
```diff
@@ -0,0 +1,42 @@
...
+      this.icon = ["fas", "check"];
+      this.width = "14px";
+    } else if (this.status === Status.Error) {
+      this.color = "red";
+      this.icon = ["fas", "times"];
+      this.width = "10px";
+    } else {
+      this.color = "orange";
+      this.icon = ["fas", "exclamation-triangle"];
+      this.width = "19px";
```
atruskie: style and icons names should not leak into component code.

switch on `this.status` in the template and use scss file if needed

src/app/services/baw-api/sites.service.ts
```diff
@@ -38,7 +38,11 @@ export class SitesService extends StandardApi<Site, [IdOr<Project>]> {
     return this.apiList(endpoint(project, Empty, Empty));
   }
   filter(filters: Filters, project: IdOr<Project>): Observable<Site[]> {
-    return this.apiFilter(endpoint(project, Empty, Filter), filters);
+    return filterMock(
+      filters,
+      (index) => new Site({ id: index, name: "PLACEHOLDER" })
+    );
+    // return this.apiFilter(endpoint(project, Empty, Filter), filters);
```
atruskie, line 45: Why are we committing this change?

atruskie (changes requested):
It's be good to understand how our bundle size was affected by introducing primeNG.

As per https://angular.io/guide/styleguide#component-custom-prefix all these `app-` prefixes on various selectors should be instead `baw-`. `app` has no meaning and the prefix might as well be removed. `baw` namespaces the components so they clearly ours.

src/app/services/baw-api/sites.service.ts
Allcharles: Sites filter is broken I'm fairly certain.

atruskie: Add in a code comment then and link to relevant baw-server issue

Allcharles:
# Before

Bundle sizes:
image

Bundle size analysis:
image

# After

Bundle sizes:
image

Bundle size analysis:
image

src/app/services/baw-api/sites.service.ts
Allcharles: Baw issue: https://github.com/QutEcoacoustics/baw-server/issues/437

src/app/component/projects/pages/harvest/harvest.component.ts
Allcharles: I have changed the system to utilize an observable now, however i have not extracted it to a harvest service for fear of a merge conflict with the changes we have made to the api services.

Allcharles:
Linking relevant issue: https://github.com/QutEcoacoustics/workbench-client/issues/221

src/app/component/projects/pill-list/pill-list.component.html
```diff
@@ -0,0 +1,19 @@
...
+  <fa-icon [icon]="['fas', 'expand-alt']"></fa-icon>
+</button>
+
+<button
+  *ngIf="!ellipsis && compress"
+  class="btn btn-primary badge badge-primary m-1"
+  (click)="toggleExpansion()"
+>
+  <fa-icon [icon]="['fas', 'compress-alt']"></fa-icon>
+</button>
```
atruskie: The only difference between these two cases is the boolean expression and the class. Perhaps these can be collapsed into a singular button?

atruskie (approved):

PR #179, opened 2020-04-09
Alias imports

Allcharles:
# Alias imports

Created custom path mapping in typescript to reduce import complexity in files.

## Changes

- Created the following paths:
  - assets
  - shared
  - component
  - directives
  - guards
  - helpers
  - interfaces
  - models
  - baw-api
  - services

Changes:

```typescript
import { Component, OnInit } from "@angular/core";
import { ActivatedRoute } from "@angular/router";
import { List } from "immutable";
import { Card } from "src/app/component/shared/cards/cards.component";
import { PageComponent } from "src/app/helpers/page/pageComponent";
import { Page } from "src/app/helpers/page/pageDecorator";
import { AnyMenuItem } from "src/app/interfaces/menusInterfaces";
import { Project } from "src/app/models/Project";
import { projectResolvers } from "src/app/services/baw-api/projects.service";
import { ResolvedModel } from "src/app/services/baw-api/resolver-common";
import {
  newProjectMenuItem,
  projectsCategory,
  projectsMenuItem,
  requestProjectMenuItem
} from "../../projects.menus";
```

Into:

```typescript
import { Component, OnInit } from "@angular/core";
import { ActivatedRoute } from "@angular/router";
import { projectResolvers } from "@baw-api/projects.service";
import { ResolvedModel } from "@baw-api/resolver-common";
import {
  newProjectMenuItem,
  projectsCategory,
  projectsMenuItem,
  requestProjectMenuItem,
} from "@component/projects/projects.menus";
import { PageComponent } from "@helpers/page/pageComponent";
import { Page } from "@helpers/page/pageDecorator";
import { AnyMenuItem } from "@interfaces/menusInterfaces";
import { Project } from "@models/Project";
import { Card } from "@shared/cards/cards.component";
import { List } from "immutable";
```

## Issues

- Haven't changed all files, waiting for review first

## Closes

Closes #162

PR #184, opened 2020-04-13
Admin modules

Allcharles:
# Extracted admin modules

Cleaned up admin pages by extracting sub components into their own modules.

## Changes

- Created `ScriptsModule`
- Created `TagGroupsModule`
- Created `TagsModule`
- Fixed StrongRoute bug where `compileRoutes()` was not compiling the root route. This was an issue for modules using `addFeatureModule()`

PR #185, opened 2020-04-13
Revert "Alias imports"

Allcharles:
Reverts QutEcoacoustics/workbench-client#179


Allcharles:
Closed this PR because it was the initial solution to the #184 rebase error. However i used the `git cherrypicker` method instead in #186 .

PR #189, opened 2020-04-14
User sites pages

### LINKED ISSUE #136: My Sites
Create both my and their sites pages (`/my_account/sites`, `/user_accounts/:id/sites`)

Allcharles:
# User Sites Pages

Created user sites pages. (`their-sites.component.ts` non-functional)

## Issues
Unable to fully implement because of the following issues:

- No route for accessing a users sites
- ShallowSites routes don't exist
- Annotations page does not properly exist
- Sites output does not contain permissions and recent audio uploads

## Closes

Addresses #136 

## Visual Changes

My Sites Page

image

Their Sites Page

image

### DIFF
--- a/src/app/component/profile/pages/sites/their-sites.component.ts
+++ b/src/app/component/profile/pages/sites/their-sites.component.ts
@@ -1,11 +1,11 @@
 import { Component } from "@angular/core";
 import { ActivatedRoute } from "@angular/router";
+import { accountResolvers } from "@baw-api/account.service";
 import { ShallowSitesService } from "@baw-api/sites.service";
-import { userResolvers } from "@baw-api/user.service";
 import {
-  myAccountCategory,
-  myAccountMenuItem,
-  mySitesMenuItem,
+  theirProfileCategory,
+  theirProfileMenuItem,
+  theirSitesMenuItem,
 } from "@component/profile/profile.menus";
 import { annotationsMenuItem } from "@component/sites/sites.menus";
 import { Page } from "@helpers/page/pageDecorator";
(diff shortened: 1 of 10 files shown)


atruskie:
Add tracking issues for the problems you mentioned, cross link them here, and link to baw-server issues in any relevant place

src/app/component/profile/pages/sites/their-sites.component.ts
```diff
@@ -0,0 +1,72 @@
...
+      (sites) =>
+        sites.map((site) => ({
+          site: {
+            label: site.name,
+            route: site.redirectPath(),
+          },
+          recentAudioUpload: "(none)",
+          permission: "FIX ME",
+          annotation: annotationsMenuItem.uri(undefined),
+        })),
```
atruskie, line 54: same as projects list page - why not map this into a `Site` model?

src/app/models/Site.ts
```diff
@@ -69,18 +69,17 @@ export class Site extends AbstractModel implements SiteInterface {
...
+      description: this.description,
     };
   }
 
-  redirectPath(project: Project): string {
+  redirectPath(project?: Project): string {
     return siteMenuItem.route.format({
-      projectId: project.id,
-      siteId: this.id
+      projectId: project?.id || this.projectIds[0] || -1,
```
atruskie: this seems dangerous - there's a possibility of a `-1` appearing in a URL? better to just return null or empty string if a valid URL is not possible

atruskie (approved):

src/app/models/Site.ts
Allcharles: I think ill just put a check, if no project.id or this.projectIds[0], return "/BROKEN_LINK".

atruskie: No I don't think so. Then you run the risk of "/BROKEN_LINK" actually getting inserted into an href.

I'd recommend:
- throwing if you want this to be a fatal error
- or returning something falsey if you want to fail gracefully.

But returning a string with non-zero length to indicate an error in a function that returns strings is not a great idea.

atruskie (approved):

Allcharles:
Baw issues:
- https://github.com/QutEcoacoustics/baw-server/issues/438
- https://github.com/QutEcoacoustics/baw-server/issues/406

PR #190, opened 2020-04-15
Models

### LINKED ISSUE #173: Abstract model cleanup
- [x] Renamed `redirectPath` to something more descriptive
- [x] Cleanup `toJSON` implementations to remove non-editable fields
- [x] Change `notes` datatype from `string` to `object/hash`
  - `notes` inputs are json blobs
  - Potentially need to use this: [class](https://github.com/eligrey/Blob.js) (recommended by [angular](https://angular.io/guide/browser-support#suggested-polyfills))

Allcharles:
# BAW Server Models

Created the basic framework for the BAW server models.

## Changes

Created the following models:

- `AnalysisJob`
- `AnalysisJobItem`
- `AudioEvent`
- `AudioEventComment`
- `AudioEventTag`
- `Bookmark`
- `Dataset`
- `DatasetItem`
- `ProgressEvent`
- `SavedSearches` <- Will rename to SavedSearch in another PR
- `Study`

Renamed `redirectionPath()` to `viewUrl` and `getViewUrl()`

## Issues

Changing models to try use Blobs likely caused some visual bugs on model forms. However this only affects pages which are connected to mock services so it has no affect on functionality.

## Closes 

Closes #173


Allcharles:
~~Need to change `navigationPath()` to `get viewUrl()`.~~ Not possible as navigationPath sometimes requires inputs.

src/app/models/AnalysisJob.ts
```diff
@@ -0,0 +1,98 @@
...
+  public get overallDuration() {
+    return new Duration();
+    // return duration(this.overallDurationSeconds) // TODO Awaiting PR #177
+  }
+
+  constructor(analysisJob: IAnalysisJob) {
+    super(analysisJob);
+
+    this.customSettings = new Blob([analysisJob.customSettings]);
+    this.overallProgress = new Blob([analysisJob.overallProgress]);
```
atruskie, line 99: this should be JSON in DB, and part of object graph in response

src/app/models/AnalysisJob.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  public get overallDuration() {
+    return new Duration();
+    // return duration(this.overallDurationSeconds) // TODO Awaiting PR #177
+  }
+
+  constructor(analysisJob: IAnalysisJob) {
+    super(analysisJob);
+
+    this.customSettings = new Blob([analysisJob.customSettings]);
```
atruskie, line 98: This will usually be something like a YAML file - Blob is correct type

src/app/models/AnalysisJob.ts
```diff
@@ -0,0 +1,98 @@
...
+    );
+  }
+
+  public toJSON() {
+    return {
+      id: this.id,
+      name: this.name,
+      annotationName: this.annotationName,
+      description: this.description,
+    };
```
atruskie: For `create` customSettings is writeable and should be serialized

src/app/models/AnalysisJob.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 103: i'd rather see this error so we catch it early, especially since intend to add this view

src/app/models/AnalysisJob.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+type Status = "??? Anthony";
```
atruskie, line 108: Relevant reading: https://github.com/QutEcoacoustics/baw-server/blob/d165c88677ac97f1fa6be47d28b7774b1bed7846/app/models/analysis_job.rb#L118-L149

src/app/models/AnalysisJobItem.ts
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }
+
+  public toJSON() {
+    return {
+      id: this.id,
+      analysisJobId: this.analysisJobId,
+      audioRecordingId: this.audioRecordingId,
+      queueId: this.queueId,
+      status: this.status,
```
atruskie: this model should always be read-only

src/app/models/AnalysisJobItem.ts
```diff
@@ -0,0 +1,63 @@
...
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 47: should throw

src/app/models/AnalysisJobItem.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+export type Status = "new" | "??? Anthony";
```
atruskie, line 52: https://github.com/QutEcoacoustics/baw-server/blob/d165c88677ac97f1fa6be47d28b7774b1bed7846/app/models/analysis_jobs_item.rb#L190-L217

src/app/models/AudioEvent.ts
```diff
@@ -0,0 +1,63 @@
...
+      startTimeSeconds: this.startTimeSeconds,
+      endTimeSeconds: this.endTimeSeconds,
+      lowFrequencyHertz: this.lowFrequencyHertz,
+      highFrequencyHertz: this.highFrequencyHertz,
+      isReference: this.isReference,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 68: throw

src/app/models/AudioEventComment.ts
```diff
@@ -0,0 +1,65 @@
...
+      flagExplain: this.flagExplain,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
+  }
+}
+
+type Flag = "??? Anthony";
```
atruskie: Remove entire model, deprecated

src/app/models/AudioEventTag.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      id: this.id,
+      audioEventId: this.audioEventId,
+      tagId: this.tagId,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
+  }
```
atruskie, line 53: leave a note: this will probably never have it's own view. Should return `null`?

src/app/models/AudioRecording.ts
```diff
@@ -41,36 +51,52 @@ export class AudioRecording extends AbstractModel
...
+      uuid: this.uuid,
+      siteId: this.siteId,
+      durationSeconds: this.durationSeconds,
+      sampleRateHertz: this.sampleRateHertz,
+      channels: this.channels,
+      bitRateBps: this.bitRateBps,
+      mediaType: this.mediaType,
+      dataLengthBytes: this.dataLengthBytes,
+      status: this.status,
+      notes: this.notes,
```
atruskie: AudioRecordings will probably always be immutable for the client app.

src/app/models/AudioRecording.ts
```diff
@@ -41,36 +51,52 @@ export class AudioRecording extends AbstractModel
...
     };
   }
 
-  redirectPath(): string {
-    throw new Error("Not Implemented");
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 100: throw.

src/app/models/AudioRecording.ts
```diff
@@ -41,36 +51,52 @@ export class AudioRecording extends AbstractModel
...
   }
 }
+
+type Status = "ready" | "uploading" | "corrupt";
```
atruskie, line 104: https://github.com/QutEcoacoustics/baw-server/blob/d165c88677ac97f1fa6be47d28b7774b1bed7846/app/models/audio_recording.rb#L39

src/app/models/Bookmark.ts
```diff
@@ -0,0 +1,62 @@
...
+      id: this.id,
+      offsetSeconds: this.offsetSeconds,
+      name: this.name,
+      description: this.description,
+      category: this.category,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 66: we'll need a `listenViewUrl` function at some point. It'll accept recording/recordingId, and optional start offset

src/app/models/Dataset.ts
```diff
@@ -0,0 +1,48 @@
...
+  public toJSON() {
+    return {
+      id: this.id,
+      name: this.name,
+      description: this.description,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 53: `null` for now - no view planned

src/app/models/DatasetItem.ts
```diff
@@ -0,0 +1,50 @@
...
+      datasetId: this.datasetId,
+      audioRecordingID: this.audioRecordingID,
+      startTimeSeconds: this.startTimeSeconds,
+      endTimeSeconds: this.endTimeSeconds,
+      order: this.order,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 51: null for now - no view planned

src/app/models/ProgressEvent.ts
```diff
@@ -0,0 +1,41 @@
...
+  public toJSON() {
+    return {
+      id: this.id,
+      datasetItemId: this.datasetItemId,
+      activity: this.activity,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 42: null for now - no view planned

src/app/models/SavedSearches.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      id: this.id,
+      name: this.name,
+      description: this.description,
+      storedQuery: this.storedQuery,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie: throw

src/app/models/SavedSearches.ts
```diff
@@ -0,0 +1,51 @@
...
+  createdAt?: DateTimeTimezone | string;
+  deletedAt?: DateTimeTimezone | string;
+}
+
+export class SavedSearches extends AbstractModel implements ISavedSearches {
+  public readonly kind: "SavedSearches" = "SavedSearches";
+  public readonly id?: Id;
+  public readonly name?: Param;
+  public readonly description?: Description;
+  public readonly storedQuery?: Blob;
```
atruskie: This will be an object of type `Filter`

src/app/models/Study.ts
```diff
@@ -0,0 +1,47 @@
...
+  public toJSON() {
+    return {
+      id: this.id,
+      name: this.name,
+      datasetId: this.datasetId,
+    };
+  }
+
+  public get viewUrl(): string {
+    return "/BROKEN_LINK";
```
atruskie, line 49: throw

src/app/models/Tag.ts
```diff
@@ -26,36 +26,28 @@ export interface TagInterface {
...
  * Tag model
  */
-export class Tag extends AbstractModel implements TagInterface {
+export class Tag extends AbstractModel implements ITag {
   public readonly kind: "AbstractModel" = "AbstractModel";
   public readonly id?: Id;
   public readonly text?: string;
   public readonly count?: number;
   public readonly isTaxanomic?: boolean;
   public readonly typeOfTag?: string;
```
atruskie, line 46: https://github.com/QutEcoacoustics/baw-server/blob/d165c88677ac97f1fa6be47d28b7774b1bed7846/app/models/tag.rb#L20-L26

src/app/models/Tag.ts
```diff
@@ -26,36 +26,28 @@ export interface TagInterface {
 /**
  * Tag model
  */
-export class Tag extends AbstractModel implements TagInterface {
+export class Tag extends AbstractModel implements ITag {
   public readonly kind: "AbstractModel" = "AbstractModel";
   public readonly id?: Id;
   public readonly text?: string;
   public readonly count?: number;
```
atruskie, line 42: this field should not be returned by the API?

src/app/models/User.ts
```diff
@@ -10,48 +10,85 @@ import {
...
+    );
+    this.lockedAt = dateTimeTimezone(user.lockedAt as string);
+    this.createdAt = dateTimeTimezone(user.createdAt as string);
+    this.updatedAt = dateTimeTimezone(user.updatedAt as string);
     this.lastSeenAt = dateTimeTimezone(user.lastSeenAt as string);
 
     this.imageUrls = user.imageUrls
-      ? user.imageUrls.map(imageUrl => {
+      ? user.imageUrls.map((imageUrl) => {
           // TODO Add /assets by default from the API so this check doesn't need to occur
```
atruskie, line 97: filed an issue?

src/app/models/User.ts
```diff
@@ -175,23 +204,15 @@ export class SessionUser extends User implements SessionUserInterface {
...
-      obj = JSON.parse(obj);
-    }
-
-    return new SessionUser(obj);
-  };
-
-  toJSON() {
+  public toJSON() {
     return {
       authToken: this.authToken,
```
atruskie, line 197: not writeable

atruskie (changes requested):
It's great that all the intermediate objects used for view rendering were removed. Removes a lot of fluff and reduces chance for weird-state partial models causing unexpected behaviour.

src/app/models/User.ts
Allcharles: I can file one, but changing this behaviour on the baw-server would require making the opposite adjustment on the baw-client (ie. removing /assets from imageUrls). The change will have to wait until both are merged.

Allcharles:
A lot of your suggestions (especially those affecting models) will have to be added in an upstream PR. Unfortunately because this PR is almost 2 weeks old a few of the models you have mentioned have been changed when i implemented decorators. I'll try collate a list of commits relevant to this review below.

src/app/models/User.ts
Allcharles: This is required currently because of the `handleAuth` function in the security service. When a user logs in, i get a User model which i then convert into a Session User model: https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/services/baw-api/security.service.ts#L84-L90. Potentially there is a better was of doing this, however my assumption was that the server would be able to take only the values it needed.

atruskie: The only model that should have the auth token is the current user model. Even then,l it doesn't need it - the auth token shouldn't be part of any User model.

Thus:
- Ensure the current session/user is passed to the security service (injected, not as part of anyhting that will get toJSON'ed)
- access the the `user.authToken` as you would normally.

Allcharles: - The `authToken` is not part of the `User` model. It is only part of the `SessionUser` model.
- I'm not sure what you mean by injecting the current `SessionUser` model into the security service, as the service is what it generating the model
- Removing the functionality from the `toJSON()` in the `User` model would still not work because I need to store the model inside localstorage and `JSON.stringify` will use the output of `toJSON()`

atruskie: Ah local storage serialization. Fair enough, carry on

Allcharles:
Fix AnalysisJob review comments: 93f3e69e7da50be6c70787a50efc6e7da31638a6

src/app/models/AudioEventTag.ts
Allcharles: Ill throw an error instead of `null`, it'll be caught easier if a view is ever made.

src/app/models/Tag.ts
Allcharles: We agreed previously that the type of tags may be changed in the future and may become an api request. This is because of the `typeofTag`s is used as a dropdown field in a couple places.

src/app/models/Tag.ts
Allcharles: Its shown in the tags list view and i didnt want to create 25 api requests when i display the list of tags. If its not needed, we can remove it from the view: https://staging.ecosounds.org/admin/tags which also removes the issue

atruskie: leave it commented - it's not a field we'd unusually return since it is an aggregate value

Allcharles: As in comment out the key, or leave a comment explaining that its in limbo? The value is already in use on the tag list page so commenting it out would require me to modify that too.

atruskie: Comment in limbo is fine.

This will have to be revisited because I do expect that field will  be output by the API. It contravenes the API conventions.

It's a little different for the rails view: it can do arbitrary queries anywhere anytime - which is what it does. _count_ is not part of the model server side either.

atruskie (approved):

PR #223, opened 2020-04-20
Association loading

Allcharles:
# Association loading baw models

Created association loading for the baw models so that they can load each other when required.

### Association Decorators

- Created `@HasMany` decorator
  - Param 1: Service Token
  - Param 2: Key to read values from
  - Param 3: Filter key, defaults to "id"
  - Param 4-n: Optional model parameters, ie. Site.filter(siteId, projectId), default to []
```typescript
export class Project extends AbstractModel implements IProject {
  public readonly siteIds?: Ids;
  @HasMany(SHALLOW_SITE, (m: Project) => m.siteIds, [OPTIONAL] "customKey", [OPTIONAL] (m: Project) => m.id)
  public sites?: Observable<Site[]>;
}
```
- Created `@HasOne` decorator
  - Param 1: Service Token
  - Param 2: Key to read values from
  - Param 3-n: Optional model parameters, ie. Site.filter(siteId, projectId), default to []
```typescript
export class Project extends AbstractModel implements IProject {
  public readonly creatorId?: Id;
  @HasOne(ACCOUNT, (m: Project) => m.creatorId, [OPTIONAL] (m: Project) => m.id)
  public creator?: Observable<User>;
}
```

### Parameter Decorators

- Created `@BawPersistAttr` decorator
```typescript
export class Project extends AbstractModel implements IProject {
  @BawPersistAttr
  public readonly id?: Id;
}
```
- Created `@BawCollection` decorator
```typescript
export class Project extends AbstractModel implements IProject {
  @BawCollection()
  public readonly siteIds?: Ids;
}
```
- Created `@BawDateTime` decorator
```typescript
export class Project extends AbstractModel implements IProject {
  @BawDateTime()
  public readonly createdAt?: DateTimeTimezone;
}
```
- Created `@BawDuration` decorator
```typescript
export class Project extends AbstractModel implements IProject {
  @BawDuration()
  public readonly duration?: Duration;
}
```

#### Options

- Currently these decorators also provide the following options:
  - persist: Insert the parameter into the `toJSON()` output of the model
    ```typescript
    @BawDuration({persist: true})
    ```
  - key: Read the value of the parameter from another key
    ```typescript
    @BawDuration({key: "durationSeconds"})
    ```

## Issues

- [x] Would like the `@HasMany` and `@HasOne` decorators to update the typing of the model instead of forcing the user to do so. May be possible through `reflect-metadata`, looking for ideas.
-  [x] Could potentially create a `@BawOutput` decorator which automatically creates the `toJSON()` function for models
- [x] Would like to create `@BawDateTime`, `@BawDuration`, and `@BawCollection` decorators to remove the need to convert ids, dates, and durations inside the constructor. However this appear to be locked as the parameters are set to readonly even when readonly property is removed. This could be done by making all properties prive (using the _property convention), and have decorators create public getters, however this loses some of the robustness of readonly vs getter, and increases the memory cost of a model.
- [x] Missing unit tests


src/app/models/AbstractModel.ts
```diff
@@ -73,3 +83,100 @@ export abstract class AbstractModel {
...
+ * @param serviceToken Injection token for API Service
+ * @param key Key to match ids against
+ */
+export function HasMany<M extends AbstractModel, T extends ApiFilter<M, any>>(
+  serviceToken: InjectionToken<any>,
+  key: string = "id"
+) {
+  return function (target: AbstractModel, field: string) {
+    // Extract association name (ie. siteIds => sites)
+    const association = field.replace("Ids", "s");
```
atruskie: So I'd defintiely like to make the pluralization manual. It get's really tricky with different word roots and we only have to do this a few dozen times.

The HasMany function should accept the name of the new field

atruskie: better yet, a typed lambda pointing to the field to populate?

```
(project) => project.sites
```

src/app/models/AbstractModel.ts
```diff
@@ -73,3 +83,100 @@ export abstract class AbstractModel {
...
+      // If field is undefined, return
+      const params = this[property];
+      if (params === undefined || params === null) {
+        return of(null);
+      }
+
+      // If result cached, return
+      const cache: object = this[AbstractModel["associationsKey"]];
+      if (cache.hasOwnProperty(associationKey)) {
+        return of(cache[associationKey]);
```
atruskie: hmm, this cache is a litte ambiguous - a comment on how it works?

src/app/models/AbstractModel.ts
```diff
@@ -73,3 +83,100 @@ export abstract class AbstractModel {
...
+      }
+
+      // Create service and request from API
+      const service = injector.get<T>(serviceToken);
+      return createRequest(service, params).pipe(
+        map((model: M | M[]) => {
+          // Save to cache and return model
+          cache[associationKey] = model;
```
atruskie: i'd prefer this cache to just be setting a 'private' field, e.g. `model._sites = xxx`

src/app/models/Project.ts
```diff
@@ -33,17 +40,27 @@ export class Project extends AbstractModel implements IProject {
...
   public readonly ownerId?: Id;
-  public readonly description?: Description;
+  @HasMany<Site, ShallowSitesService>(SHALLOW_SITES_SERVICE)
+  public readonly siteIds?: Ids;
+
+  // Associations
+  public readonly sites?: Observable<Site[]>;
+  public readonly creator?: Observable<Id>;
+  public readonly updater?: Observable<Id>;
+  public readonly owner?: Observable<Id>;
```
atruskie: Why is this not `Observable<User>`?

src/app/models/Project.ts
```diff
@@ -33,17 +40,27 @@ export class Project extends AbstractModel implements IProject {
...
   public readonly updaterId?: Id;
   public readonly updatedAt?: DateTimeTimezone;
+  @HasOne<User, AccountService>(ACCOUNT_SERVICE)
   public readonly ownerId?: Id;
-  public readonly description?: Description;
+  @HasMany<Site, ShallowSitesService>(SHALLOW_SITES_SERVICE)
+  public readonly siteIds?: Ids;
+
+  // Associations
+  public readonly sites?: Observable<Site[]>;
```
atruskie: Why is this not `Observable<Site>`?

atruskie (changes requested):
RE:
> Would like the @HasMany and @HasOne decorators to update the typing of the model instead of forcing the user to do so. May be possible through reflect-metadata, looking for ideas.

That's a no go. See https://github.com/microsoft/TypeScript/issues/4881

> Could potentially create a @BawOutput decorator which automatically creates the toJSON() function for models

Nice idea. Maybe a better name is needed, but nice idea.

> Would like to create @BawDateTime, @BawDuration, and @BawCollection decorators to remove the need to convert ids, dates, and durations inside the constructor

I don't understand the issue here. Let's set up a call.

---

Your solution is nice and seems robust. 
I'm going to suggest a few changes:

- Move the association decorators to the association fields. I.e. `HasMany` should sit on on `public readonly sites`
- I'm not sure why `sites` and other associations are readonly. Actually I know why, because the decorator runs when the object is created thus they can be
   - They're meant to be lazy, so it doesn't matter if they're not `readonly` - unless we're going for an immutable vibe
   - they can stay readonly, as long as we gurantee API requests aren't fired of by those observables until something iterates/subscribes to them

- If moving association decorator to association field, you can then use a lambda to get the `id` or `ids` needed for the service request. This avoids the pluralization statement you have in each method. I.e. `HasMany<Sites, ShallowSiteService>(SHALLOW_SITES_SERVICE)`, becomes `HasMany<Sites, ShallowSiteService>(SHALLOW_SITES_SERVICE, (p) => p.siteIds)`. It is more verbose but it is also self-ducmenting.

- Speaking of verbosity, I wonder if we can make the decorator shorter by just supplying the DI token, and using a `switch case` in the getter code you wrote that'll have the right typing (along with some typescript type guards). I'd normally dislike this approach, but we're stuck with the tokens for now. So We'd end up with: `HasMany<Sites>(SHALLOW_SITES_SERVICE,  (p) => p.siteIds))`.
If the decorator is moved to the association, we'd have the type of the association property as well. I wonder if it could become:  `HasMany(SITES,  (p) => p.siteIds))`.

src/app/models/Project.ts
Allcharles: Its an array of sites that gets returned

src/app/models/AbstractModel.ts
Allcharles: I've cleaned up the caching code a bit, hopefully its more readable.

src/app/models/AbstractModel.spec.ts
```diff
@@ -0,0 +1,722 @@
...
+      const date = DateTime.fromISO("2019-01-01T00:00:00");
+      const model = createModel(["date"], {
+        id: 1,
+        date,
+      });
+      expect(model.toJSON()).toEqual({ date: date.toISO() });
+    });
+
+    it("should handle Duration", () => {
+      const duration = Duration.fromMillis(100000);
```
atruskie, line 87: nit: we have consistent philosophy that time (and all units) should only ever be measured in un-scaled SI units. Change to seconds rather than milliseconds.

src/app/models/AbstractModel.spec.ts
```diff
@@ -0,0 +1,722 @@
...
+    it("should retrieve empty metadata", () => {
+      expect(model.getMetadata()).toBe(undefined);
+    });
+  });
+});
+
+describe("Attribute Decorators", () => {
+  describe("BawPersistAttr", () => {
+    it("should append key to model attributes", () => {
+      class MockModel extends AbstractModel {
```
atruskie, line 150: i think a separate test file per decorator would make these much easier to parse

src/app/models/AbstractModel.spec.ts
```diff
@@ -0,0 +1,722 @@
...
+      const model = createModel(null);
+      expect(model.date).toEqual(null);
+    });
+
+    it("should convert timestamp string", () => {
+      const model = createModel("2019-01-01T00:00:00");
+      expect(model.date).toEqual(DateTime.fromISO("2019-01-01T00:00:00"));
+    });
+  });
+
```
atruskie, line 325: I can't think of a spot where we actually need to persist datetimes, but round-tripping the timezone offset would be important if we were.

E.g. if `2020-04-27T12:20:00+10:00` were sent then it would be important to send `2020-04-27T12:20:00+10:00` back.

Places where this is **NOT** important: any of the created_at, updated_at, or deleted_at fields (any timestamp the records an action relevant to the server/database.

Places where this **is** important: `AudioRecording.recordedAt` - any date that represents a date of something happening in the real world.

src/app/models/AbstractModel.ts
```diff
@@ -24,6 +30,11 @@ export abstract class AbstractModel {
    */
   public readonly kind: string;
 
+  /**
+   * Model attributes. This is used to generate the toJSON() output of the model.
+   */
+  private _attributes: string[];
```
atruskie, line 36: use a symbol ?

private [attributesSymbol]: string[]

src/app/models/AnalysisJob.ts
```diff
@@ -37,62 +46,63 @@ export interface IAnalysisJob {
...
-    // return duration(this.overallDurationSeconds) // TODO Awaiting PR #177
-  }
+  // Associations
+  @HasOne(SCRIPT, (m: AnalysisJob) => m.scriptId)
+  public script?: Observable<Script>;
+  @HasOne(ACCOUNT, (m: AnalysisJob) => m.creatorId)
+  public creator?: Observable<User>;
+  @HasOne(ACCOUNT, (m: AnalysisJob) => m.updaterId)
+  public updater?: Observable<User>;
+  @HasOne(ACCOUNT, (m: AnalysisJob) => m.deleterId)
```
atruskie, line 91: these are probably common enough you *could* define defaults

atruskie (approved):

src/app/models/AbstractModel.spec.ts
Allcharles: That is, to my understanding of my code, currently how the system works. The value comes in as `2020-04-27T12:20:00+10:00`, and goes back as `2020-04-27T12:20:00+10:00`. This is because the `toJSON()` method in the `AbstractModel` handles conversion of javascript types back to their value. Ie, Converting DateTime objects back to their ISO format.

atruskie: Great, add a unit test please

Issue #191, opened 2020-04-15
Baw client parity


Allcharles:
(M) = Migrated

List of pages to replicate in workbench-client before parity is reached with baw server:

- [ ] Recent Recordings
  - [ ] (M) Home (`/listen`) https://github.com/QutEcoacoustics/workbench-client/issues/206
  - [ ] (M) Listen (`/listen/:recordingId`) https://github.com/QutEcoacoustics/workbench-client/issues/207
- [ ] Library
  - [ ] (M) Home (`/library`) https://github.com/QutEcoacoustics/workbench-client/issues/208
  - [ ] (M) Library Item (`/library/:recordingId/audio_events/:audioEventId`) https://github.com/QutEcoacoustics/workbench-client/issues/209
- [ ] (M) Visualize (`/visualize`) https://github.com/QutEcoacoustics/workbench-client/issues/210
- [ ] Citizen Science
  - [ ] (M) About Study (`/citsci/:studyName`) https://github.com/QutEcoacoustics/workbench-client/issues/215
  - [ ] (M) Listen (`/citsci/:studyName/listen`) https://github.com/QutEcoacoustics/workbench-client/issues/216
  - [ ] (M) Listen Id (`/citsci/:studyName/listen/:sampleNum`) https://github.com/QutEcoacoustics/workbench-client/issues/217
  - [ ] (M) Responses (`/citsci/:studyName/responses`) https://github.com/QutEcoacoustics/workbench-client/issues/218
- [ ] Analysis Jobs
  - [ ] (M) List (`/audio_analysis`)  https://github.com/QutEcoacoustics/workbench-client/issues/211
  - [ ] (M) New (`/audio_analysis/new`) https://github.com/QutEcoacoustics/workbench-client/issues/212
  - [ ] (M) Details (`/audio_analysis/:analysisJobId`) https://github.com/QutEcoacoustics/workbench-client/issues/213
  - [ ] (M) Analysis Results (`/audio_analysis/:analysisJobId/results:path*?`) https://github.com/QutEcoacoustics/workbench-client/issues/214

atruskie:
Priority:

1. Visualize
2. CS
3. Analysis Jobs
4. Library

Do not start: Listen. More planning needed.

PR #227, opened 2020-04-30
Created missing BAW services

### LINKED ISSUE #219: Known Baw Services
Create the following services:

- [x] Account (`/user_accounts`)
- [x] Analysis Jobs (`/analysis_jobs`)
- [x] Analysis Job Items (`/analysis_jobs/:analysis_job_id/audio_recordings`)
- [ ] Analysis Results (`/analysis_jobs/:analysisJobId/results{path}`)
- [x] Audio Events (`/audio_recordings/:audioRecordingId/audio_events`)
- [x] Audio Event Tag (`/audio_recordings/:recordingId/audio_events/:audioEventId/tags/filter`)
- [x] Audio Recording (`/audio_recordings/:audioRecordingId`)
- [x] Bookmarks (`/bookmarks`)
- [x] Dataset (`/datasets`)
- [x] Dataset Items (`/datasets/:datasetId/item`)
- [x] Progress Events (`/progress_events`)
- [x] Projects (`/projects`)
- [x] Questions (`/questions` `/studies/:studyId/questions`)
- [x] Responses (`/responses` `/studies/:studyId/responses`)
- [x] Saved Searches (`/saved_searches`)
- [x] Scripts (`/scripts`)
- [x] Sites (`/sites` `/projects/:id/sites`)
- [x] Studies (`/studies`)
- [x] Tags (`/tags`)
- [x] Tag Groups (`/tag_groups`)
- [x] Taggings (`/audio_recordings/:recordingId/audio_events/:audioEventId/taggings`)
- [x] User (`/my_account`)

Allcharles:
# Created missing BAW services

Implemented all of the known missing BAW services so that association loading can occur. This is also a precursor to the requirements for #191.

## Changes

- Created new services

## Issues

- Missing Analysis Results services
- Haven't implemented custom routes (ie. /download)
- Haven't verified all routes are functional on BAW server

## Closes

Closes #219


src/app/models/Response.ts
```diff
@@ -0,0 +1,60 @@
...
+  creatorId?: Id;
+  createdAt?: DateTimeTimezone | string;
+}
+
+export class Response extends AbstractModel implements IResponse {
+  public readonly kind: "Answer" = "Answer";
+  @BawPersistAttr
+  public readonly id?: Id;
+  @BawPersistAttr
+  public readonly data?: Blob;
```
atruskie: this field should be JSON, just like `notes`. Probably best typed as `any` atm

src/app/models/User.ts
```diff
@@ -180,6 +178,7 @@ export interface ISessionUser extends IUser {
  * A user model for the website user
  */
 export class SessionUser extends User implements ISessionUser {
+  // ! All fields are persisted because model is saved to, and read from, localstorage
```
atruskie, line 124: Okay, I know not releated to this change but I'd like to address this.

Storing an entire user profile in local storage is way too much information.

I'm going to  nix SessionUser extending from User.

We really only need to store a session token (AuthToken) - when a session is resumed we can fetch the full user object/profile from the server. If it is conveinient, we can also store user id, user name, and image. But i'd prefer to store as little as possible.

src/app/services/baw-api/audio-event-tags.service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { stringTemplate } from "@helpers/stringTemplate/stringTemplate";
+import { AudioEvent } from "@models/AudioEvent";
+import { AudioEventTag } from "@models/AudioEventTag";
+import { AudioRecording } from "@models/AudioRecording";
+import { Observable } from "rxjs";
+import { ApiFilter, Filter, id, IdOr, IdParam, option } from "./api-common";
+import { BawApiService, Filters } from "./baw-api.service";
+
+const audioRecordingId: IdParam<AudioRecording> = id;
+const audioEventId: IdParam<AudioEvent> = id;
```
atruskie, line 13: is it worth factoring these out? multiple urls will use the same named url parameters

atruskie (approved):
Why are all tests for the services in a `tests` sub folder? This departs from established conventions?

I'd make a new issue for requested changes to SessionUser. This is otherwise good to go.

src/app/services/baw-api/audio-event-tags.service.ts
Allcharles: yes and no, i would prefer to extract the actual endpoint so that the following could happen:

```typescript
// Current
const audioRecordingId: IdParam<AudioRecording> = id;
const audioEventId: IdParam<AudioEvent> = id;
const endpoint_v1 = stringTemplate`/audio_recordings/${audioRecordingId}/audio_events/${audioEventId}/tags/${option}`;

// Future
import {parentPath} from './parent';
const audioEventId: IdParam<AudioEvent> = id;
const endpoint_v2 = stringTemplate`${parentPath}/audio_events/${audioEventId}/tags/${option}`;
```

Unfortunately, i can't think of a simple solution to this as i'm not familiar enough with template literals just yet. You're solution would be potentially more dangerous as there is a big difference between `IdParam` and `IdParamOptional` which would be hard to detect when the value is imported.

src/app/models/User.ts
Allcharles: I have tried to reduce the stored information already to what i suspect will potentially be needed across the site. Here are the values stored and why:

- `id` => Self explanatory
- `email` => Should be removed
- `userName` => Used by header and menu links
- `imageUrls` => Used by header (could instead just request the User data in the header, but why duplicate requests?
- `preferences` => I had assumed system wide preferences were here and would potentially be needed in future
- `rolesMask` => Needed for menus
- `rolesMaskNames` => Should be removed (have rolesMask already)
- `timezoneInformation` => I had assumed this would be used system wide somewhere in future such as automatically converting times or time formatting
- `authToken` => Self explanatory

Something i have neglected, because it doesnt exist yet, is the difference between the `toJSON` outputs of `User` and `SessionUser`. I should probably re-enable some options on the `User` model which are editable via forms, and use the above for only the `SessionUser`.

Allcharles: Revision:
`User` model output:

- id
- email
- userName
- imageUrls
- preferences
- rolesMask
- timezoneInformation

`SessionUser` model output:

- id
- authToken
- userName
- imageUrls
- preferences
- rolesMask
- timezoneInformation

atruskie: > * userName
> * imageUrls
> * preferences
> * rolesMask
> * timezoneInformation
> 
> `SessionUser` model output:
> 
> * id
> * authToken
> * userName
> * imageUrls
> * preferences
> * rolesMask
> * timezoneInformation

Okay, the proposed model seems fine. Could we see a unit test that asserts *only* the required properties are saved in SessionUser?

I still have an issue with `SessionUser` extending `User`. User has a bunch of extra fields/properties that will never be populated for a SessionUser.

src/app/services/baw-api/audio-event-tags.service.ts
atruskie: That's something to look into, doesn't need to happen yet

src/app/models/User.ts
Allcharles: No need to add the assertion, I will use the `@BawPersistAttr` to create the `toJSON()` functions

atruskie: Normally, correct, yes, it would not need to be tested - we don't need to reproduce the tests for `@BawPersistAttr`. In this case however, I think it is important we don't store potentially sensitive information so we should test for that explicitly. Different use-case even if testing same fields.

Allcharles: I already have a test for this, I will update the test name to increase clarity.

PR #233, opened 2020-05-11
Admin orphan site pages

### LINKED ISSUE #134: Admin Orphan Sites
Create admin orphan sites page (`/admin/orphans`)

Allcharles:
# Admin Orphan Site Pages

Create admin orphan site pages.
@mFerroudj

## Changes

- Added images to RenderView options
- Re-enabled form tests (were accidentally disabled)
- Create admin orphan site pages

## Issues

- Testing for image render views is sub-optimal as there does not appear to be a simple method of waiting for the image to verify if it exists
- Setting form inputs as hidden does not stop formly from accessing the models property. This is a problem for the edit forms, as we use the schema to define both the form and the DetailView. With the current system, models loaded by association are requested by the form even though they are not used causing errors in the console and unit tests. Formly issue: https://github.com/ngx-formly/ngx-formly/issues/1215

## Closes

Relates to #134. Awaiting API update

## Visual Changes

Orphan Sites Page
image

Orphan Site Page
image


src/app/component/admin/orphan/list/list.component.html
```diff
@@ -0,0 +1,58 @@
+<ng-container *ngIf="rows && !error">
+  <h1>Orphan Sites</h1>
+  <p>
+    There are {{ totalModels }} orphan sites. A site becomes orphaned when it
+    can no longer be access from any project.
```
atruskie:
```suggestion
    can no longer be accessed from any project.
```

atruskie (approved):
> Setting form inputs as hidden does not stop formly from accessing the models property. This is a problem for the edit forms, as we use the schema to define both the form and the DetailView. With the current system, models loaded by association are requested by the form even though they are not used causing errors in the console and unit tests.

Will the `readonly` attribute fix this? If not, perhaps it's worth asking for a feature to ignore certain properties.

Allcharles:
I can see two simple solutions to the issue:

1. Part of me is really tempted to completely remove the .json files, and have the object auto-generated by some attribute decorators on the baw models. The only reason i haven't done so was because you wanted to extract the .json files to the server in the future. That would allow me to completely bypass the issue by generating different inputs for the different forms, and detail pages
2. I am going to try creating a custom formly element which doesn't touch the models value, hopefully this will stop the observable from being run

To clarify, the issue is not that the form attempts to edit the associated model. The issue is that it attempts to read the associated model. The model requires the injector value to access associated models, however during the new and edit forms, this value is not set causing errors to be thrown whenever a model is requested. It also causes a lot of unnecessary code in the test suites to intercept the model requests.

atruskie:
We've 

> I can see two simple solutions to the issue:
> 
> 1. Part of me is really tempted to completely remove the .json files, and have the object auto-generated by some attribute decorators on the baw models. The only reason i haven't done so was because you wanted to extract the .json files to the server in the future. That would allow me to completely bypass the issue by generating different inputs for the different forms, and detail pages
> 2. I am going to try creating a custom formly element which doesn't touch the models value, hopefully this will stop the observable from being run
> 
> To clarify, the issue is not that the form attempts to edit the associated model. The issue is that it attempts to read the associated model. The model requires the injector value to access associated models, however during the new and edit forms, this value is not set causing errors to be thrown whenever a model is requested. It also causes a lot of unnecessary code in the test suites to intercept the model requests.

We've discussed this offline; we'll split the schema into two parts. A base set that formly will use, and a fuller view of the model that other components can use.

atruskie (approved):

PR #235, opened 2020-05-13
Profile Page Cleanup

### LINKED ISSUE #135: My Projects
Create both my and their project pages (`/my_account/projects`, `/user_accounts/:id/projects`)

### LINKED ISSUE #136: My Sites

### LINKED ISSUE #137: My Bookmarks
Create both my and their bookmark pages (`/my_account/bookmarks`, `/user_accounts/:id/bookmarks`)

### LINKED ISSUE #192: Profile
Create both my and their profile pages (`/user_accounts/:userId`, `/my_account`).

Allcharles:
# Cleanup Profile Pages

Cleaned up a few of the profile pages.

## Changes

- Created my bookmarks page
- Created boilerplate for my annotations page
- Cleanup profile page to display values
- Updated `PagedTableTemplate` class
- Updated services to find models a specific user has access to

## Issues

- Project/Site services currently cannot retrieve the possible list of user accessible models because `accessLevel` has not been added as a filterable output
- Most popular tags currently not possible to generate as they do not provide a count for the number of times they have been used. Would have to take the list of all tags, calculate how many times they have been used, and then order @atruskie 

## Closes

Relates to #137 #136 #135 #192 

### Personal Account

Profile
image

Bookmarks
image

Projects
image

Sites
image

### User Account

Profile
image

Bookmarks
image

Projects
image

Sites
image


mFerroudj:
Hi Charles. For Sites: I thought we had 106 sites in total. Each site has 4 points and each point is a location where a sensor is deployed.

Allcharles:
The images taken are using the staging BAW server and will not match up with the actual values you will see on the official website. A lot of the models you will see in these images are mocked (ie. I have manually faked the data to get an output), which is why they show "PLACEHOLDER SITE".

Allcharles:
As for points, i have not yet integrated them into the system. It could be an idea to add a points as a profile metric. @atruskie Should we have projects and sites as metrics and switch to points and sites on the gov version, or have projects, sites, and points?

atruskie:
@mFerroudj remember this software serves both Ecosounds and the A2O. These pages show which projects, sites, and bookmarks a user has access to.

You have a valid point, we should add a points view. But we'll do that in a separate change. @mFerroudj  do you have any feedback on the look of these *particular* pages as is?
If your feedback is related to a different feature we won't discuss it in this pull request.

atruskie:
@Allcharles, re:

> Most popular tags currently not possible to generate as they do not provide a count for the number of times they have been used. Would have to take the list of all tags, calculate how many times they have been used, and then order @atruskie

For aggregate requests we will define custom endpoints. Map this requirement to an issue on baw-server. Something liek `/tags/statistics`

src/app/component/admin/tag-group/list/list.component.ts
```diff
@@ -57,6 +56,8 @@ export class AdminTagGroupsComponent extends PagedTableTemplate<
         model: tagGroup,
       }))
     );
+
+    this.filterKey = "groupIdentifier";
```
atruskie, line 60: If this does what I think it does, probably better to be `tag.text`

src/app/component/profile/pages/bookmarks/their-bookamrks.component.spec.ts
```diff
@@ -0,0 +1,72 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
```
atruskie: filename contains spelling mistake

src/app/component/shared/items/item/item.component.ts
```diff
@@ -47,7 +47,7 @@ import { StrongRoute } from "src/app/interfaces/strongRoute";
       </span>
     </div>
   `,
-  changeDetection: ChangeDetectionStrategy.OnPush
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
atruskie, line 51: I'd leave a comment explaining why we can use this strategy (immutable objects I presume?)

src/app/component/shared/items/items/items.component.ts
```diff
@@ -38,9 +39,9 @@ import { ItemInterface } from "../item/item.component";
       </div>
     </div>
   `,
-  changeDetection: ChangeDetectionStrategy.OnPush
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
atruskie, line 44: comment why

src/app/services/baw-api/api-common.ts
```diff
@@ -46,6 +46,30 @@ export function option(x?: New | Filter | Empty) {
...
+    ...meta,
+    filter: {
+      ...filter,
+      [key]: {
+        eq: model instanceof AbstractModel ? model.id : model,
+      },
+    },
+  };
+}
+
```
atruskie: This method is redundant for `service<T>` and `Model<T>`. You don't need `/filter` to get a specific object of the same type... you just use `/:id`

It is useful for `service<T>` and `Model<U>` - i.e. some other model that can be queried that is associated with `T`.

Add documentation to that effect maybe?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -255,11 +255,88 @@ export interface Paging {
   next?: string;
 }
 
+interface Combinations {
```
atruskie: Maybe `conditions`

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -255,11 +255,88 @@ export interface Paging {
...
+  greaterThanOrEqual?: number;
+  notGteq?: number;
+  notGreaterThanOrEqual?: number;
+}
+
+type Range =
+  | string[]
+  | number[]
+  | {
+      interval?: string;
```
atruskie: interval has a narrow syntax that is acceptable: `(\[|\()[-+]?[0-9]*\.?[0-9]+,[-+]?[0-9]*\.?[0-9]+(\]|\))`

May be able to strongly type it?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -255,11 +255,88 @@ export interface Paging {
...
+      from?: number;
+      to?: number;
```
atruskie: `from` and `to` are exclusive to `interval`

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -255,11 +255,88 @@ export interface Paging {
...
+/**
+ * Api response inner filter
+ */
+export type InnerFilter<T = undefined> =
+  | (Combinations &
+      Comparisons &
+      Subsets &
+      { [P in keyof T]?: Combinations & Comparisons & Subsets })
+  | {};
+
```
atruskie: Wow

src/app/services/baw-api/bookmark/bookmarks.service.ts
```diff
@@ -32,9 +34,18 @@ export class BookmarksService extends StandardApi<Bookmark> {
...
+  filter(filters: Filters<Bookmark>): Observable<Bookmark[]> {
     return this.apiFilter(endpoint(Empty, Filter), filters);
   }
+  filterByCreator(
+    filters: Filters<Bookmark>,
+    user?: IdOr<User>
+  ): Observable<Bookmark[]> {
+    return this.apiFilter(
+      endpoint(Empty, Filter),
+      user ? filterById<Bookmark>(filters, "creatorId", user) : filters
```
atruskie: reading this from GH I can't confirm, but that `creatorId` key is strongly typed? I think it is, nice!

atruskie (approved):
Minor nits, but on the whole a great change

src/app/services/baw-api/baw-api.service.ts
Allcharles: image

src/app/component/shared/items/item/item.component.ts
Allcharles: Tbh, i should convert this to take immutable objects only (although this isn't the reason for the change detection strategy). We can set this change detection strategy because its a pure component (only thing that changes is the input so it doesn't need to do change detection). Leaving a comment would be a bit overkill but i'll add it

src/app/services/baw-api/baw-api.service.ts
Allcharles: This is getting close to what i want. Typing isn't perfect yet as a programmer can still add keys that aren't part of the model, and values which do not match the key. Both should be possible to fix.

Allcharles: Also if i can get this working properly, it should be able to accept empty filters (`{}`). Currently i have a workaround with the `| {};` bit which im not overly happy with.

src/app/services/baw-api/baw-api.service.ts
Allcharles: Regex typing is not currently possible: https://github.com/Microsoft/TypeScript/issues/6579. Could use `Range` as type checking for an Interval class however its not recommended for performance reasons: https://github.com/microsoft/TypeScript/issues/15480
I saw that its possible to strong type the value using the following brute force "hack":
```typescript
type Interval = "[0, 1]" | "[0, 2]" | ... | ["999", "999"];
```
Depending on the size of the solution, it would be possible to use something like https://github.com/KendallHopkins/FormalTheory to generate all possible options and just store the type in a file.

Edit: Did a rough calc, and i think there are about 16 million possibilities for the regex (potentially an order of magnitude more)

src/app/services/baw-api/bookmark/bookmarks.service.ts
Allcharles: Yup, its strongly typed

src/app/services/baw-api/baw-api.service.ts
Allcharles: Fixed with the following change:

```typescript
export type InnerFilter<T = {}> = Combinations &
  Comparisons &
  Subsets &
  { [P in keyof T]?: Combinations & Comparisons & Subsets };
```

I still haven't added the ability for the filter to verify the subsets are valid for the key type, however I've decided that'd be over the top.

src/app/component/admin/tag-group/list/list.component.ts
Allcharles: `TagGroup` does not have a text parameter. `text` is attached to the `Tag` model

src/app/services/baw-api/api-common.ts
Allcharles:
```typescript
/**
 * Modify a base filter to add a foreign key condition
 * @param filters Base Filters
 * @param key Foreign key
 * @param model Foreign key value
 */
export function filterByForeignId<T extends AbstractModel>(
```

src/app/services/baw-api/bookmark/bookmarks.service.ts
Allcharles: Only thing it doesn't do is validate that the key provided is of type Id, specifically because I'm unsure how generic I want to make the function.

src/app/services/baw-api/baw-api.service.ts
atruskie: Fair enough

src/app/services/baw-api/baw-api.service.ts
atruskie: Yup we can't regex match the type. Could we just make it a type though, i.e a class?

Allcharles: Do you want the class to type check the input at run time, or basically just be a string wrapper?

src/app/component/shared/items/item/item.component.ts
atruskie: Well that why is exactly the context I was missing... hence the need.

src/app/services/baw-api/baw-api.service.ts
atruskie: It can do a bit of checking yeah

Allcharles:
> 
> 
> @Allcharles, re:
> 
> > Most popular tags currently not possible to generate as they do not provide a count for the number of times they have been used. Would have to take the list of all tags, calculate how many times they have been used, and then order @atruskie
> 
> For aggregate requests we will define custom endpoints. Map this requirement to an issue on baw-server. Something liek `/tags/statistics`

Updated issue: https://github.com/QutEcoacoustics/baw-server/issues/449

atruskie (approved):

mFerroudj:
1. I think the profile looks good. 
2. The way the sites are displayed looks better now although we still don't know which site has data unless we go through all sites
3. just remember that we have roughly 106 sites and each site contain 4 points where recorders are deployed

atruskie:
@mFerroudj :

> The way the sites are displayed looks better now although we still don't know which site has data unless we go through all sites

This change only shows sites a user has access to on their personal profile. It does not affect the main site listing page.

> just remember that we have roughly 106 sites and each site contain 4 points where recorders are deployed

The list is paged. But also we have projects on Ecosounds that have hundreds of sites. Those constratinsts are not unqiue.

Lastly, this Pull Request has closed. You missed the opportunity to review and provide feedback for this changeset.

PR #236, opened 2020-05-13
Decorator cleanup

### LINKED ISSUE #229: Improving model decorators
List of improvements for model decorators:

- [x] Possible to improve association decorator typing using `keyof` similar to Jasmine `spyOn` function where methods can be provided as a string representing the key:
  ```typescript
  @HasMany(SHALLOW_SITE, (m: Project) => m.siteIds, (m:Project) => m.secondKey, ...)
  // vs
  @HasMany(SHALLOW_SITE, "siteIds", "secondKey", ...)
  ```
  - This would somewhat simplify the Association Decorator code as the key can be used instead of the callback function
- [x] Possible to add typing to attribute decorator using a similar method to the one discussed above. Currently the key option is not type checked.
- [ ] ~Potentially standardise `@BawPersistAttr` decorator to follow the same system as other decorators. i.e.~
  ```typescript
  @BawAttribute({persist: true})
  ```
  - However this is based on the premise that `@BawAttribute` by itself would have some purpose
  - Could be an idea just to turn it into a function so that it is at least the same format as the other decorators
- [ ] Attempt to add more intelligent warnings, currently models will spam the console when an associated model's primary key is null or undefined
- [ ] Somehow remove the need for `InjectionTokens` (eg. `SHALLOW_SITE`), currently `InjectionTokens` are used as a workaround to prevent cyclic dependencies between models
- [x] Simplify associated model output.
  ```typescript
  projects.sites // observable
  // vs
  projects.sites // array
  ```
- [ ] Re-introduce access to associated model observable

Allcharles:
# Cleanup AbstractModel decorators

Added improved typing to AbstractModel decorators.

## Changes

- Extract Associated and Attribute decorators into separate files
- Association Decorator Changes:
  - Param 1: Service Token
  - Param 2: Key to read values from
  - Param 3: Filter key, defaults to "id"
  - Param 4-n: Optional model parameters, ie. Site.filter(siteId, projectId), default to []
  - Before
    ```typescript
    export class Project extends AbstractModel implements IProject {
      public readonly siteIds?: Ids;
      @HasMany(SHALLOW_SITE, (m: Project) => m.siteIds, [OPTIONAL] "customKey", [OPTIONAL] (m: Project) => m.id)
      public sites?: Observable<Site[]>;
    }
    ```
  - After
    ```typescript
    export class Project extends AbstractModel implements IProject {
      public readonly siteIds?: Ids;
      @HasMany(SHALLOW_SITE, "siteIds", [OPTIONAL] "customKey", [OPTIONAL] "id")
      public sites?: Observable<Site[]>;
    }
    ```
- Attribute decorator added typing to override key option

## Issues

- Couldn't get Association decorators to type check if input key contains Id/Ids. Expected solution to be something similar to the following, however it doesn't work:
  - Current:
    ```typescript
    export function HasOne<M extends AbstractModel>(
      serviceToken: ServiceToken<...>,
      modelIdentifier: keyof M,
      ...modelParameters: ReadonlyArray<keyof M>
    )
    ````
  - Potential?
    ```typescript
    export function HasOne<M extends AbstractModel, K extends keyof M = keyof M>(
      serviceToken: ServiceToken<...>,
      modelIdentifier: M[K] extends Id ? K : never,
      ...modelParameters: ReadonlyArray<keyof M>
    )
    ````

## Closes

Relates to #229


atruskie (approved):
Very nice

PR #237, opened 2020-05-18
Annotations Service

Allcharles:
# Annotations Service

Created annotations service

### DIFF
--- a/src/app/models/Annotation.ts
+++ b/src/app/models/Annotation.ts
@@ -6,7 +6,7 @@ import {
   SHALLOW_SITE,
   TAG,
 } from "@baw-api/ServiceTokens";
-import { Id, Ids } from "@interfaces/apiInterfaces";
+import { Id } from "@interfaces/apiInterfaces";
 import { BehaviorSubject, Observable } from "rxjs";
 import { flatMap, map } from "rxjs/operators";
 import { AbstractModel } from "./AbstractModel";
@@ -17,34 +17,25 @@ import type { Site } from "./Site";
 import type { Tag } from "./Tag";
 
 export interface IAnnotation {
-  id: number;
-  // Filter by audioEventTag.audioEventId
-  audioEventTagIds?: Ids;
-  tagIds?: Ids;
-  // Show using audioEvent.audioRecordingId
-  audioRecordingId?: Id;
-  // Show using audioRecording.siteId
-  siteId?: Id;
+  id: Id;
+  audioEvent: Observable<AudioEvent>;
+  audioEventTags: Observable<AudioEventTag[]>;
+  tags: Observable<Tag[]>;
+  audioRecording: Observable<AudioRecording>;
+  site: Observable<Site>;
 }
 
 export class Annotation extends AbstractModel implements IAnnotation {
   public readonly kind: "Annotation" = "Annotation";
-  public id: number;
-  public audioEventTagIds?: Ids;
-  public tagIds?: Ids;
-  public audioRecordingId?: Id;
-  public siteId?: Id;
+  public id: Id;
   private _audioEvent?: Observable<AudioEvent>;
   private _audioEventTags?: Observable<AudioEventTag[]>;
   private _tags?: Observable<Tag[]>;
   private _audioRecording?: Observable<AudioRecording>;
   private _site?: Observable<Site>;
 
-  constructor(annotation: IAnnotation, injector: Injector) {
-    super(annotation, injector);
-  }
-
-  public setAudioEvent(audioEvent: AudioEvent) {
+  constructor(audioEvent: AudioEvent, injector: Injector) {
+    super({ id: audioEvent.id }, injector);
     this._audioEvent = new BehaviorSubject<AudioEvent>(audioEvent);
   }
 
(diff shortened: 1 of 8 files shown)


Allcharles:
It could be an idea to increase the power of the association decorators so that they can handle custom filters and nested relationships, however i shied away because it was simpler to implement the one off logic in the Annotation model. If there are other such associations it might be good to implement.

Allcharles:
Reviewing the ecosounds AngularJS client the following calls are made:

- Call `/audio_events/filter`
- Call `/tags/filter` (this skips the `AudioEventTag` request)
  ```
  {"filter":{"audio_events.id":{"in":[269917]}}}
  ```
- Call `/audio_recordings/filter`
  ```
  {"filter":{"id":{"in":[623590]}},"projection":{"include":["id","siteId","durationSeconds","recordedDate"]}}
  ```
- Use `AudioRecording` to find `Site`
- Use `Site` to find `Project`

Consistently minimized the request sizes by using projection to only request wanted fields.

src/app/models/Annotation.ts
```diff
@@ -0,0 +1,148 @@
...
+import type { AudioEventTag } from "./AudioEventTag";
+import type { AudioRecording } from "./AudioRecording";
+import type { Site } from "./Site";
+import type { Tag } from "./Tag";
+
+export interface IAnnotation {
+  id: Id;
+  audioEvent: Observable<AudioEvent>;
+  audioEventTags: Observable<AudioEventTag[]>;
+  tags: Observable<Tag[]>;
```
atruskie, line 23: I don't think it's necessary to have both of these.

I'd suggest keeping audioEventTags since that model had meta information about when the tag was applied to the event.

src/app/models/Annotation.ts
```diff
@@ -0,0 +1,148 @@
...
+  audioRecording: Observable<AudioRecording>;
+  site: Observable<Site>;
+}
+
+export class Annotation extends AbstractModel implements IAnnotation {
+  public readonly kind: "Annotation" = "Annotation";
+  public id: Id;
+  private _audioEvent?: Observable<AudioEvent>;
```
atruskie, line 31: This is not a standalone model.

It should extend AudioEvent, it should not take an AudioEvent as a field.

An _annotation_ is an _audio event_ with one more tags.

src/app/models/Annotation.ts
```diff
@@ -0,0 +1,148 @@
...
+        const site = sites.length > 0 ? sites[0] : null;
+        this._site = new BehaviorSubject<Site>(site);
+        return site;
+      })
+    );
+    return this.site;
+  }
+
+  public get viewUrl(): string {
+    throw new Error("Method not implemented.");
```
atruskie, line 146: this would be the library single item page eventually

atruskie (changes requested):
I think you've modelled this incorrectly.

An annotation is just an audio event. We use that name because it has semantic meaning (to users, and publications) and we've previously used that name because it was useful to have a separate class that was used (that had extra/relevant properties) in editing audio events on the listen page.

It is important that the following semantics hold.

```
let a: Annotation = /* ... */
a.startTimeSeconds;  // 123
a.site.id; // 4
a.site.project.name; // "foobar"
a.tags; // [ Tag, Tag, Tag]
```

src/app/models/Annotation.ts
Allcharles: `AudioEventTag` only gives the following values:

- id
- audioEventId
- tagId
- creatorId/At
- updaterId/At

I need the tag text a lot of the time. Is there something im missing?

Allcharles:
> 
> 
> I think you've modelled this incorrectly.
> 
> An annotation is just an audio event. We use that name because it has semantic meaning (to users, and publications) and we've previously used that name because it was useful to have a separate class that was used (that had extra/relevant properties) in editing audio events on the listen page.
> 
> It is important that the following semantics hold.
> 
> ```
> let a: Annotation = /* ... */
> a.startTimeSeconds;  // 123
> a.site.id; // 4
> a.site.project.name; // "foobar"
> a.tags; // [ Tag, Tag, Tag]
> ```

The semantics you have provided are in direct contrast with how all other models work where associated models are accessed through an observable. I can make the children non-observables, however the Association model would no longer extend the AudioEvent model because that already has definitions for audioRecording and audioEventTags using observables

src/app/models/Annotation.ts
atruskie: There are different use cases.

If you're showing annotations on some summary page then: `audioEvent.audioEventTags.select(aet => aet.tag.text)` is approproate.

If you're editing an anntoation, you don't need text at all: `audioEvent.audioEventTags.push( /*new value*/)`

Regardless both are associations (one is transitive) of the audio event.

atruskie:
>  direct contrast with how all other models work where associated models are accessed through an observable

Not so much direct contrast. The observables are inthe end just a wrapper for the objects. The example above omits the wrapper but the relational semantics should be equivalent.
I'd prefer there was no wrapper at all.

>  I can make the children non-observables, however the Association model would no longer extend the AudioEvent model because that already has definitions for audioRecording and audioEventTags using observables

Do both!

Allcharles:
PR blocked until some further changes are made to the association decorators.

Allcharles:
Closed in favour of #567

PR #239, opened 2020-06-01
Unresolved models

### LINKED ISSUE #229: Improving model decorators

Allcharles:
# Modify associated model typing

Removed wrapping observable for associated models. Change to models can be seen below:

```typescript
// before
projects.sites // Observable<Site[]>
// after
projects.sites // Array<Site>
```

## Changes

- Renamed `AudioRecordingComponent` from detail.ts to details.ts so it matches the naming convention of other files

## Issues

- Removed access to associated model observable, however it should take minimal effort in the future to re-add if required

## Closes

Related to #229


src/app/component/shared/detail-view/detail-view.component.spec.ts
```diff
@@ -150,9 +150,36 @@ describe("DetailViewComponent", () => {
...
+        expect(true).toBeFalsy();
+      });
+
+      it("should handle hasOne unresolved associated model", () => {
+        expect(true).toBeFalsy();
+      });
+
+      it("should handle hasMany unresolved associated model", () => {
+        expect(true).toBeFalsy();
+      });
```
atruskie: I assume these are intentionally failing?

src/app/component/shared/detail-view/render-field/render-field.component.ts
```diff
@@ -43,13 +43,14 @@ import { toRelative } from "src/app/interfaces/apiInterfaces";
...
+        [field]="child"
       ></baw-render-field>
     </ng-template>
   `,
 })
-export class RenderFieldComponent extends WithUnsubscribe() implements OnInit {
-  @Input() view: ModelView;
+export class RenderFieldComponent extends WithUnsubscribe()
+  implements OnChanges {
+  @Input() field: ModelView;
```
atruskie: Given the number of places where you had used `value` below I feel that's an indication to what the name of this property should be. Suggested name: `value`

src/app/models/AbstractModel.ts
```diff
@@ -101,3 +103,39 @@ export abstract class AbstractModel {
...
+  private static readonly models = [];
+
+  static get one(): UnresolvedModel {
+    return this.model;
+  }
+
+  static get many(): AbstractModel[] {
+    return this.models;
+  }
+}
```
atruskie: Merge these definitions into `UnresolvedModel`, remove `SingletonUnresolvedModel`

src/app/models/AbstractModel.ts
```diff
@@ -101,3 +103,39 @@ export abstract class AbstractModel {
     return this[AbstractModel.metaKey];
   }
 }
+
+export class UnresolvedModel extends AbstractModel {
+  public readonly resolved = false;
+  public readonly kind = "UnresolvedModel";
+
+  constructor() {
```
atruskie, line 115: add explicit accessibility modifier?

src/app/models/AbstractModel.ts
```diff
@@ -101,3 +103,39 @@ export abstract class AbstractModel {
...
+
+  static get many(): AbstractModel[] {
+    return this.models;
+  }
+}
+
+export function isResolvedModel(
+  model: AbstractModel | UnresolvedModel
+): model is UnresolvedModel {
+  return model.resolved;
```
atruskie: I think we should just do a type check here and remove the `resolved` property completely.

src/app/models/AssociationDecorators.ts
```diff
@@ -165,17 +164,22 @@ function createModelDecorator<M extends AbstractModel, S>(
...
     // Create service and request from API
     const service = injector.get(serviceToken.token);
-    const request = createRequest(service, identifier, parameters).pipe(
-      map((response) => {
-        // Change cached value to BehaviorSubject so that it will instantly return
-        updateCache(target, cachedModelKey, new BehaviorSubject(response));
-        return response;
-      })
+    createRequest(service, identifier, parameters).subscribe(
+      (model) => updateCache(target, cachedModelKey, model),
```
atruskie: Ok, this is great, however, if the observeable emits more than once we'll only get the last value.

Why might it omit more than once? Say we want the first two pages of API results to load by default for all associations.

Also, `updateCache` should probably be renamed to `updateBackingField`

atruskie (changes requested):
Looks great! I was confused by how simple the changes ended up being. Nice work.

Minor changes needed for naming and esoteric corner case.

src/app/models/AbstractModel.ts
Allcharles: Am i able to do that whilst also keeping the typings for AbstractModel the same as the AbstractModel?

atruskie: Should be able to? Nothing changes really. You can have static and non static members in one class... Therefore no reason to have a class with only static members in it.

Yes a static method can instantiate an instance of a type it is attached to.

src/app/models/AssociationDecorators.ts
Allcharles: So if the response is in array, append to the array. What about single responses? Just override still?

atruskie: Yeah, override. I can't think of any other reasonable behaviour

atruskie (approved):
I'd suggest moving the abstract model tests from the detail view component. The tests are really about testing the AbstractModel association loading, and the detail view component is just a handy vehicle test bed.

The settimout fake style async is a little code-smelly. I've talked with you offline about it and I suggest dropping the interval to 0 where possible and also making a named function that makes the code easier to read. Something  like
```
await WaitOne()
```

Allcharles:
> I'd suggest moving the abstract model tests from the detail view component. The tests are really about testing the AbstractModel association loading, and the detail view component is just a handy vehicle test bed.

So move the tests into the AssociatedDecorator test file?

atruskie:
> > I'd suggest moving the abstract model tests from the detail view component. The tests are really about testing the AbstractModel association loading, and the detail view component is just a handy vehicle test bed.
> 
> So move the tests into the AssociatedDecorator test file?

For tests that are primarily related to the associations, yeah.

Another file is fine too if they're a different style of test. I'd be tempted to name it something like: `AssociationLoadingInComponents.spec.ts`

Allcharles:
Okay, i wasn't a huge fan of extracting the tests as they were to `AssociationLoadingInComponents.spec.ts`, so i created a mock component so that future changes to the `DetailViewComponent` do not affect the tests. I have also kept the tests inside `DetailViewComponent` as they are still relevant to the component.

atruskie (approved):

PR #241, opened 2020-06-09
Async test cleanup

Allcharles:
# Implement async/await tests

Cleanup `setTimeout` and `tick()` based tests. This was replaced by the ability to create a promise which resolves once the observables `next()` or `error()` function is called.

## Changes

- Created `nStepsObservable` function which returns a Promise which will resolve after n `testStepIntervals` occur
- Created `assertImage` function to verify image `src` and `alt`
- Updated old tests to use new method
- Re-wrote the following component tests:
  - `HomeComponent`
  - `UserBadgeComponent`
  - `AppComponent`


src/app/test/helpers/general.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Create an observable which will return after a number of steps
+ * @param subject Subject to update
+ * @param output Return value for subject
+ * @param isError Call `subject.error()`
+ * @param steps Number of interval steps to wait (default = 0)
+ */
+export function nStepObservable<T>(
+  subject: Subject<T>,
+  output: T | ApiErrorDetails,
```
atruskie: this should probably be a callback (or at least support an optional callback, or immediate value), otherwise the value that is returned after the async queue will have been computed *before* the method was even called.

src/app/test/helpers/general.ts
```diff
@@ -0,0 +1,30 @@
+import { ApiErrorDetails } from "@baw-api/api.interceptor.service";
+import { Subject } from "rxjs";
+
+/**
+ * Default step size for timeouts
+ */
+export const testStepInterval = 16;
```
atruskie: Could this be 0?

If you're wondering how to do n steps (where n > 0) then you do multiple `setTimeouts`, one for each `steps`

atruskie: If my suggestion (or assertion does not work, then leave a comment explaining why you settled on 16)

atruskie (changes requested):

src/app/test/helpers/general.ts
Allcharles: So each step triggers the next step recursively?

Ie.

```typescript
setTimeout(() => {
  // step 0
  setTimeout(() => {
    // step 1
  }. 0);
}, 0);
```

atruskie: yeah recursive, though more like:

```
var waitOne = setTimout(() => {
  if (stepsRemaining >0) {
    stepsRemaining--;
   waitOne(); 
  }
}, nextRender);
```

atruskie (approved):

PR #300, opened 2020-07-08
Authenticated Images

### LINKED ISSUE #232: Add sanitization to image payloads
Currently images are not sanitized before being displayed on the website. The following steps should happen before displaying an image:

- Check if URL matches API host (i.e. staging.ecosounds)
- Validate image: https://stackoverflow.com/questions/18837735/check-if-image-exists-on-server-using-javascript

This can be automated by creating `@BawImage` and `@BawImages` decorators for models

Allcharles:
# Authenticated Images

Create authenticated image requests and handle defaults for model images.

## Changes

- Created `AuthenticatedImageDirective`
- Created `BawImage` decorator
- Update app `` elements to use new directive
- Added ng-mocks library (https://github.com/ike18t/ng-mocks)
- Rewrote all CardModule component unit tests using spectator

## Issues

- [x] Need to append authentication token to urls
  - Append to urls which contain the `API_ROOT` in their path only
  - This should be handled by the `ImageDirective`
  - Add input to `ImageDirective` which disables adding authentication token to urls
- [x] Need to rename `ImageDirective` and `ImageDirectiveModule` to more descriptive names
- [x] Tests are failing
- [x] Need to write tests for decorator and directive
- [x] Re-write card component tests to jasmine
- [x] Create default image url if none provided for `BawImage` directive
- [x] Edit `assertImage` function to assert that images are using the `ImageDirective`
- [x] Rebase branch

## Closes

Closes #232


src/app/directives/image/image.directive.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    it("should handle multiple url errors", () => {
+      const imageUrls = modelData.imageUrls();
+      spectator = createDefaultDirective(imageUrls);
+
+      const image = getImage();
+      [1, 2, 3].forEach(() => createImgErrorEvent(image));
+      assertImage(image, imageUrls[3].url, "alt");
+    });
```
atruskie: very minor nit: using more descriptive test names would act like documentation.

Instead of `should handle url error` maybe: `given an invalid url, it loads the next available url`

and:
 `given multiple bad urls, it tries to load each until one succeeds`

src/app/directives/image/image.directive.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+      assertImage(image, imageUrls[0].url, "alt");
+    });
+  });
+
+  describe("thumbnail", () => {
+    it("should use thumbnail url", () => {
+      const imageUrls = modelData.imageUrls();
+      spectator = createThumbnailDirective(imageUrls, ImageSizes.MEDIUM);
+      // MEDIUM image size is created at 2nd index by modelData.imageUrls
+      assertImage(getImage(), imageUrls[2].url, "alt");
```
atruskie: Don't hard code this index. Try `imageUrls.filter(x => x.size == imageSizes.MEDIUM).url`

src/app/directives/image/image.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+  SimpleChanges,
+} from "@angular/core";
+import { SecurityService } from "@baw-api/security/security.service";
+import { API_ROOT, ASSET_ROOT } from "@helpers/app-initializer/app-initializer";
+import { ImageSizes, ImageUrl } from "@interfaces/apiInterfaces";
+
+export const image404RelativeSrc = "/assets/images/404.png";
+
+@Directive({
+  // tslint:disable-next-line: directive-selector
```
atruskie, line 19: explain why we think it is ok to disable this warning

src/app/directives/image/image.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+    @Inject(ASSET_ROOT) private assetRoot: string,
+    private securityApi: SecurityService,
+    private imageRef: ElementRef<HTMLImageElement>
+  ) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
+    // On Component Initial Load
+    if (changes.src.isFirstChange()) {
+      // Set attribute so that unit tests can track authenticated images
+      this.imageRef.nativeElement.setAttribute("bawImage", "true");
```
atruskie: Any other solution? This just seems like noise

src/app/directives/image/image.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  constructor(
+    @Inject(API_ROOT) private apiRoot: string,
+    @Inject(ASSET_ROOT) private assetRoot: string,
+    private securityApi: SecurityService,
+    private imageRef: ElementRef<HTMLImageElement>
+  ) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
+    // On Component Initial Load
```
atruskie, line 52: you should reset `srcIndex` if the input changes - it might be a new array, differently sized, or with different valid/invalid urls

src/app/directives/image/image.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+})
+export class AuthenticatedImageDirective implements OnChanges {
+  @Input() src: ImageUrl[];
+  @Input() thumbnail: ImageSizes;
+  @Input() disableAuth: boolean;
+
+  /**
+   * Tracks which src value to display, unless thumbnail is used
+   */
+  private srcIndex = 0;
```
atruskie: another way you could handle tracking bad images would be to:
- tag the image object/url when it's faulty
- or keep a `Set<>` of URLs

Either way might make it easier to track bad urls. See comment about resetting index below

src/app/directives/image/image.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+      this.displayThumbnail = false;
+    }
+
+    this.setImageSrc();
+  }
+
+  /**
+   * Retrieve thumbnail image url or return null
+   */
+  private retrieveThumbnailIfExists(): string | null {
```
atruskie: could probably express this with a `.filter` for succinctness

atruskie (changes requested):
Excellent work. I'd say the specs for the image directive need a little more descriptive documentation - as per the comment I think explaining what the test is testing in the test title would be enough.

Also ( i know we've talked about this before) but why not have the model decorator append the auth token?

Allcharles:
I had the AuthenticatedImageDirective handle appending authToken because it has better access to angulars DI. If I had it part of the decorator, it would require a minor rewrite of the generic decorator builder (not a huge problem), and the model does not always get instantiated with the injector service which I require to append the authToken (bit more of a problem, although I don't think this currently affects any cases in the app).

atruskie (approved):

Issue #51, opened 2019-12-17
Custom formly templates


Allcharles:
Need to create the following custom formly templates for various forms in the application:
- [ ] Typeahead Input (connected to api)
- [x] Checkbox Input
- [x] File Input (inbuilt version only provides path to file)
- [x] Question/Answer Field
 
![image](https://user-images.githubusercontent.com/3955116/71048162-1c2aec80-218a-11ea-91b1-f18bedc7ae41.png)

- [x] Question/Answer/Action Field

![image](https://user-images.githubusercontent.com/3955116/71048181-2b119f00-218a-11ea-86f8-2e6f916804af.png)

- [x] Timezone Input
  - Should return output as an IANA value: https://www.iana.org/time-zones
  - Luxon has access to the IANA dataset, may be possible to access through it: https://moment.github.io/luxon/docs/manual/zones.html#iana-support
- [ ] Recaptcha Field
- [x] Table Selection Field
  - Multi
  - Single

![image](https://user-images.githubusercontent.com/3955116/73633366-46892880-46aa-11ea-9c1c-184643529752.png)

Allcharles:
Bug fix for the following error here: https://github.com/ngx-formly/ngx-formly/issues/1981

`Type 'AbstractControl' is missing the following properties from type 'FormControl': registerOnChange, registerOnDisabledChange, _applyFormState`

Allcharles:
Recaptcha library: https://github.com/Enngage/ngx-captcha
Relevant link to agx-formly: https://github.com/ngx-formly/ngx-formly/issues/1473

Allcharles:
Blocked until the following baw server provides recaptcha seeds.

Related issues:
https://github.com/QutEcoacoustics/baw-server/issues/414
https://github.com/QutEcoacoustics/baw-server/issues/415
https://github.com/QutEcoacoustics/baw-server/issues/416
https://github.com/QutEcoacoustics/baw-server/issues/424

Allcharles:
Extracted recaptcha to issue #356 and dropped the requirement for a typeahead component (currently it has no use case).

PR #342, opened 2020-07-22
IANA timezone

Allcharles:
# IANA Timezone Input

Created an IANA compatible timezone input. When baw-server is updated to allow setting of timezones, this should be compatible.

## Problems

- No unit tests, attempted but gave up for now
  - Will need to verify outputs are correct when baw-server is fixed
  - Issues may be spectator related, might have to write tests using default library
- Changes break site, register, and edit profile forms
  - Will be addressed in another PR which merges into this one

## Issues

Related: #51  
Related: https://github.com/QutEcoacoustics/baw-server/issues/410

## Visual Changes

image

image

image

image


Allcharles:
Looking into the `npm run stats` output, it looks like the repo now contains both `luxon` and `moment`. `moment` is being included with `moment-timezone` so I will need to look into removing the `moment` dependency, or replacing `luxon` with `moment`. Keeping both is a waste of resources.

image

Allcharles:
Using dynamic imports:
image

image

Allcharles:
Using [countries-and-timezones](https://github.com/manuelmhtr/countries-and-timezones):
image

image

## Note:

List of IANA timezones includes deprecated values

Allcharles:
Using [tzdb](https://github.com/vvo/tzdb/):

image

image

Allcharles:
Using tzdb with dynamic imports:

image

image

src/app/component/shared/formly/timezone-input.component.ts
```diff
@@ -1,67 +1,133 @@
...
+
+      <input
+        #instance="ngbTypeahead"
+        type="text"
+        class="form-control"
+        placeholder="Type a city or country name."
+        [editable]="false"
+        [ngbTypeahead]="search"
+        [inputFormatter]="formatter"
+        [resultTemplate]="rt"
```
atruskie: We can choose a better name than `rt`

src/app/component/shared/formly/timezone-input.component.ts
```diff
@@ -1,67 +1,133 @@
...
+          field.templateOptions.label +
+            (field.templateOptions.required ? " *" : "")
+        }}
+      </label>
+
+      <ng-template #rt let-r="result" let-t="term">
+        <ngb-highlight
+          [result]="r.currentTimeFormat"
+          [term]="t"
+        ></ngb-highlight>
```
atruskie: the screenshot in the PR shows the textual labels using the `name` value instead of `currentTimeFormat` - is that an issue with the screenshot or is this code not working as intended?

atruskie (changes requested):

src/app/component/shared/formly/timezone-input.component.ts
Allcharles: Screenshot, i forgot to update it, my bad

atruskie (approved):

Allcharles:
Final breakdown of bundle:

image

PR #354, opened 2020-07-27
Custom Form Input Cleanup

### LINKED ISSUE #341: Lat/Long input selector
Create a custom lat/long input for forms which integrates with google maps. This should allow the user to switch between manually entering the latitude and longitude coordinates, or dragging a map marker around. This will also make it easier for researchers to visually verify the location is correct without checking the sites detail page after.

This will require heavy modification of the `MapComponent`.

Relates to #51 

### Current System

image

### Proposed

temp

Allcharles:
# Custom Form Inputs

Cleanup form behaviour by creating a few custom formly inputs which handle specific form validations.

## Changes

- Modified timezone input
  - Added some visual feedback to help users identify when a timezone is properly set
- Create password input
  - This creates a password, and password confirmation field which validate each other
- Created location input
  - This creates a latitude and longitude field
  - Both fields are validated to ensure they are either both set, or unset. And that the values are valid (-90 -> 90 and -180 -> 180)
  - Displays position of lat/lng on map so users can verify it is in the right location
    - Ended up not making it so user can drag marker, may be revisited later if requested as a feature
- Modified `RenderFieldComponent` to better handle images
- Extracted custom formly inputs into a singular module
- Extracted `IndicatorComponent` into a module
- Added `XOR` type
- Fixed issue in baw interceptor where incoming error messages were not being converted to camelCase

## Problems

- Unit tests are broken (separate PR)

## Issues

Closes #341 
Related to #51 

### Lat/Lng Input

image
image
image

### Password Input

image
image
image
image

### Timezone Input

image
image
image
image

#### Handling array of images

image
image

#### Handling single image

image


Allcharles:
Large jump in file size appears to be because i now have my testing data mixed into the production code. This will be removed once i take the mocks out of the `ShallowSiteService` class. For now, it shouldn't be an issue

src/app/models/AssociationDecorators.ts
```diff
@@ -7,27 +7,31 @@ import { AbstractModel, UnresolvedModel } from "./AbstractModel";
...
  * Creates an association between the updaterId and its user model
+ * TODO Modify to handle ghost users
  */
 export function Updater<M extends AbstractModel & { updaterId?: Id }>() {
   return HasOne<M>(ACCOUNT, "updaterId");
 }
 
 /**
  * Creates an association between the deleterId and its user model
+ * TODO Modify to handle ghost users
```
atruskie, line 34: anonymous users?

src/app/models/Site.ts
```diff
@@ -107,4 +120,34 @@ export class Site extends AbstractModel implements ISite {
...
+    return this.latitude ?? this.customLatitude;
+  }
+
+  /**
+   * Get site longitude
+   */
+  public getLongitude(): number {
+    return this.longitude ?? this.customLongitude;
+  }
+
```
atruskie, line 137: seems like duplicated logic with the constructor?

atruskie (approved):
Looks great.

One note: for the standard image size arrays returned by the API there is really no reason we'd want to show all of the sizes (as per your demo image). I assume you were just testing support for different size images?

src/app/models/AssociationDecorators.ts
Allcharles: deleted/404 ones yes

Allcharles:
>One note: for the standard image size arrays returned by the API there is really no reason we'd want to show all of the sizes (as per your demo image). I assume you were just testing support for different size images?

Removed feature in child branch

src/app/models/Site.ts
Allcharles: Yes and no, the latitude and longitude for a model are only set if the user can actually read the un-obfuscated coordinates. For all other users, they will still use the custom values.

Allcharles: The reason the constructor sets the latitude and longitude to be the same as the custom values is because the API will read the lat/lng values when changing the coordinates of the site. Setting these values for sites where the user does not actually know the correct lat/lng removes the semantic difference provided by the API.

PR #363, opened 2020-08-02
Mock services

Allcharles:
# Mocking Services

Created a better method of mocking services for unit tests. Now unit tests can import `MockAppConfigModule` and `MockBawApiModule` when components use any of the included services. This new implementation also takes advantage of the new `ng-mocks` dependency to remove the need for creating mocks for services ahead of time. This means mocks should never be out of sync with services.

## Problems

- New mock services do not provide empty Subjects by default (means some tests had to be updated to return empty Subjects otherwise the component would crash)

### DIFF
--- a/src/app/services/baw-api/site/sites.service.ts
+++ b/src/app/services/baw-api/site/sites.service.ts
@@ -14,23 +14,21 @@ import {
   IdOr,
   IdParam,
   IdParamOptional,
-  New,
   option,
   StandardApi,
 } from "../api-common";
 import { Filters } from "../baw-api.service";
 import { Resolvers } from "../resolver-common";
 
-type Orphans = "orphans";
-const Orphans: Orphans = "orphans";
-function shallowOption(x?: New | Filter | Empty | Orphans) {
-  return x ? x : Empty;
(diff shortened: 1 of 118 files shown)


Allcharles:
Large reduction in file size is because testing libraries were completely removed from the production build.

src/app/services/baw-api/site/sites.service.ts
```diff
@@ -15,17 +14,23 @@ import {
...
+const Orphans: Orphans = "orphans";
+function shallowOption(x?: New | Filter | Empty | Orphans) {
+  return x ? x : Empty;
+}
+
 const projectId: IdParam<Project> = id;
 const siteId: IdParamOptional<Site> = id;
 const endpoint = stringTemplate`/projects/${projectId}/sites/${siteId}${option}`;
-const endpointShallow = stringTemplate`/sites/${siteId}${option}`;
+const endpointShallow = stringTemplate`/sites/${siteId}${shallowOption}`;
```
atruskie: We should consider the orphan route a different route/endpoint.
It will only have the `filter` option (not `new`)

atruskie (approved):

PR #407, opened 2020-08-21
Config refinement

### LINKED ISSUE #401: Do not set fallback image when fallback image fails
**Describe the bug**

When we use our img directive, and when the url is set to a src that fails, we set a fallback image.

If that fallback image fails to load, then the fail loading code sets the fallback image again. An infinitum. 

**To Reproduce**
1. Set an image to a url that will fail
1. Ensure the fallback image laso fails
1. Go to any page

**Expected behavior**

The fallback image should only be set once.

**Screenshots**

image

 - OS: Windows
 - Browser: Chrome
 - Version: 84

**Additional context**

Happens on beta deploy. The missing fallback image is a separate issue.

### LINKED ISSUE #402: ERROR TypeError: Cannot read property 'home' of undefined
**Describe the bug**

```
main-es2015.4d979fccb747e600b501.js:1 ERROR TypeError: Cannot read property 'home' of undefined
    at e.ngOnInit (main-es2015.4d979fccb747e600b501.js:1)
    at Rn (main-es2015.4d979fccb747e600b501.js:1)
    at Fn (main-es2015.4d979fccb747e600b501.js:1)
    at Dn (main-es2015.4d979fccb747e600b501.js:1)
    at Ur (main-es2015.4d979fccb747e600b501.js:1)
    at main-es2015.4d979fccb747e600b501.js:1
    at Ur (main-es2015.4d979fccb747e600b501.js:1)
    at go (main-es2015.4d979fccb747e600b501.js:1)
    at main-es2015.4d979fccb747e600b501.js:1
    at Ur (main-es2015.4d979fccb747e600b501.js:1)
```

**To Reproduce**
1. Go to 'beta.staging.ecosounds.org'

**Expected behavior**

For home to be defined

**Screenshots**

image

 - OS: Windows
 - Browser: Chrome
 - Version: 84

**Additional context**

Possibly related to the error: `e.url.toLowerCase is not a function` which occurs in the body of `baw-cms`

image

### LINKED ISSUE #403: Malformed URLs in production deploy
**Describe the bug**

Many of the URLs in a production deploy don't seem to be correct.

For example:

```
<img _ngcontent-workbench-client-c75="" alt="Profile Icon" src="undefined/assets/images/404.png">
```
See the `undefined` in that url.

```
https://beta.staging.ecosounds.org/undefined/system/users/images/000/000/009/span1/DSC02329.JPG?1443423283
```

**To Reproduce**
1. Go to 'beta.staging.ecosounds.org'

**Expected behavior**

These URLs should work, and in an error case should report useful messages.

**Screenshots**

 - OS: Windows
 - Browser: Chrome
 - Version: 84

**Additional context**

It looks like the root cause of this is missing properties in the deployed environment.json.  The values:

https://github.com/QutEcoacoustics/workbench-client/blob/b672a8092ae559b21e30c8ef273cabc4d270ae19/src/assets/environment.json#L7-L8

Are not provided in the production/staging environments and the config file needs to be adjusted to handle this.

Further, for the second URL it looks like it's requesting a image from the API server, but the fully formed URL is pointing to the client server.

Allcharles:
# Configuration File Refinements

Refined how the configuration file works to match production expectations.

## Changes

- Removed the following properties from config:
  - `cmsRoot`
  - `assetRoot`
  - `siteDir`
  - `siteRoot`
- Added the following properties to config:
  - `environment.ga.domain`
- Fixed image directive bug where fallback image failing would cause an infinite loop
- Enhanced error message when cms values are improperly set

## Issues

Closes #403
Closes #402 
Closes #401 

## Visual Changes

Example of CMS page failing to load
image


Allcharles:
Noticed a mistake, image.directive should not prepend /assets to local resources

src/app/directives/image/image.directive.ts
```diff
@@ -125,16 +124,19 @@ export class AuthenticatedImageDirective implements OnChanges {
...
 
   /**
    * Prepend the asset root to any local urls
    * @param url Url to potentially format
    */
   private formatIfLocalUrl(url: string): string {
-    if (url.startsWith("/") && !url.startsWith(this.assetRoot)) {
-      return this.assetRoot + url;
+    if (url.startsWith("/") && !url.startsWith("/assets")) {
+      return "/assets" + url;
```
atruskie, line 139: this should be an app-wide constant

atruskie: Actually, I have an issue with this entire function. We shouldn't be formatting a URL. URLs passed to image direction should be valid or not valid.

src/app/services/app-config/app-config.service.ts
```diff
@@ -38,21 +39,33 @@ export class AppConfigService {
...
+  /**
+   * Retrieve cms url path
+   * @param cms CMS page to retrieve
+   */
+  public getCms(cms: keyof CMS): string {
+    console.log(this._config);
+
+    return this.values.cms?.[cms]
+      ? this.values.cms[cms]
+      : "/assets/content/error.html";
```
atruskie, line 69: I won't mandate this for this PR, but this will end up as a HTTP REST service, just like any other API call

atruskie (changes requested):
So the urls for the expectOne in the tests for CMS will change once feature is implemented. Fine for now though.

I'd like to see the removal of asset URL modifying code in the image directive. Assets should be known values for us, in which case we should be able to construct the URLS in a principled fashion.

Allcharles:
Had a private discussion, and the following behaviour should be achieved by this PR:

Remove the following from `environment.json`:

- assetRoot
- cmsRoot

Handle urls the following ways
- local resources (workbench-client):
  - Relative urls: siteDir (config) + assetDir (defined in app) + name
  - Absolute urls: siteRoot (config) + assetDir (defined in app) + name
- api resources:
  - Relative urls: apiRoot + name
  - Absolute urls: name

Allcharles:
I'm going to separate this work into two different PRs: Fixing SecureImageDirective bug, and applying config changes. This PR wont be fixed, however its being left so I can extract code, and read reviews

Allcharles:
Closed in favour of #423 and #422

Issue #413, opened 2020-08-24
Convert Shared Components into Modules


Allcharles:
The current implementation of shared modules is overly complicated, and can make testing fairly difficult (and brittle). All shared components should instead be grouped into modules, with optional testing modules if the component has behaviour which makes it difficult to test in parent components.

As a plus, this will help with #221

PR #415, opened 2020-08-24
Model description

### LINKED ISSUE #379: Model description markup
Currently, when models are displayed, the system uses the models `description` parameter. This should instead be changed to `descriptionHtml` where relevant (such as the details pages). Allowing HTML injection however is a potential point of attack and research should be done into evaluating which is a safe method of accepting it onto the page. Maybe only allow whitelisted html elements if possible?

Allcharles:
# Model Description Markup

Updated model descriptions, and short descriptions, so that the proper markup is applied.

## Changes

- Update home page to display short description with truncation
- Updated project list page to display short description with truncation
- Updated project details page to enable html formatting
- Updated site details page to enable html formatting
- Added new dependency: https://github.com/DiZhou92/ngx-line-truncation

## Problems

- Unit tests could be made a bit tidier (ref #413)
- Would like to implement short descriptions when available https://github.com/QutEcoacoustics/baw-server/issues/486
- Site details styling not displaying because of https://github.com/QutEcoacoustics/baw-server/issues/487

## Issues

Closes #379 

### Home Page

image

### Project List

image

### Project Details

image

### Site Details

image


src/app/components/shared/cms/cms.component.ts
```diff
@@ -48,6 +48,8 @@ export class CmsComponent extends WithUnsubscribe() implements OnInit {
       .pipe(takeUntil(this.unsubscribe))
       .subscribe(
         (data) => {
+          // TODO Validate if this is needed?
```
atruskie, line 53: I have confirmed the server will sanitize rendered markdown blobs

atruskie (approved):
So the API will now return properly formatted and truncated taglines for markdown blobs. See https://github.com/QutEcoacoustics/baw-server/pull/488

The `*_html_tagline` props from https://github.com/QutEcoacoustics/baw-server/pull/488 should obviate the need for the `stripHtml` function.

src/app/components/shared/cms/cms.component.ts
Allcharles: Currently, this is needed so that some non-approved html elements can be applied to CMS pages. Namely, the home component uses the `<style>` element to apply styling (although that is no longer needed because of #421).

Allcharles:
Updated branch to handle https://github.com/QutEcoacoustics/baw-server/pull/488

src/app/components/shared/cms/cms.component.ts
Allcharles: #421 had an update which now requires access to the `<style>` element again. I wont be able to remove the sanitization applied to the CMS without breaking the home page.

atruskie (approved):
Only note is that we call those "user ids and dates" _stamp_. So it might be better to name them `userStamps` but meh... doesn't need to change

PR #421, opened 2020-08-27
Landing image

Allcharles:
# Landing Page Hero Image

Updated the landing page hero image to use images provided by the QUT ecoacoustics division staff. Special credits to the following individuals who provided images to use:

- Marina D. A. Scarpelli
- David Tucker
- Jinglan Zhang

## Changes

- Removed old landing page image
- Added QUT landing page images
- Added credits to landing page image (will need to be updated to show the proper people once on CMS)
- Added README to track proper ownership
- Currently, image changed of window size. This is only to display that the image can change in the future, this is not the final implementation

## Problems

- Cannot change who is credited with the photo because the CMS page is pure HTML. This will be fixed once moved to the CMS service.
- New design is not phone compatible

### Before

image

### Draft 2

image
image
image
image

### Final Product

image


Allcharles:
Manually tested E2E tests, they are working

atruskie (approved):
Some design notes:

It'd be good if the background:
- remain fixed to window
- the sections below were transparent
- the site title was white, larger, and overlaid on the image

The credits text should be less prominent (and not look like a button). Text should also be: "Photo credit: xxx".

Allcharles:
Manually tested E2E tests, they are working

PR #422, opened 2020-08-27
Image directive 404 bug

### LINKED ISSUE #401: Do not set fallback image when fallback image fails

Allcharles:
# Image Directive Infinite Loop Bug

`AuthenticatedImageDirective` had a bug where, if the 404 image failed to load, it would infinitely attempt to load the 404 image. This has been fixed so that that it will give up, and leave the image `src` blank. Also took the opportunity to remove the reference to the assetsRoot config value as it will be deprecated.

## Issues

Closes #401


Allcharles:
Manually ran e2e tests, they are passing

atruskie (approved):

PR #423, opened 2020-08-27
Config Refinements V2

### LINKED ISSUE #402: ERROR TypeError: Cannot read property 'home' of undefined

### LINKED ISSUE #403: Malformed URLs in production deploy

Allcharles:
# Config Refinements

Refined how the configuration file works to match production expectations. This is a re-implementation the previous attempt: https://github.com/QutEcoacoustics/workbench-client/pull/407.

## Changes

- Implemented better error checking/handling for CMS components
- Removed `cmsRoot` and `assetRoot` from app config
- Added `domain` to `ga` config value

## Problems

- Some potential pain points to take this refinement further. Specifically referencing this requested change: https://github.com/QutEcoacoustics/workbench-client/pull/407#issuecomment-678833721
  - The change, as it currently stands, is not possible. This is a limitation in how baw models are created. To achieve the desired result, I would have to use the Inject service to retrieve the `API_ROOT`, `SITE_DIR`, and `SITE_ROOT`. However, the model is not guaranteed to be created with the Inject service attached (ie. In the case of forms which effectively lose this value). If I try to implement this behaviour somewhere else, such as the `AuthenticatedImageDirective`, it will be impossible to determine if the resource is an local asset, or api asset. The best middle ground I can think of, is to either use the current implementation (which works fine and will only mean changing some of the default api outputs, however even those arent required), or to have the model prepend the image url with an identifier such as `local_asset/user_profile.jpg` and `api_asset/user_profile.jpg`, which is then interpreted and replaced by `AuthenticatedImageDirective`.

## Issues

Closes #403 
Closes #402 

### New CMS error handling:
image

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -12,6 +12,7 @@ import {
   Configuration,
 } from "@helpers/app-initializer/app-initializer";
 import { ngExpressEngine } from "@nguniversal/express-engine";
+import { assetRoot } from "@services/app-config/app-config.service";
 import express from "express";
 import { existsSync, readFileSync } from "fs";
 import { join } from "path";
@@ -52,7 +53,7 @@ export function app(path: string): express.Express {
   server.set("views", distFolder);
 
   // special case rendering our settings file - we already have it loaded
-  server.get("/assets/environment.json", (request, response) => {
+  server.get(`${assetRoot}/environment.json`, (request, response) => {
     response.type("application/json");
     response.send(rawConfig);
   });
--- a/src/app/components/about/pages/credits/credits.component.spec.ts
+++ b/src/app/components/about/pages/credits/credits.component.spec.ts
@@ -5,6 +5,7 @@ import {
 import { async, ComponentFixture, TestBed } from "@angular/core/testing";
 import { RouterTestingModule } from "@angular/router/testing";
 import { AppConfigService } from "@services/app-config/app-config.service";
+import { cmsRoot } from "@services/app-config/app-config.service.spec";
 import { MockAppConfigModule } from "@services/app-config/app-configMock.module";
 import { SharedModule } from "@shared/shared.module";
 import { CreditsComponent } from "./credits.component";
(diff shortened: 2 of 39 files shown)


Allcharles:
Manually ran e2e tests, they are passing.

atruskie (approved):
Noted on the issues with getting config values into models. The answer seems to be making the models `@Injectable` as per https://stackoverflow.com/questions/48167637/inject-a-service-in-a-model-class-angular.

However, we'll implement this when we need to.

Allcharles:
That solution may run into the issue with the angular compiler removing decorators in prod mode. However, it may be possible to bypass that using the solution you presented, by having the model class in a separate file to the `@Injectable` builder.

PR #466, opened 2020-09-16
Implemented NxCloud

### LINKED ISSUE #463: Investigate speeding up compile time using Nx Cloud
Nx Cloud offers to speed up compile time by up to 10 times. This is performed by maintaining compiled versions of the local files, with recompilation only happening with modified files. Currently the angular cli only performs this for the node_modules libraries. This library will NOT improve the compile time of the CI, only the dev computer on repeat builds.

Relevant Link: https://nx.app/angular-cli-faster

Allcharles:
# Implemented NxCloud

Implemented NxCloud to speedup compile times for build, test, e2e commands. This significantly decreases the time it takes to compile repeat builds, it should also speed up the CI as it will use the same compilation steps that have already been performed during development.

## Changes

- Using NxCloud to cache compiled files to speed up dev and CI processes
- Implemented node_module caching to speed up CI processes

## Problems

- Windows cache is slower than `npm install`, looks to be related to the fact that node_modules for windows are 800MB instead of the 250MB for linux/mac. @atruskie Should we removed caching from Windows, or just hope they eventually improve windows performance later down the line?
  - Potentially related issues: https://github.com/actions/cache/issues/336 https://github.com/actions/cache/issues/267#issuecomment-662600812 

## Issues

Fixes #463


.github/workflows/workflow.yml
```diff
@@ -18,11 +18,25 @@ jobs:
...
         with:
           node-version: 12.x
 
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
```
atruskie: https://docs.npmjs.com/configuring-npm/folders.html#cache, `%AppData%/npm-cache on Windows

Allcharles:
Alright looks like the reason windows is failing to cache node_modules is potentially because of https://github.com/actions/cache/issues/144. Hopefully by tomorrow the cache will be destroyed and the bug will self resolve.

Edit: Cache is now resolving

atruskie (approved):

PR #481, opened 2020-09-22
QSP Filtering

### LINKED ISSUE #477: Populate google maps with all sites
Currently, when visiting the project details page, only the sites in the current filter page are displayed in the google maps. This should instead show all sites by default, however this functionality is await on the following issue: https://github.com/QutEcoacoustics/baw-server/issues/495. This should be a filter request on page load, which uses projection to limit the data transfer to `name`, `customLatitude`, and `customLongitude`. If the current page of sites returns first, initially populate the map with those values.

This feature update should also add radio button inputs below the maps component to enable switching between display all sites and displaying all sites in the current filter. We may also add the option to change the map drawing type (hybrid/satellite/etc).

Allcharles:
# Filtering and Paging with QSP support

Update to base PR converting infinite scrolling to paged setup. This better allowed for tracking filters and paging without complicated logic to retrieve and store scroll heights when saving queries to url history.

## Changes

- Changed from infinite scrolling to pagination
- Store filter in query params
- Store page number in query params
- Push previous filter/page requests to browser history to enable users to return to previous searches
- Extracted duplicate logic from project list and details pages into single abstract class

## Problems

- Not as fancy a solution as infinite scrolling (however this isn't a main feature and a lot of users may never use it)
- Previous PR review talks about creating a `pagedFilter` on the services which would abstract some of the filtering/paging logic, this may be something to implement as filtering/paging becomes more prevalent in the app

## Issues

Deals With #477 

### List Page
image

#### Single Page

image

### Multi Page

image

## Final Checklist


src/app/components/projects/pages/details/details.component.scss
```diff
@@ -3,7 +3,7 @@
 
 #sites {
   display: grid;
-  grid-template-columns: 1fr 1fr;
+  grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
```
atruskie, line 6: 💯

src/app/helpers/paginationTemplate/paginationTemplate.ts
```diff
@@ -0,0 +1,198 @@
...
+import { NgbPaginationConfig } from "@ng-bootstrap/ng-bootstrap";
+import { noop, Observable, Subject } from "rxjs";
+import { filter, map, mergeMap, takeUntil } from "rxjs/operators";
+
+const queryKey = "query";
+const pageKey = "page";
+
+@Directive()
+// tslint:disable-next-line: directive-class-suffix
+export abstract class PaginationTemplate<I, M extends AbstractModel>
```
atruskie: What type do we expect `I` to be? Constraint needed?

src/app/helpers/paginationTemplate/paginationTemplate.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Read the previous filter query and page number from the url,
+   * and update the page to match
+   */
+  private updateFromUrl() {
+    const params = this.route.snapshot.queryParams;
+    this.filter = params[queryKey] ?? "";
+    this._page = parseInt(params[pageKey], 10) || 1;
```
atruskie, line 130: Use a central int parsing function that removes the base paramter as an option

src/app/helpers/paginationTemplate/paginationTemplate.ts
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Handle filter events
+   */
+  public onFilter(input: string, page: number = 1) {
+    this._page = page;
+    this.filter = input;
+    this.loading = true;
+    this.apiRequest$.next();
+    this.updateQueryParams(this._page, input);
+  }
```
atruskie: Some repeated logic here.

Also it seems like `apiResponse$` could `tap` before `getModels` to  set `loading` to true. That way the event logic is inside the rest of the chain.

Similarily, the query string update should be part of the chain as well

src/app/helpers/paginationTemplate/paginationTemplate.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.config.maxSize = 3;
+    this.config.pageSize = defaultApiPageSize;
+    this.config.rotate = true;
+    this.displayPagination = false;
+
+    this._page = 1;
+
+    this.apiRequest$
+      .pipe(
+        mergeMap(() => this.getModels()),
```
atruskie: Is `MergeMap` the right choice here?  Maybe `switchMap`?

https://www.learnrxjs.io/learn-rxjs/operators/transformation/switchmap

> Why use switchMap?
The main difference between switchMap and other flattening operators is the cancelling effect. On each emission the previous inner observable (the result of the function you supplied) is cancelled and the new observable is subscribed. You can remember this by the phrase switch to a new observable.
This works perfectly for scenarios like typeaheads where you are no longer concerned with the response of the previous request when a new input arrives. This also is a safe option in situations where a long lived inner observable could cause memory leaks, for instance if you used mergeMap with an interval and forgot to properly dispose of inner subscriptions. Remember, switchMap maintains only one inner subscription at a time, this can be seen clearly in the first example.

src/app/helpers/paginationTemplate/paginationTemplate.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Handle filter events
+   */
+  public onFilter(input: string, page: number = 1) {
+    this._page = page;
+    this.filter = input;
+    this.loading = true;
+    this.apiRequest$.next();
```
atruskie: everywhere, where next is called, why not pass page and filter into the `next` function to make it immutable?

src/app/helpers/paginationTemplate/paginationTemplate.ts
```diff
@@ -0,0 +1,198 @@
...
+          ...this.defaultFilter,
+          paging: { page: this._page },
+          filter: this.filter ? (innerFilter as any) : undefined,
+        },
+        ...this.apiParams()
+      )
+      .pipe(
+        map((models: M[]) => {
+          this.loading = false;
+          this.collectionSize = models?.[0]?.getMetadata()?.paging?.total || 0;
```
atruskie: probably better to use `null` to indicate an error value?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -258,22 +266,29 @@ export abstract class BawApiService<T extends AbstractModel> {
...
   items?: number;
+  /** Total number of items for filter request */
   total?: number;
+  /** Maximum page number */
   maxPage?: number;
+  /** @deprecated */
   current?: string;
+  /** @deprecated */
   previous?: string;
+  /** @deprecated */
```
atruskie: why deprecated?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -360,34 +375,49 @@ export type InnerFilter<T = {}> = Combinations<T> &
  * https://github.com/QutEcoacoustics/baw-server/wiki/API:-Filtering
  */
 export interface Filters<T = {}, K extends keyof T = keyof T> {
+  /** Filter settings */
   filter?: InnerFilter<T>;
+  /** Include or exclude keys from response */
   projection?: {
+    /** Whitelist keys */
```
```suggestion
    /** include keys */
```

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -360,34 +375,49 @@ export type InnerFilter<T = {}> = Combinations<T> &
...
     include?: K[];
+    /** Blacklist keys */
```
```suggestion
    /** exclude keys */
```

atruskie (changes requested):
Looks pretty neat.

I'd say more work is needed dealing with the observables - it looks like you're encapsualting and unwrapping them which AFAIK defeats the point. E.g. the `apiUpdate` function is a callback, which could have been invoked after a promise, whereas for an observable one is meant to subscribe to them.

src/app/helpers/paginationTemplate/paginationTemplate.ts
atruskie: I think I just grokked this. You need merge map to flatten the multiple observables returned by the filter request?

Which is what bugged me about getModels in the first place...

src/app/helpers/paginationTemplate/paginationTemplate.ts
Allcharles: I is the interface for the Abstract Model `M`, there is no constraint i can apply that i know of

src/app/helpers/paginationTemplate/paginationTemplate.ts
Allcharles: `parseInt` defaults to base 10, linter was just complaining so i added the 10. Wrapping this function with another one seems like overkill, and is something another programmer new to the repo would be unlikely to look for (leading it to partial use throughout the project).

src/app/helpers/paginationTemplate/paginationTemplate.ts
atruskie: Interface for the child? As in the concrete model?

Maybe just add a comment

src/app/services/baw-api/baw-api.service.ts
Allcharles: AFAIK there is no use for these values in the new workbench-client. They don't provide much useful information that the system can't already calculate, and realistically I can't think of a single scenario these would be the best option to use. I set to deprecated to hint to future developers that there is likely a better solution for tracking what the current, previous and next filters are.

src/app/helpers/paginationTemplate/paginationTemplate.ts
atruskie: > If radix is undefined, 0, or unspecified, JavaScript assumes the following:

>If the input string begins with "0x" or "0X" (a zero, followed by lowercase or uppercase X), radix is assumed to be 16 and the rest of the string is parsed as a hexidecimal number.
If the input string begins with "0" (a zero), radix is assumed to be 8 (octal) or 10 (decimal). Exactly which radix is chosen is implementation-dependent. ECMAScript 5 clarifies that 10 (decimal) should be used, but not all browsers support this yet. For this reason, always specify a radix when using parseInt.
If the input string begins with any other value, the radix is 10 (decimal).

parseInt is well known for its quirks

src/app/services/baw-api/baw-api.service.ts
atruskie: See HATEOAS

Ideally our API returns URLs to most of its own resources. It doesn't do that for most... though that's a separate issue.

The client is not the only client for the API. Other clients may not have enough information to form their own links. So this feature certainly isn't deprecated.

Maybe a comment explaining they're just not used. Or even, remove them from the type system altogether

src/app/helpers/paginationTemplate/paginationTemplate.ts
Allcharles: This is not an error. It just means there are no models in the filter request, and i need to show that on the page

src/app/services/baw-api/baw-api.service.ts
Allcharles: I think I'll just remove them from the type system

src/app/services/baw-api/baw-api.service.ts
Allcharles: Is this inclusion? Its a list of all the keys which will be returned, not a list of additional keys to include?

src/app/helpers/paginationTemplate/paginationTemplate.ts
Allcharles: While that explains the `parseInt` linter rule, it doesn't explain why I should implement a wrapper for it to convert values to base 10?

atruskie: The lint rule depends on the linter working.

Personally, I prefer to remove any possibility of the function using alternate behaviour.

Though I concede with the linter working it is sufficient as is.

src/app/services/baw-api/baw-api.service.ts
atruskie: it is inclusion.

By default, the filter has an implicit 'include' everything. If you override it, it only returns the explicit list you asked for.

src/app/helpers/paginationTemplate/paginationTemplate.ts
Allcharles: Ended up removing it, the typings are going to be a little off but it shouldn't be an issue

atruskie (approved):

Issue #49, opened 2019-12-12
Content Management System (CMS)


Allcharles:
Content management system should enable the website to retrieve `.html` files from the BAW API.

Feature set:

- Sanitise HTML to reduce risk of XSS (https://angular.io/api/platform-browser/DomSanitizer)
- Convert CMS page internal routes to use the angular routing system (https://github.com/QutEcoacoustics/workbench-client/issues/49#issuecomment-565868825)
- Insert data into CMS page (https://github.com/QutEcoacoustics/workbench-client/issues/49#issuecomment-572854175)
- Allow CMS to provide custom CSS styling?

atruskie:
So we'll be using a pre-built CMS. In particular a [headless](https://headlesscms.org/) one.

We're not going to build and part of the CMS or include any module in the angular app.

Whichever service we use, and including the ones we have now, we'll always return HTML to the client

Allcharles:
Another big issue for the CMS service is how we are going to handle internal routes from the HTML blobs. We can convert any '/' routes to use the router module however this leaves the opportunity for the CMS files to become out of sync whenever a route is changed.

Allcharles:
Found another issue for the CMS service; how can we pass data to the CMS page? An example of this is the harvest page which changes its instructions to guide the user on how to harvest audio for a specific project and site. We could make this part of the angular app, however 99% of the page is just text which should be on the CMS.

Allcharles:
Potential solution to adding links (both internal and external) to CMS pages: https://medium.com/@adrianfaciu/using-the-angular-router-to-navigate-to-external-links-15cc585b7b88

Allcharles:
Another issue attached to this is determining how to best retrieve CMS files from the server. Currently, the cms file paths are stored in a config which is sent to the workbench-client. Inside the workbench-client, a manually written (and updated) typing for the CMS files is maintained as a way of ensuring values exist. This is not a long term solution if the number of CMS files dramatically increases.

atruskie:
> Found another issue for the CMS service; how can we pass data to the CMS page? An example of this is the harvest page which changes its instructions to guide the user on how to harvest audio for a specific project and site. We could make this part of the angular app, however 99% of the page is just text which should be on the CMS.

We'll use a custom helper, like https://github.com/QutEcoacoustics/baw-server/blob/master/app/helpers/cms_helpers.rb.
Those functions are exposed to CMS pages (safely). From there we can pull data from route info or QSP.



[...]

PR #541, opened 2020-10-19
CMS API

Allcharles:
# CMS API

Updated cms pages to utilise the new baw server api for cms pages.

## Changes

- Created cms service
- Updated cms component to use cms service
- Removed cms urls from environment
- Updated components which utilise cms component

## Issues

Relates to #49


src/app/components/home/home.component.scss
```diff
@@ -0,0 +1,45 @@
...
+    line-height: 1.1;
+  }
+
+  h2 {
+    font-weight: 400;
+  }
+}
+
+#hero-image {
+  background-image: url("/assets/images/landing/IMG_1318.JPG");
```
atruskie, line 43: this would be a candidate image that could be moved to the CMS. Not sure how it would actually work.

CMS has support for uploading arbitrary files. If we could assign them to a category, then use API to fetch a random one from category...

src/app/components/shared/cms/cms.component.ts
```diff
@@ -1,68 +1,44 @@
...
-        },
-        (err: ApiErrorDetails) => {
-          this.error = err;
-          this.loading = false;
-          this.ref.detectChanges();
-        }
-      );
+    this.cms.get(this.page).subscribe(
+      (blob) => {
+        this.blob = blob;
```
atruskie: I don't think sanitization is actually needed, and it seems you've come that conclusion as well. Can you leave a coment explaining why we think this ok? (and why we are allowed to bypass angular's sanitizer service?)

src/app/components/sites/pages/harvest/harvest.component.html
```diff
@@ -1,3 +1,90 @@
-<baw-wip>
-  <baw-cms [page]="page"></baw-cms>
-</baw-wip>
+<ng-container *ngIf="site">
```
atruskie, line 1: I think you're aware, but thte contents of this page will change. For now it is fine as is.

src/app/components/home/home.component.ts
```diff
@@ -1,84 +1,35 @@
...
-            Photo Credit: QUT Ecoacoustic Staff
-          </small>
-        </div>
-      </div>
-    </div>
-  `,
+  styleUrls: ["./home.component.scss"],
+  templateUrl: "./home.component.html",
   // tslint:disable-next-line: use-component-view-encapsulation
   encapsulation: ViewEncapsulation.None,
```
Allcharles: This shouldn't be needed anymore

src/app/services/baw-api/cms/cms.service.ts
```diff
@@ -0,0 +1,37 @@
+import { HttpClient } from "@angular/common/http";
+import { Inject, Injectable } from "@angular/core";
+import { DomSanitizer, SafeHtml } from "@angular/platform-browser";
+import { API_ROOT } from "@helpers/app-initializer/app-initializer";
+import { stringTemplate } from "@helpers/stringTemplate/stringTemplate";
+import { Observable } from "rxjs";
+import { map } from "rxjs/operators";
+
+export enum CMS {
+  HOME = "",
```
atruskie, line 17: Add comment along the lines of:

> The workbench server has a set of well-known pages that always exist for important pages. 
> Since we always need to fetch these core pages we've defined thei slugs in-app to save time.
> These slugs changing will be considered a breaking change in the server.
>
> More dyanmic pages may exist in the future, in which case the CMS API will be used to disocver what pages are available.

src/app/services/baw-api/cms/cms.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    private sanitizer: DomSanitizer,
+    @Inject(API_ROOT) private apiRoot: string
+  ) {}
+
+  public get(cms: CMS): Observable<SafeHtml> {
+    return this.http
+      .get(this.apiRoot + endpoint(cms), { responseType: "text" })
+      .pipe(
+        map((response: string) =>
+          this.sanitizer.bypassSecurityTrustHtml(response)
```
atruskie: Ah - this answers my question from above

atruskie (approved):
The CMS page components look very neat 👍

We probably only need one page for the harvest pages - not one for each of site/point.

Great work

src/app/components/home/home.component.ts
atruskie: I can't even remember what it did?

Allcharles: It was applying styles to the cms page. It was a dodgy solution that's no longer needed

src/app/components/shared/cms/cms.component.ts
Allcharles: I also need to double check if this sanitisation is actually working the way we want. I should probably add some unit tests to validate it wont filter out styles and scripts

src/app/components/shared/cms/cms.component.ts
```diff
@@ -30,7 +39,12 @@ export class CmsComponent extends WithUnsubscribe() implements OnInit {
 
     this.cms.get(this.page).subscribe(
       (blob) => {
-        this.blob = blob;
+        // Using html fragments instead of angular sanitization sanitization
+        // uses innerHTML to insert html blobs, and this will not run <script>
+        // tags. This is a safety feature of innerHTML and cannot be bypassed.
```
atruskie:
```suggestion
        // Using html fragments instead of innerHTML. 
        // In the HTML5 spec, script tags that are inserted via InnerHTML will not be executed.
        // Using a document fragment allows us to insert any tag.
        // NOTE: Angular's Sanitization is ignored since we are bypassing Angular's normal rendering system.
        // NOTE: It might be useful to consider using ShadowDom to isolate these CMS HTML fragments from 
        //       the rest of the site. This would prevent, for example, a careless CSS global style in the CMS fragment
        //       from affecting the rest of the angular site.
```

atruskie (approved):

Issue #457, opened 2020-09-10
Cleanup PagedTableTemplate


Allcharles:
A number of changes could be made which would improve the quality of the admin pages layout and code. Namely the following should be implemented:

- Re-implement PagedTableTemplate to reduce the amount of noise
  - Combine columns, sortKeys, and rowsCallback into a single object
  - Solution should hopefully remove the need for the TableRow interface as everything is stored in a single array
  - The end result should look similar to the following:

list.component.html
```html
<ng-container *ngIf="rows && !error">
  <h1>List</h1>

  <ngx-datatable
    #table
    bawDatatableDefaults
    [columnMode]="ColumnMode.force"
    [columns]="tableColumns"
    [rows]="tableRows"
    [count]="totalModels"
    [offset]="pageNumber"
    (page)="setPage($event)"
    (sort)="onSort($event)"
  >
    <ngx-datatable-column
      *ngFor="let col of columns"
      [name]="col.name"
      [width]="col.width"
      [maxWidth]="col.width"
      [sortable]="col.sortable">
    </ngx-datatable-column>

    <ngx-datatable-column
      name="Model"
      width="140"
      maxWidth="140"
      [sortable]="false"
    >
      <ng-template let-column="column" ngx-datatable-header-template>
      </ng-template>
      <ng-template let-value="value" ngx-datatable-cell-template>
        <a
          class="btn btn-sm btn-default mr-2"
          [routerLink]="value.adminViewUrl"
        >
          Details
        </a>
        <a class="btn btn-sm btn-default" [routerLink]="value.viewUrl">
          View
        </a>
      </ng-template>
    </ngx-datatable-column>
  </ngx-datatable>
</ng-container>

<baw-error-handler *ngIf="error" [error]="error"></baw-error-handler>
```

list.component.ts
```typescript
class ListComponent extends PagedTableTemplate<TableRow, AnalysisJob> implements OnInit {
  public columns = [
    {
      key: "id",
      name: "Id",
      width: 50,
      sortable: true,
      transform: (model: AnalysisJob) => model.id
    },
    {
      key: "name",
      name: "Name",
      sortable: true,
      transform: (model: AnalysisJob) => model.name
    },
  ];

  constructor(api: AnalysisJobsService) {
    super(api);
  }
}
```

- Re-implement filter/paging to match the current implementation of the `paginationTemplate`. Currently the filter and page number are not stored and will be forgotten whenever a user leaves the page

[...]

PR #543, opened 2020-10-21
Analysis jobs

### LINKED ISSUE #128: Admin Analysis Jobs
Create admin analysis jobs page (`/admin/analysis_jobs`)

Allcharles:
# Analysis jobs

Created the analysis job pages for the admin dashboard

## Changes

- Created Admin Analysis Jobs List Page
- Created Admin Analysis Job Details Page

## Problems

- Attempted to convert the detail components to spectator, however I couldn't figure out why the tests were failing

## Issues

Fixes #128

### Analysis Jobs List

image

### Analysis Job Details

image


src/app/components/admin/analysis-jobs/analysis-job.schema.json
```diff
@@ -0,0 +1,88 @@
+{
+  "fields": [
+    {
+      "key": "id",
+      "hide": true,
+      "templateOptions": {
+        "label": "Id"
+      }
+    },
```
atruskie: this schema will likely be reused for user-facing pages

src/app/components/admin/analysis-jobs/list/list.component.ts
```diff
@@ -0,0 +1,84 @@
...
+        name: analysisJob.name,
+        script: analysisJob.scriptId,
+        creator: analysisJob.creatorId,
+        started: analysisJob.startedAt.toRelative(),
+        status: analysisJob.overallStatus,
+        statusUpdated: analysisJob.overallStatusModifiedAt.toRelative(),
+        model: analysisJob,
+      }))
+    );
+  }
```
atruskie, line 55: This paged table template seems very noisy.

- mapping of models is redundant (mostly)
- column names are verbose
- informal view models, and duplicated objects

It seems to me we should pass:

- models
- table options
    - columns? : ColumnOptions[] // optional: shows some columns be default, like `id` and `name`
        - name: string
        -  { name: string, displayName?: string, transform?: (x: T) => Y }

src/app/components/admin/analysis-jobs/list/list.component.ts
```diff
@@ -0,0 +1,84 @@
...
+interface TableRow {
+  id: Id;
+  name: Param;
+  script: Id;
+  creator: Id;
+  started: string;
+  status: string;
+  statusUpdated: string;
+  model: AnalysisJob;
+}
```
atruskie, line 74: this string of props now repeated three times... `columns`, view model creation, and now typing

src/app/components/admin/analysis-jobs/list/list.component.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Path to view analysis job details
+   * @param model Analysis Job
+   */
+  public detailsPath(model: AnalysisJob) {
+    return adminAnalysisJobMenuItem.route.format({
+      analysisJobId: model.id,
+    });
+  }
```
atruskie: define on model, remove this function

src/app/components/admin/audio-recordings/audio-recording.schema.json
```diff
@@ -2,170 +2,149 @@
   "fields": [
     {
       "key": "id",
-      "type": "input",
+      "hide": true,
```
atruskie, line 5: maybe what you mean to express here is `readOnly`?

src/app/components/shared/detail-view/render-field/render-field.component.ts
```diff
@@ -89,8 +90,11 @@ export class RenderFieldComponent extends WithUnsubscribe()
     super();
   }
 
-  public ngOnInit(): void {
-    this.ngOnChanges();
+  public ngOnInit() {
+    // This is required because unit tests do not run ngOnChanges
+    if (!this.model && !this.display && !this.children) {
+      this.ngOnChanges();
+    }
```
atruskie: link to angular bug?

Otherwise it seems like there should be a shim in the *test helpers* that patches this... not the other way around

src/app/models/AnalysisJob.ts
```diff
@@ -96,7 +97,7 @@ export class AnalysisJob extends AbstractModel implements IAnalysisJob {
   }
 
   public get viewUrl(): string {
-    throw new Error("AnalysisJob viewUrl not implemented.");
+    return adminAnalysisJobMenuItem.route.format({ analysisJobId: this.id });
```
atruskie: incorrect. 

`viewUrl` should point to `/analysis_jobs/123` - the view normal users see
`adminViewUrl` could point to `/admin/analysis_jobs/123`

atruskie (changes requested):
- the columns seem wider than necessary on the details page

src/app/components/admin/analysis-jobs/list/list.component.ts
Allcharles: `columns` object is annoying defined that way by the datatable library

src/app/components/admin/analysis-jobs/list/list.component.ts
Allcharles: This is an admin only link. The model will redirect to the models actual page. Ie. if there is a page for viewing analysis jobs that regular users can access, the model will redirect to that page. I could define a `adminViewUrl` path on the model if you want?

src/app/components/admin/analysis-jobs/analysis-job.schema.json
Allcharles: Alright, i've enabled editing for the name, annotationName, and customSettings. Not sure if the user should be able to change any of the other parameters

src/app/components/admin/analysis-jobs/list/list.component.ts
Allcharles: I could rewrite this to clean it up and reduce the amount of code, however all of the admin pages are done, and changing it would just be spending time cleaning up pages that are:

1. working
2. not important (for the most part, some even appear unused)
3. unlikely to change much in the future
4. changes would require re-implementing/testing all of the admin list pages

I would suggest instead creating an issue with my ideas on how to re-implement this code

src/app/components/admin/analysis-jobs/list/list.component.ts
Allcharles: This is used to type check the entire class. Any changes to this will force the others to be updated at the same time so they will never de-sync in theory.

src/app/components/admin/audio-recordings/audio-recording.schema.json
Allcharles: I rarely want to use readOnly, if users can't edit a field it shouldn't be shown in the form (unless you can think of an example where the additional information would help users?). The hidden fields are just for the admin detail pages where i want to show all of the model parameters

src/app/models/AnalysisJob.ts
Allcharles: Currently im using that as a placeholder so that components wouldn't crash as the analysis jobs route doesnt exist. I've added a TODO to highlight the issue

src/app/components/shared/detail-view/render-field/render-field.component.ts
Allcharles: Its not a bug, but a "feature" of unit tests where assigning values to a component does not trigger the ngOnChanges event detecting even when calling `detectChanges()`. You have to create a host component which wraps the test component and passes the variables through. Ive implemented that now so that this is no longer needed

Allcharles:
Adjusted the table column sizes to try fit closer to the maximum size of cells

src/app/components/admin/analysis-jobs/list/list.component.ts
Allcharles: Added to #457

src/app/components/admin/analysis-jobs/list/list.component.ts
Allcharles: Added to #457

src/app/components/admin/analysis-jobs/analysis-job.schema.json
atruskie: you can refer to params methods in controller to check

src/app/components/admin/audio-recordings/audio-recording.schema.json
atruskie: well you probably still want readOnly.

as for admin pages, why would I not want to see all information?

atruskie (approved):

src/app/components/admin/audio-recordings/audio-recording.schema.json
Allcharles: These are not hidden from the admin view, only the form. Most of this schema is defining how the form will be displayed and has no affect on the DetailsViewComponent

PR #555, opened 2020-10-26
Paged table template revamp

### LINKED ISSUE #457: Cleanup PagedTableTemplate

Allcharles:
# Redesign of pagedTableTemplate

Re-implementing `pagedTableTemplate` to reduce the amount of boilerplate and re-declarations of types

## Changes

- Created `pagedTableTemplateV2` (this will later be renamed to `pagedTableTemplate`)
- Utilised new system in `AdminUserListComponent`

## Problems

- Need to rewrite unit tests
- Need to expand solution to encompass all dependent classes
- Need to update template to handle filter/sorting

## Issues

Fixes #457 

## Final Checklist

### DIFF
--- a/src/app/helpers/tableTemplate/pagedTableTemplate.ts
+++ b/src/app/helpers/tableTemplate/pagedTableTemplate.ts
@@ -13,10 +13,173 @@ import {
   SortType,
   TableColumn,
 } from "@swimlane/ngx-datatable";
-import { Subject } from "rxjs";
+import { Observable, Subject } from "rxjs";
 import { debounceTime, distinctUntilChanged, takeUntil } from "rxjs/operators";
 import { defaultDebounceTime } from "src/app/app.helper";
 import { PageComponent } from "../page/pageComponent";
+
+export interface ColumnData<T, M extends AbstractModel> {
+  key: string;
+  name: string;
+  transform: (model: M) => any;
+  width?: number;
+  sortKey?: keyof T;
+  isFilterKey?: boolean;
+  customTemplate?: boolean;
+}
+
+@Directive()
+// tslint:disable-next-line: directive-class-suffix
+export abstract class PagedTableTemplateV2<T, M extends AbstractModel>
+  extends PageComponent
+  implements OnInit {
+  @ViewChild(DatatableComponent) public table: DatatableComponent;
+
+  // Table variables
+  public ColumnMode = ColumnMode;
+  public SortType = SortType;
+  public SelectionType = SelectionType;
+  public columns: TableColumn[] = [];
+  public rows: any[] = [];
+  public selected: any[] = [];
+  public sortKeys: { [key: string]: keyof T } = {};
+  public filterKey: keyof T;
+  public totalModels: number;
+
+  /** API Error Response for Table Data */
+  public error: ApiErrorDetails;
+  /** API Error Response from Resolvers */
+  public failure: boolean;
+  public loadingData: boolean;
+  public models: ResolvedModelList = {};
+  public pageNumber: number;
+  public filterEvent$ = new Subject<string>();
+  private filters: Filters<M>;
+
+  constructor(
+    protected api: ApiFilter<any, any>,
+    public columnData: ColumnData<T, M>[],
+    private route?: ActivatedRoute,
+    private getUrlParameters: (component: any) => AbstractModel[] = () => []
+  ) {
+    super();
+    this.pageNumber = 0;
+    this.filters = {};
+
+    this.columnData.forEach((column) => {
+      this.columns.push({ name: column.name });
+      if (column.sortKey) {
+        this.sortKeys[column.key] = column.sortKey;
+      }
+      if (column.isFilterKey) {
+        this.filterKey = column.sortKey;
+      }
+    });
+
+    this.filterEvent$
+      .pipe(
+        debounceTime(defaultDebounceTime),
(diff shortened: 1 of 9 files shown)


src/app/helpers/tableTemplate/pagedTableTemplate.ts
```diff
@@ -13,11 +13,211 @@ import {
...
 
+export interface ColumnData<T, M extends AbstractModel> {
+  key: string;
+  name: string;
+  transform: (model: M) => any;
+  width?: number;
+  sortKey?: keyof T;
+  isFilterKey?: boolean;
+  cellTemplate?: () => any;
+}
```
atruskie, line 29: This type should maybe extend ngx-datatables own column type? https://swimlane.gitbook.io/ngx-datatable/api/column/inputs

Or maybe just use that type by default?

Especially since they have:
- automatic name inference
- autoResize, and more column width options (like flex grow)

atruskie: Yeah, they have `table.columns` input which accepts an array of strings. 

I think this component be best defined as:
- default: no options needed. Shows some default columns
- a string[] of columns to show (most columns don't need anything special done to them)
- accepts an array of [`TableColumn`](https://github.com/swimlane/ngx-datatable/blob/master/projects/swimlane/ngx-datatable/src/lib/types/table-column.type.ts)

src/app/helpers/tableTemplate/pagedTableTemplate.ts
```diff
@@ -13,11 +13,211 @@ import {
...
+  }
+
+  /**
+   * Create datatable rows
+   */
+  protected createRows(models: M[]) {
+    return models.map((model) => {
+      const row = {};
+      this.columnData.forEach(
+        (column) => (row[column.key] = column.transform(model))
```
atruskie, line 192: `TableColumn` has a `pipe` property that accepts a `pipeTransform` that could replace this

atruskie (commented):
I think the big issue is the coupled concerns of filtering/paging/sorting data with showing the data.

There should be a filter service which returns an observable, whose next function accepts `(filter, page, sort)`.

After that, the pagedTable function accepts this service object, hooks in for events, and renders models. The card UI for projects/sites/points should work the same way.

In this way we decouple the data from the presentation.

We can make use of ngx-datatables's `externalPaging` and `externalSorting` options too: https://github.com/swimlane/ngx-datatable/blob/master/src/app/paging/paging-server.component.ts

Also, why is this not its own component? Why do we need to extend another component to use it?

We shoud cover design requirements before any other work progresses

Allcharles:
This change is going to be temporarily halted and left for late development due to time constraints. While it possesses the ability to clean up a lot of boilerplate code, it is not currently a priority. Here are some resources that may help future developers understand the process undertaken:

- https://nils-mehlhorn.de/posts/angular-material-pagination-datasource
- https://nils-mehlhorn.de/posts/indicating-loading-the-right-way-in-angular

However, there is a major flaw in the pagination design the article uses, which makes it incompatible with QSP (cannot set an arbitrary page, it always increments from 1). This will need to be worked around.

Allcharles:
Closed in favour of #1907

PR #578, opened 2020-11-05
Cleanup

Allcharles:
# General Cleanup

General cleanup of tests and basic components.

## Changes

- Cleaned up home component
- Cleaned up app component and tests
- Updated secondary and action menus to handle disabled links
- Implemented menu layout for Page Not Found component
- Implemented ghost users for Permissions Shield Component
- Minor cleanup to baw api services
- Merged menu components into the `/components/shared/menu` folder and created a `@menu` import path

### Test Coverage Before PR:

```
=============================== Coverage summary ===============================
TOTAL: 1809 SUCCESS
Statements   : 85.77% ( 3405/3970 )
Branches     : 81.4% ( 906/1113 )
Functions    : 75.33% ( 962/1277 )
Lines        : 85.26% ( 3152/3697 )
================================================================================
```

### Test Coverage After PR:

```
TOTAL: 1850 SUCCESS
=============================== Coverage summary ===============================
Statements   : 86.45% ( 3444/3984 )
Branches     : 83.21% ( 912/1096 )
Functions    : 75.76% ( 969/1279 )
Lines        : 85.99% ( 3191/3711 )
================================================================================
```

### Page Not Found Page

Page not found link is disabled so that users are unable to click on it, this also makes it highlighted.

image

### Regular Users in Permissions Shield Component

image

### Ghost Users in Permissions Shield Component

image

## Final Checklist


src/app/app.component.ts
```diff
@@ -31,49 +40,62 @@ export class AppComponent extends WithUnsubscribe() implements OnInit {
...
+        // This is a display component, break out of while loop and check page info
+        break;
+      }
+
+      // Component has a child, keep digging further down
+      displayComponent = displayComponent.firstChild;
+      count++;
+    }
+
+    if (count === 50) {
```
atruskie:
```suggestion
    if (count >= 50) {
```

src/app/components/shared/menu/button/button.component.ts
```diff
@@ -7,20 +7,24 @@ import { MenuAction } from "@interfaces/menusInterfaces";
 @Component({
   selector: "baw-menu-button",
   template: `
+    <!--
+      No tooltips for disabled buttons because of:
+      https://github.com/ng-bootstrap/ng-bootstrap/issues/1250#issuecomment-274916839
+    -->
```
atruskie: if this is needed, I believe bootstrap's solution is to use a wrapper element to which the tool tip is applied

src/app/models/User.ts
```diff
@@ -92,10 +92,24 @@ export class User extends AbstractModel implements IUser {
...
+  }
+
+  public static get unknownUser(): User {
+    return new User({
+      id: -2,
+      email: "unknown-user@noreply.com.au",
+      userName: "Unknown User",
+    });
+  }
+
```
atruskie: these users should be constant, and frozen. Then they can be compared via reference equality.

Unless there's a specific need to return a different instance each time?

src/app/services/baw-api/account/accounts.service.ts
```diff
@@ -40,7 +42,21 @@ export class AccountsService extends StandardApi<User> {
...
-    return this.apiShow(endpoint(model, Empty));
+    return this.apiShow(endpoint(model, Empty)).pipe(
+      // Return unknown or deleted user depending on error code
+      catchError((err: ApiErrorDetails) => {
+        switch (err.status) {
+          case apiReturnCodes.unauthorized:
+            console.warn("Returning Unknown User");
+            return of(User.unknownUser);
+          case apiReturnCodes.notFound:
+            console.warn("Returning Not Found User");
```
atruskie: these probably don't need to be logged - maybe better to add comments and explain the situations for which these occur a little more.

- not found: only occurs when a user if soft-deleted
- unknown: only occurs when user is anonymous to server

atruskie (approved):

src/app/models/User.ts
Allcharles: My reasoning at the time was that it would prevent changes to the User object. However `Object.freeze()` fixes that

Allcharles: As it turns out, we cannot freeze these objects. The decorator which modifies the user image urls occurs during runtime and will fail if the user object is frozen.

atruskie (approved):

PR #627, opened 2020-11-27
Cleanup services

Allcharles:
# Extract service logic from components

This PR aims to extract repeated filter types from components, and into a service where is can more easily be tested, and logic is abstracted from the UI.

## Changes

- Tightened typing on model decorators
- Tightened typing on api service unit tests
- Added tests for custom filter api service functions
- Extracted some filter queries from components to api services

## Final Checklist


src/app/components/sites/site/site.component.ts
```diff
@@ -93,18 +90,23 @@ class SiteComponent extends PageComponent implements OnInit {
...
+    this.filterByDates("asc", { paging: { items: 1 } })
       .pipe(takeUntil(this.unsubscribe))
       .subscribe(
         (recordings) => (this.recordingsStart = recordings[0]?.recordedDate),
         (err) => console.log({ err })
       );
   }
+
+  private filterByDates(
+    direction: "desc" | "asc",
```
atruskie: this "desc" | "asc" string literal type is hardly unique. It should probably be referring to an existing type/alias?

src/app/helpers/advancedTypes.ts
```diff
@@ -26,3 +26,8 @@ export type XOR<T, U> = T | U extends Record<string, any>
  * Response may be a promise, or may return in real time
  */
 export type MayBeAsync<T> = T | Promise<T>;
+
+/** Extract keys from a type which return a specific type */
+export type KeysOfType<T, TProp> = {
```
atruskie, line 31: cool

src/app/models/AssociationDecorators.ts
```diff
@@ -11,28 +12,32 @@ import { User } from "./User";
  * Creates an association between the ownerId and its user model
  */
 export function owner<Parent extends AbstractModel & { ownerId?: Id }>() {
-  return hasOne<Parent, User>(ACCOUNT, "ownerId");
+  const key: keyof Parent = "ownerId";
+  return hasOne<Parent, User>(ACCOUNT, key as any);
```
atruskie: why is there a need to cast back to `any`?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -122,10 +122,10 @@ export abstract class BawApiService<Model extends AbstractModel> {
   /**
    * Retrieve user details from session cookie. Null if no user exists.
    */
-  public getLocalUser(): SessionUser | null {
+  public getLocalUser(): SessionUser | undefined {
```
atruskie: Comment out of date?

Possibility for null still exists down below, but then type signature out of date?

atruskie (approved):
I like how for the common service functions you've still allowed for customization by accepting more filters, which you then merge with whatever function specific filter you need. Very flexible.

src/app/models/AssociationDecorators.ts
Allcharles: Yes, for some reason typescript cant see that the key in this case will be used to return an `Id`. Couldn't figure out how to get typescript to like it so i added some intermediate typing and then changed it to any to pass the typecheck.

atruskie: Ah that makes the rest of it make sense now. It'll do 👍

PR #651, opened 2020-12-07
StrongRoute QSP

Allcharles:
# Strong Route Query Parameters

Added tracking of query parameters to the `StrongRoute`. This will allow for links in the action/secondary menus to handle dynamic parameters.

## Changes

- Added query parameters to `StrongRoute`
- Updated `MenuComponent` and `InternalLinkComponent` to handle applying `StrongRoute` query params to the link

## Final Checklist

### DIFF
--- a/src/app/components/security/pages/login/login.component.ts
+++ b/src/app/components/security/pages/login/login.component.ts
@@ -18,7 +18,6 @@ import {
   defaultErrorMsg,
   FormTemplate,
 } from "@helpers/formTemplate/formTemplate";
-import { AnyMenuItem } from "@interfaces/menusInterfaces";
 import { StrongRoute } from "@interfaces/strongRoute";
 import { List } from "immutable";
 import { ToastrService } from "ngx-toastr";
@@ -135,7 +134,7 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
 LoginComponent.linkComponentToPageInfo({
   category: securityCategory,
   menus: {
-    actions: List<AnyMenuItem>([
+    actions: List([
       confirmAccountMenuItem,
       resetPasswordMenuItem,
       unlockAccountMenuItem,
--- a/src/app/components/security/pages/login/login.component.ts
+++ b/src/app/components/security/pages/login/login.component.ts
@@ -108,7 +108,7 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
     if (this.redirectBack) {
       this.location.back();
     } else if (this.redirectUrl instanceof StrongRoute) {
-      this.router.navigateByUrl(this.redirectUrl.toString());
+      this.router.navigateByUrl(this.redirectUrl.toRouterLink());
     } else if (this.redirectUrl.startsWith("/")) {
       this.router.navigateByUrl(this.redirectUrl);
     } else {
--- a/src/app/components/security/pages/login/login.component.ts
+++ b/src/app/components/security/pages/login/login.component.ts
@@ -19,6 +19,7 @@ import {
   FormTemplate,
 } from "@helpers/formTemplate/formTemplate";
 import { AnyMenuItem } from "@interfaces/menusInterfaces";
+import { StrongRoute } from "@interfaces/strongRoute";
 import { List } from "immutable";
 import { ToastrService } from "ngx-toastr";
 import { fields } from "./login.schema.json";
@@ -41,7 +42,7 @@ import { fields } from "./login.schema.json";
 class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
   public fields = fields;
   private redirectBack: boolean;
-  private redirectUrl: string;
+  private redirectUrl: string | StrongRoute;
 
   public constructor(
     @Inject(API_ROOT) private apiRoot: string,
@@ -70,7 +71,7 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
       this.notifications.error("You are already logged in.");
     }
 
-    this.redirectUrl = homeMenuItem.route.toString();
+    this.redirectUrl = homeMenuItem.route;
     const noHistory = 1;
     const navigationId =
       (this.location.getState() as any)?.navigationId ?? noHistory;
@@ -78,7 +79,7 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
 
     // If no redirect, redirect home
     if (typeof redirect === "boolean" && !redirect) {
-      this.redirectUrl = homeMenuItem.route.toString();
+      this.redirectUrl = homeMenuItem.route;
       return;
     }
 
@@ -106,6 +107,8 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
   protected redirectUser() {
     if (this.redirectBack) {
       this.location.back();
+    } else if (this.redirectUrl instanceof StrongRoute) {
+      this.router.navigateByUrl(this.redirectUrl.toString());
     } else if (this.redirectUrl.startsWith("/")) {
       this.router.navigateByUrl(this.redirectUrl);
     } else {
--- a/src/app/components/shared/menu/internal-link/internal-link.component.spec.ts
+++ b/src/app/components/shared/menu/internal-link/internal-link.component.spec.ts
@@ -72,7 +72,7 @@ describe("MenuInternalLinkComponent", () => {
     defaultLink = menuRoute({
       icon: ["fas", "home"],
       label: "home",
-      route: StrongRoute.base.add("home"),
+      route: StrongRoute.newRoot().add("home"),
       tooltip: () => "tooltip",
     });
   });
@@ -142,7 +142,7 @@ describe("MenuInternalLinkComponent", () => {
         route: "/brokenlink",
         link: menuRoute({
           ...defaultLink,
-          route: StrongRoute.base.add("brokenlink"),
+          route: StrongRoute.newRoot().add("brokenlink"),
         }),
       });
       spec.detectChanges();
@@ -154,7 +154,7 @@ describe("MenuInternalLinkComponent", () => {
         route: "/brokenlink",
         link: menuRoute({
           ...defaultLink,
-          route: StrongRoute.base.add("wronglink"),
+          route: StrongRoute.newRoot().add("wronglink"),
         }),
       });
       spec.detectChanges();
--- a/src/app/components/shared/menu/internal-link/internal-link.component.spec.ts
+++ b/src/app/components/shared/menu/internal-link/internal-link.component.spec.ts
@@ -1,6 +1,7 @@
 import { Location } from "@angular/common";
 import { RouterLink, RouterLinkWithHref } from "@angular/router";
 import { RouterTestingModule } from "@angular/router/testing";
+import { StrongRouteDirective } from "@directives/strongRoute/strong-route.directive";
 import { MenuRoute, menuRoute } from "@interfaces/menusInterfaces";
 import { StrongRoute } from "@interfaces/strongRoute";
 import { createHostFactory, SpectatorHost } from "@ngneat/spectator";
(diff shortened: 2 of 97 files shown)


src/app/components/security/pages/login/login.component.ts
```diff
@@ -106,6 +107,8 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
   protected redirectUser() {
     if (this.redirectBack) {
       this.location.back();
+    } else if (this.redirectUrl instanceof StrongRoute) {
+      this.router.navigateByUrl(this.redirectUrl.toString());
```
Allcharles: Add a test for this scenario if none specifically exists

src/app/components/shared/menu/internal-link/internal-link.component.spec.ts
```diff
@@ -47,6 +48,7 @@ describe("MenuInternalLinkComponent", () => {
         [placement]="placement"
         [tooltip]="tooltip"
         [route]="route"
+        [qsp]="qsp"
```
Allcharles, line 46: Maybe a better name would be `[params]`? Maybe investigate to check that the RouterLink directive doesn't interfere with using `[queryParams]`

src/app/interfaces/strongRoute.ts
```diff
@@ -1,47 +1,64 @@
...
    * @param config Additional router configurations
    * @param isRoot Is this a root StrongRoute
    */
   private constructor(
-    parent: StrongRoute,
-    name: string,
-    config: Partial<Route>,
+    parent: StrongRoute = undefined,
+    name: string = StrongRoute.rootPath,
+    qsp: QSP = () => ({}),
```
Allcharles: Missing description

src/app/interfaces/strongRoute.ts
```diff
@@ -55,49 +72,47 @@ export class StrongRoute {
...
+    return new StrongRoute(this, name, qsp, config, true);
   }
 
   /**
    * Method used for templating a route with parameters.
    * Use this in a template like so:
-   * <a [routerlink]="route.Format({projectId: 1, siteId: 1})" />
+   * <a [routerlink]="route.format({projectId: 1, siteId: 1})" />
    */
   public format(args: { [key: string]: string | number }): string {
```
Allcharles: I would like to see this function merged into `toRoute()` and `toString()` as both are effectively useless if the StrongRoute has parameters

src/app/components/shared/menu/internal-link/internal-link.component.spec.ts
atruskie: this is one of the rare occasions I don't mind the abbreviation so much

also there should be no interfereance with router link

src/app/components/shared/menu/internal-link/internal-link.component.ts
```diff
@@ -34,6 +36,7 @@ export class MenuInternalLinkComponent implements OnChanges {
   @Input() public id: string;
   @Input() public link: MenuRoute;
   @Input() public route: string;
+  @Input() public qsp: Params;
```
atruskie: re last comment: if any name change, i'd change it so that it was the same as routerLink's, i.e. `queryParams`, if they behave the same way

src/app/helpers/advancedTypes.ts
```diff
@@ -31,3 +31,6 @@ export type MayBeAsync<T> = T | Promise<T>;
 export type KeysOfType<T, TProp> = {
   [P in keyof T]: T[P] extends TProp ? P : never;
 }[keyof T];
+
+/** Sets type to either be T or null */
+export type Potential<T> = T | null;
```
atruskie: rename this to `Option` - it's a common functional concept and we'll do better in the long run if we reuse name

https://en.wikipedia.org/wiki/Option_type

https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/options

src/app/interfaces/strongRoute.ts
```diff
@@ -1,47 +1,64 @@
...
-  public readonly parent: StrongRoute;
-  public readonly name: string;
-  public readonly isParameter: boolean;
-  public readonly fullRoute: string;
-  public readonly children: StrongRoute[] = [];
+  private static readonly rootPath = "";
+
+  /** Page component associated with this strong route */
+  public pageComponent: Potential<Type<PageComponent>>;
+  /** Root strong route of this strong route */
```
atruskie: Good docs do more than restate what is in the name of the documented element😸 

Maybe add why this is useful? E.g. allows us to form absolute routes/paths even if we only have access to current node.

Ditto this comment to all documented elements in this file

src/app/interfaces/strongRoute.ts
```diff
@@ -1,47 +1,64 @@
...
+  public readonly root: StrongRoute;
+  /** Parent strong route of this strong route */
+  public readonly parent?: StrongRoute;
+  /** Route path/name to add/create in the strong route tree */
+  public readonly name?: string;
+  /** Route query parameters */
+  private readonly qsp?: QSP;
+  /** Is this strong route a parameter (ie. :siteId) */
```
atruskie:
```suggestion
  /** Is this strong route segment a route parameter? (ie. :siteId) */
```

src/app/interfaces/strongRoute.ts
```diff
@@ -1,47 +1,64 @@
...
   private readonly full: StrongRoute[];
+  /** Additional configuration options to append to this strong route */
   private readonly config: Partial<Route>;
 
   /**
    * Constructor
    *
    * @param parent Components parent
-   * @param name Route name
+   * @param name Route name (do not include '/')
```
atruskie: should make this an actual validation as well as a doc

src/app/interfaces/strongRoute.ts
```diff
@@ -1,47 +1,64 @@
...
+  /** Is this strong route a parameter (ie. :siteId) */
+  private readonly isParameter: boolean;
+  /** Children strong routes of this strong route */
+  private readonly children: StrongRoute[] = [];
+  /** List of parameter strong routes in this strong routes path */
   private readonly parameters: StrongRoute[];
+  /** List of parent strong routes (including current strong route) */
   private readonly full: StrongRoute[];
+  /** Additional configuration options to append to this strong route */
   private readonly config: Partial<Route>;
```
atruskie: might do with a better name. Something like `angularRouteConfig` ?

src/app/interfaces/strongRoute.ts
```diff
@@ -1,47 +1,64 @@
...
-    parent: StrongRoute,
-    name: string,
-    config: Partial<Route>,
+    parent: StrongRoute = undefined,
+    name: string = StrongRoute.rootPath,
+    qsp: QSP = () => ({}),
+    config: Partial<Route> = {},
     isRoot?: boolean
   ) {
     this.root = this;
```
atruskie: it'd be good to not set this here, but instead set it down in the else case of the root check below

src/app/interfaces/strongRoute.ts
```diff
@@ -55,49 +72,47 @@ export class StrongRoute {
...
+      ...config,
+    };
   }
 
   /**
    * A base level (root) route component
    */
   public static get base() {
-    return new StrongRoute(null, null, {});
+    return new StrongRoute();
```
atruskie: is it intended to more than one base? If so this is a constructor helper and it should be a method

If not, this should not return a strong route every time it is called as a property

src/app/interfaces/strongRoute.ts
```diff
@@ -136,17 +151,24 @@ export class StrongRoute {
   /**
    * Compile the list of routes for a module
    *
-   * @param callback Callback function (usually: GetRouteConfigForPage)
+   * @param callback Callback function (usually: getRouteConfigForPage)
```
atruskie: Add: allows our 'pages' to add extra route data or modifications when they're set up

src/app/interfaces/strongRoute.ts
```diff
@@ -136,17 +151,24 @@ export class StrongRoute {
...
 
     const sortRoutes = (a: Route, b: Route): -1 | 0 | 1 => {
-      if (a.path === null && b.path === null) {
+      // Root route wins
+      if (a.path === StrongRoute.rootPath || b.path === StrongRoute.rootPath) {
+        return a.path === StrongRoute.rootPath ? -1 : 1;
+      }
+
+      const aPathExists = isInstantiated(a.path);
+      const bPathExists = isInstantiated(b.path);
```
atruskie: `Exists` seems misleading the paths/routes may exist

Maybe `IsInstantiated`

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
    * Example output: "/home/house"
    */
   public toString(): string {
-    return "/" + (this.fullRoute ? this.fullRoute : "");
+    // Prevent double '/' at start of route string
+    const path = this.toRoute().join("/");
+    return path.length >= 2 ? path.substr(1) : path;
```
atruskie: maybe better expressed with 
```suggestion
    return path.startsWith('/')  ? path.substr(1) : path;
```

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
    * Example output: "/home/house"
    */
   public toString(): string {
```
atruskie: Add doc.

toString returns a single string of all route components joined together. Useful for wherewe want to see a simple diagnostic representation of the strong route.

Should not be used in router links

example output: `/projects/:projectId/sites/:siteid?filter_name=:name`

src/app/interfaces/strongRoute.ts
atruskie: Yeah, no.

Format returns a whole, templated route (i.e. no missing parameters). It serves a distinct purpose, different from the other two methods

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
...
 
   /**
-   * Router representation of the route
-   * ! This will use a path relative to the current page if directly
-   * inserted into the [routerLink] directive
+   * Router representation of the route inserted into the
+   * [routerLink] directive
    *
-   * Example output: ["home", "house"]
+   * Example output: ["/", "home", "house"]
```
atruskie: Add:

Outputs in the routerlink @Input format, i.e. an array of args passed to `Router#createUrlTree` via `routerLink`. See more https://angular.io/api/router/RouterLink

example output: `['/', 'projects', ':projectId', 'sites', ':siteid']`

to use queryPararms call the `queryParams` method

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
...
-    return this.full.slice(1).map((x) => x.name);
+    if (this.full.length > 0) {
+      const fullRoute = this.full
+        .map((x) => x.name)
+        .filter((fragment) => !!fragment);
+
+      return ["/", ...fullRoute];
+    }
+
+    return ["/"];
```
atruskie: don't we have a string constant for `"/"`?

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
...
    *
-   * Example output: ["home", "house"]
+   * Example output: ["/", "home", "house"]
    */
   public toRoute(): string[] {
-    return this.full.slice(1).map((x) => x.name);
+    if (this.full.length > 0) {
+      const fullRoute = this.full
+        .map((x) => x.name)
+        .filter((fragment) => !!fragment);
```
atruskie: what's being filtered out here?

Do we have many names that are null?

src/app/interfaces/strongRoute.ts
atruskie: Make sure QSP are represented

src/app/interfaces/strongRoute.ts
atruskie: Example output:

`/projects/123/sites/456?filter_name=abc`

atruskie: Add argument to accept QSPs?

src/app/interfaces/strongRoute.ts
atruskie: Note there's nothing we can do here to validate if query params are being used or not.

When a user uses this in a routerLink, the routerLink args are passed to `@Input` `routerLinks` and the query params are passed separately to `queryParams`. These are two separate options and we can't control/validate that they both occur in the template

atruskie: What we could do is make our own `strongRouterLink` which ensures QSPs are provided or else fails.

We could also advise to always use `format` to ensure QSPs are formatted

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
...
+    return ["/"];
+  }
+
+  /**
+   * Query parameters which should be supplied with this route
+   * and should be inserted into the [queryParams] directive
+   *
+   * @param params Route data parameters
+   */
+  public queryParams(params: Params) {
```
atruskie: add usage example:

`<a [routerlink]="siteRoute.toRoute()" queryParams="siteRoute.qsp({filer_name: abc})" />`

src/app/interfaces/strongRoute.ts
```diff
@@ -192,18 +221,37 @@ export class StrongRoute {
...
+    return this.qsp(params);
```
atruskie: Does this do any validation?

atruskie (changes requested):
This got complex in a few ways.

Near the end I realised I made a few erroneous assumptions as to the purpose of the some of the strong route functions.

We should meet and clarify

Allcharles:
Image of comments:
image

src/app/directives/strongRoute/strong-route.directive.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Additional query parameters to apply to the StrongRoute. By
+   * default, all of the angular route parameters are already given
+   * to the StrongRoute.
+   */
+  @Input() public queryParams: Params;
+
+  /** Points to original classes 'router' */
+  private _router: Router;
+  /** Points to original classes 'route' */
```
atruskie: these comments don't add much value

src/app/directives/strongRoute/strong-route.directive.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+  });
+
+  // TODO Current implementation does not work with routerLinkActive
+  xit("should not interfere with routerLinkActive", () => {});
+
+  describe("strongRoute", () => {
+    it("should handle undefined strongRoute", () => {
+      setup(undefined);
+      spec.detectChanges();
+      expect(spec.directive).toBeTruthy();
```
atruskie: nit: never liked testing truthyness. Why not test it is an isntance of strong route directive class? Gives more informative error if/when test fails.

src/app/interfaces/strongRoute.ts
```diff
@@ -1,152 +1,311 @@
...
+   * List of StrongRoutes which have isParameter set to true in the hierarchy
+   * from the root StrongRoute, to the current StrongRoute
+   * (ie. `[":siteId", ":projectId"]`)
+   */
   private readonly parameters: StrongRoute[];
+  /**
+   * List of all StrongRoutes in the hierarchy going from the root StrongRoute, to
+   * the current StrongRoute
+   */
   private readonly full: StrongRoute[];
```
atruskie: these two fields seem redundant. We're implementing a linked list  here, why do we need to more than double memory usage by storing references to all items in the list in another array

atruskie (approved):

PR #807, opened 2021-01-27
Restyling Disabled Menu Item Links

### LINKED ISSUE #779: Emphasis Disabled Menu Links
Disabled links in the secondary and action menus could be misleading for users. More emphasis needs to be added to highlight that the links are disabled. Currently the light green colouring given when hovering over the menu item could confuse users into thinking the link is active.

Current State:

image

Allcharles:
# Restyling Disabled Menu Item Links

This is to cleanup the styling of the disabled menu item links in the secondary and action menus.

## Changes

- Changed styling of disabled menu items
- Merged `MenuInternalLinkComponent` and `MenuExternalLinkComponent` into a singular component as much of the code is the same
- Rewrote action menu unit tests
- Rewrote secondary menu unit tests
- Added `highlight` property to menu items to allow the not found menu item to appear active/highlighted regardless of route
- Removed behaviour where any disabled links would appear as active menu items

## Problems

- This will potentially cause some merge conflicts with the migration branch because of overlapping changes in the secondary menu unit tests

## Issues

- Closes #779 

### Disabled `MenuLinkComponent` and `MenuActionComponent`

image

### Disabled and Highlighted `MenuLinkComponent`

image

## Final Checklist


src/app/components/shared/menu/button/button.component.scss
```diff
@@ -13,6 +11,15 @@ button {
...
   }
+
+  &:disabled {
+    cursor: not-allowed;
+
+    &:hover,
+    &:focus {
+      background-color: transparent;
+    }
+  }
```
atruskie: this seems like a potentially reusable style, nothing about it needs it to be limited to the scope of a button. Perhaps refactor?

Also: rather than setting the background color to transparent, why not set it to `unset`? That should have the same effect and be semantically clearer. Also (probably no longer relevant) needless use of transparency historically affected page performance.

src/app/components/shared/menu/link/link.component.scss
```diff
@@ -1,3 +1,8 @@
+.disabled {
+  cursor: not-allowed;
+  opacity: 0.75;
+}
+
```
atruskie: bad indentation?

~~does this style compund with the .button:disabled style I saw before?~~

I think this is a good example for what I meant by refactoring the .button:disabled style.

src/app/components/shared/menu/link/link.component.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.link.active ||
+      this.link.highlight ||
+      currentRoute === this.location.path().split("?")[0];
+  }
+
+  private handleExternalLink(params: Params) {
+    const uri = this.externalLink.uri(params);
+
+    // Prepend apiRoot to relative links
+    this.href = uri.startsWith("/") ? this.apiRoot + uri : uri;
```
atruskie: hmmm

this is only correct at the moment? apiRoot should be the route of the API, the external link could be pointing anywhere?

while the old-client is currently hosted on the api server I didn't want to be tied to that. Maybe leave a TODO or a note, this seems like a point of silent failure

src/app/components/shared/menu/link/link.component.ts
```diff
@@ -0,0 +1,111 @@
...
+          <ng-container *ngTemplateOutlet="linkDetails"></ng-container>
+        </a>
+      </ng-template>
+    </span>
+
+    <!-- Link Details -->
+    <ng-template #linkDetails>
+      <div class="icon"><fa-icon [icon]="link.icon"></fa-icon></div>
+      <span id="label">{{ link.label }}</span>
+      <span class="d-none" [id]="id">{{ tooltip }}</span>
```
atruskie: Why does the tooltip need an ID?

src/app/interfaces/menusInterfaces.ts
```diff
@@ -90,6 +90,11 @@ export interface MenuItem extends LabelAndIcon {
    * The indentation of this link
    */
   indentation?: number;
+  /**
+   * Force highlighting of this link, this is similar to setting the links `active`
+   * property to true.
+   */
```
atruskie, line 97: Add note for when this might be needed?

src/app/components/error/error.menus.ts
```diff
@@ -7,5 +7,6 @@ export const pageNotFoundMenuItem = menuRoute({
   label: "Page Not Found",
   route: pageNotFoundRoute,
   disabled: true,
+  highlight: true,
```
atruskie, line 16: Add note for why this is needed

atruskie (changes requested):
A few minor changes, mostly questions. Looks great otherwise.

src/app/components/shared/menu/link/link.component.ts
Allcharles: The ID is to help screen readers, however looking online this doesn't appear to be a standard and I'm not sure where i got the example from. The tooltip generated by ngbTooltip should be sufficient for screen readers i would hope

atruskie: Yeah I don't see how a sequentially repeating auto-id would help.

Oh I see. if ngb is using the `aria-describedby="#some-id"` method of linking the target and the tooltip then this makes sense...

Confirmed:

https://github.com/ng-bootstrap/ng-bootstrap/blob/c332b8ba630fc893aaa3513d68a6a6d14322321d/src/tooltip/tooltip.ts#L205-L218

And looks like it is auto-generated:

https://github.com/ng-bootstrap/ng-bootstrap/blob/c332b8ba630fc893aaa3513d68a6a6d14322321d/src/tooltip/tooltip.ts#L151

And unless I'm misreading this, the `id` input is not even used https://github.com/ng-bootstrap/ng-bootstrap/commit/088363551b5d5ecd602a82aedfc97d43a75a080f

atruskie (approved):
💯

PR #808, opened 2021-01-27
Google Maps Fix

### LINKED ISSUE #1338: Google maps causing unit test failures
Some component tests load the google maps bundle, which can cause random errors to be thrown during tests. This can be fixed in two ways:

a) Changing how google maps is loaded during unit tests
b) Creating a mock wrapper for components which tangentially depend on google maps (ie. forms which use the `LocationInputComponent`)

Allcharles:
# Fixed broken tests due to google maps update

A few unit tests were failing because of a change in how google maps functions. The underlying system has been written to be more fault tolerant, and less likely to fail outside of the `MapComponent`.

## Changes

- Modified `MapComponent` to better lazy load google maps api bundle according to angular documentation
- Created `MapService`
- Removed need for `embedGoogleMaps` function as embedding is no longer performed

## Issues

Fixes #1338 

### No Markers

image

### Map Loading

image

### Map Success

image

### Map Failure

image

## Final Checklist


src/app/services/map/map.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    return this.isMapLoaded$.getValue();
+  }
+
+  /**
+   * Determine if a map marker is valid
+   */
+  public isMarkerValid(marker: MapMarkerOption): boolean {
+    return (
+      typeof marker?.position?.lat === "number" &&
+      typeof marker.position.lng === "number"
```
atruskie: could add some bounds validations here too

- not null/undefined
- lat in [-90, 90]
- long in [-180, 180]

src/app/services/map/map.service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Load google maps bundle
+   */
+  protected loadMap() {
+    // Only add api key if it exists
+    let mapsUrl = MapService.mapUrl;
+    if (this.apiKey) {
+      mapsUrl += "?key=" + this.apiKey;
+    }
```
atruskie: otherwise this is some kind of error state?

It means the environment failed to load or it is misconfigured?

Perhaps console.error or set a field the map component can bind to to display error state?

Probably only in production mode - i imagine you still want it to work without a key in development

atruskie: Seeing what you did below, I'd suggest 

```
if (environment.production && !this.apiKey) {
    this.isMapLoaded$.next(MapState.failure);
    return;
}
```

src/app/services/map/map.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    // Jsonp requests allow us to run the api bundle without
+    // dealing with cors issues
+    this.http
+      .jsonp(mapsUrl, "callback")
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe(
+        () => this.isMapLoaded$.next(MapState.success),
+        () => this.isMapLoaded$.next(MapState.failure)
+      );
+  }
```
atruskie: reading up on jsonp suggests to me it is deprecated for good reason in favour of CORS.

This function is also ambiguous - I had to refresh my knowledge of JSON-P to understand we're inserting the content into a script tag. That's clever!

Minimal change: document behaviour.
Better change: insert the tag and bind to an event (examples: https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement#examples)

atruskie (approved):
Great work, great simplification.

I've left a few comments on the service itself that I think would be beneficial

src/app/services/map/map.service.ts
Allcharles: I'm going with the documentation approach. Running away from the insert tag approach because its what caused the tests to error out for unknown reasons.

atruskie (approved):
💯

Allcharles:
I'm having some major issues with testing the maps component, and any components which depend on it. I may need to rewrite some things to make it easier to import mock map components/services.

PR #877, opened 2021-02-16
Model viewUrl query parameters

### LINKED ISSUE #772: AbstractModel `viewUrl`s
Currently, the `viewUrl` method for AbstractModels can only handle setting the url path, and route parameters (by combining the StrongRoute with the routeParams. This is limiting as the model is unable to affect the query parameters of its route. There are two methods of bypassing this problem:

- Create a MenuItemDirective which has the following properties:
  - Accepts in the following input:

  ```typescript
  @Directive({
    selector: "a[menuItem]"
  })
  class MenuItemDirective {
    @Input() menuItemPath: {
      menuItem: NavigatableMenuItem,
      routeParams: RouteParams,
      queryParams: Params
    }
  ```

  - This will function similar to the StrongRouteDirective, with the added benefit of only requiring one input, and will allow interchangable use of `MenuLink`s and `MenuRoute`s
  - This will require adding a default route for external routing similar to the following: https://adrianfaciu.dev/posts/angular-router-external-links/
  - The downside to this solution is that it will require a large chunk of changes to existing code
- Models create copies of the StrongRoute and modify the object to have the correct routeParams and queryParams:
  - Requires minor modification to the StrongRoute class (add`.copy()`, `defaultRouteParams`, `defaultQueryParams`)
  - Requires some changes to angular UI code to change from `<a [routerLink]="model.viewUrl"/>` to `<a strongRoute [strongRoute]="model.viewUrl />"
  - Before

  ```typescript
    return libraryMenuItem.route.toRouterLink({projectId: this.id});
  }
  ```

  - After

  ```typescript
  public get viewUrl(): StrongRoute {
    const route = libraryMenuItem.route.copy();
    route.defaultRouteParams = (params) => ({customValue: true});
    route.defaultQueryParams = (params) => ({customValue: true});
    return route;
  }
  ```

- Rewrite how StrongRoute works so that we can completely bypass this issue with `RouterLinkWithHref` and `StrongRouteDirective` requiring multiple inputs

Allcharles:
# Update viewUrl to handle query paramaters

Updated how the `viewUrl` for models works, and how they interact with the angular router.

## Changes

- Created `UriDirective`
- Updated models to return viewUrls including query parameters

## Issues

Closes #772 

## Final Checklist


src/app/test/helpers/html.ts
```diff
@@ -99,7 +98,20 @@ export function assertHref(
...
+/**
+ * Assert uri link
+ *
+ * @param target Target element
+ * @param route Route text
+ */
+export function assertUri(target: HTMLElement, route: string) {
+  expect(target).toBeTruthy("No route detected");
+  expect(target).toHaveAttribute("ng-reflect-uri");
+  expect(target).toHaveAttribute("href", encodeURI(route));
```
atruskie: rename uri to url

src/app/models/Tag.ts
```diff
@@ -59,10 +59,7 @@ export class Tag extends AbstractModel implements ITag {
   }
 
   public get viewUrl(): string {
-    console.warn("Tag viewLibraryUrl method not implemented");
-    // /library?reference=all&tagsPartial=[tag.text]
-    // TODO This link is not working (issue #772)
-    return libraryMenuItem.route.toRouterLink({
+    return libraryMenuItem.route.format(undefined, {
```
atruskie, line 62: this `undefined` seems strange - is there a reason for it?

I'm going to guess we're omitting a route parameter? But which one?

src/app/models/Script.ts
```diff
@@ -71,6 +71,11 @@ export class Script extends AbstractModel implements IScript {
   }
 
   public get viewUrl(): string {
-    return adminScriptMenuItem.route.toRouterLink({ scriptId: this.id });
+    console.warn("Script viewUrl not implemented.");
+    return "not_implemented";
```
atruskie: throw - this will just lead to a _misleading_ 404

Is there a reason we can't just return the properly formed url? It will still 404 but at least it is self fixing when the associated view is eventually created

src/app/models/SavedSearch.ts
```diff
@@ -50,7 +50,7 @@ export class SavedSearch extends AbstractModel implements ISavedSearch {
   }
 
   public get viewUrl(): string {
-    console.error("SavedSearch viewUrl not implement");
+    console.warn("SavedSearch viewUrl not implemented.");
     return "not_implemented";
```
atruskie: See comments on script.ts - should not return "not_implemented"

src/app/models/AbstractModel.ts
```diff
@@ -42,6 +43,7 @@ export abstract class AbstractModel {
   public readonly kind: string;
 
   /**
+   * TODO Update comment
```
atruskie: carpe diem

src/app/directives/uri/uri.directive.ts
```diff
@@ -0,0 +1,51 @@
...
+  Params,
+  Router,
+  RouterLinkWithHref,
+  UrlTree,
+} from "@angular/router";
+import { withUnsubscribe } from "@helpers/unsubscribe/unsubscribe";
+
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: "a[uri]",
```
atruskie: `uri` ➡`url`

atruskie: I know we talked about documenting over namespacing, but https://angular.io/guide/styleguide#directive-custom-prefix is a compelling argument. 

Suggested `bawUrl`

src/app/directives/uri/uri.directive.ts
```diff
@@ -0,0 +1,51 @@
...
+})
+export class UriDirective extends withUnsubscribe(RouterLinkWithHref) {
+  @Input() public uri: string;
```
atruskie: Does not match specification from #772: `string | UrlTree`

src/app/directives/uri/uri.directive.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    // Construct URL from uri, baseURI does not matter
+    const url = new URL(this.uri, document?.baseURI || "http://localhost/");
+
+    // Extract query parameters from uri
+    const queryParams: Params = {};
+    url.searchParams.forEach((value, key) => (queryParams[key] = value));
+
+    return this._router.createUrlTree([url.pathname], {
```
atruskie: Why not: [`parseUrl`](https://angular.io/api/router/Router#parseurl) as defined in spec?

It handles the all the cases you handle manually: https://github.com/angular/angular/blob/e3f09ce845cef9fd2c89d39d0d822114c51e68fa/packages/router/src/url_tree.ts#L289-L294

atruskie: You can use object.assign to then merge in the base directive params with the created urlTree

src/app/directives/strongRoute/strong-route.directive.ts
```diff
@@ -67,7 +63,11 @@ export class StrongRouteDirective
...
       queryParamsHandling: this.queryParamsHandling,
-      preserveFragment: this.preserveFragment,
+      preserveFragment: attrBoolValue(this.preserveFragment),
     });
   }
 }
+
+function attrBoolValue(s: any): boolean {
+  return s === "" || !!s;
+}
```
atruskie, line 73: repeated function in uri.directive.ts

atruskie (changes requested):
Fantastic work. 

Minor changes:
-  `uri` should be `url`
- We should add a section to contributing.md that explains what custom directives are in use (particularly since these could appear magical or external/angular supplied directives)
- You used the `string` and `format` pairing a lot. Is there a reason for that over `UrlTree` and `formatUrlTree`? Technically it's more work for the browser - though it is unlikely to be a significant performance penalty
- I think you need a consistent approach for dealing with urls that are not yet ready. In this review I saw:
    - throwing an exception
    - returning `""`
    - returning `"not_implemented"`
    - returning `""/play_url"`
    - various combinations of the above with and without console errors or warns

src/app/models/Tag.ts
Allcharles: `libraryMenuItem` has no route parameters

Allcharles:
Alright, I see a few comments related to `UrlTree` and `parseUrl` so I'll answer them here. I can only create `UrlTree`s if I have access to the `Router` and `ActivatedRoute` injectables unfortunately. As the models are currently implemented, they do not have guaranteed access to the angular DI. So I cannot use it as a viable system. As for `parseUrl`, as I suspected during our talks, it sanitizes query parameters which defeats the purpose of using it. the `URL` class handles all of the logic we need with minimal hacking.

Edit: You linked to `URLParser` which appears that it might handle what i want, i will give it a try

atruskie (approved):
Perfect

Issue #614, opened 2020-11-23
Implement Data Request Page


Allcharles:
Implement the data request page. Through this page, users should be able to download the annotations for a point, site, project, user, or a custom request from the above. Currently, the forms are only mocked up and is not functional. This page should be affected by the url parameters which can specify a point/site/project/user id.

The following permutations should exist:

- Data Request
- Data Request for site (projectId, siteId, timezone)
- Data Request for user (userId, timezone)

Issue #610, opened 2020-11-23
Broken Forms


Allcharles:
List of broken, or partially implemented forms in the workbench-client:

- [x] Contact Us
  - API route does not exist (QutEcoacoustics/baw-server#415)
- [ ] Admin > Scripts
  - [ ] New
    - API route does not exist (QutEcoacoustics/baw-server#435)
  - [ ] Edit
    - API route does not exist (QutEcoacoustics/baw-server#435)
- [ ] Admin > Tags
  - [ ] New
    - Notes field is incorrectly formatted
    - Form does not handle custom error info
  - [ ] Edit
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/449)
    - Form does not handle custom error info
  - [ ] Delete
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/449)
- [x] Data Request
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/416)
- [ ] Profile
    - [ ] Edit Profile
      - Should prefill email
      - Form submission fails (https://github.com/QutEcoacoustics/baw-server/issues/429 ?)
    - [ ] Edit Password
      - Form submission fails (https://github.com/QutEcoacoustics/baw-server/issues/429 ?)
    - [ ] Delete Profile
      - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/429)
- [x] Report Problem
  - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/414)
- [x] Security
  - [x] Confirm Account
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/506)
  - [x] Reset Password
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/409)
  - [x] Unlock Account
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/505)
  - [x] Register
    - API route does not exist (https://github.com/QutEcoacoustics/baw-server/issues/424)

Allcharles:
Register route does not provide proper error reporting for users. If for example, the username has already been taken, the error message will only tell the user that they are not logged in. This would be fixed by creating the JSON API endpoint, or creating some regex scripts to extract the various error messages from the response HTML and error the observable if a match is found. Similar issues will plague all forms which are implemented using the hacky method of replicating baw-server's own internal requests

hudson-newey:
[...]

PR #968, opened 2021-04-12
Data request Page

Allcharles:
# Implemented Data Request Page

## Changes

- Created `simpleFormTemplate` class for forms which do not navigate after submission
  - Potentially should rename to make `simpleFormTemplate` and `formTemplate` to simpler to understand the difference
- Updated @voo/tzdb library for https://github.com/vvo/tzdb/pull/163
- Implemented Data Request page and forms
- Simplified CheckboxComponent formly declaration so that wrapper does not need to be specified anymore
- Set default timezone value to timezone of host server (potentially set to a different default?)
- Attempt to cleanup variable names in resolvers

## Problems

- Did not add tests because either, its too tricky to implement (ie. Blob downloading), or page has temporary implementation

## Issues

Deals with #614
Deals with #610

## Final Checklist


src/app/app.helper.ts
```diff
@@ -29,3 +29,29 @@ export const formlyRoot: ConfigOption = {
...
+   */
+  const anchorEl = window.document.createElement("a");
+  anchorEl.href = window.URL.createObjectURL(blob);
+  anchorEl.download = filename;
+  anchorEl.hidden = true;
+  document.body.appendChild(anchorEl);
+  // Blocked by IE
+  anchorEl.click();
+  document.body.removeChild(anchorEl);
+}
```
atruskie, line 57: Why is this needed?

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
+import { Component, Input, OnInit } from "@angular/core";
```
atruskie, line 1: I would name this `annotations-download` - the name `annotations` is likely a name we'll want to use elsewhere for other annotation related things

**EDIT**: I see you've given the component a more precise name, change name of file to match?

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+import { Region } from "@models/Region";
+import { Site } from "@models/Site";
+import { User } from "@models/User";
+import { ToastrService } from "ngx-toastr";
+import { Observable } from "rxjs";
+import { takeUntil } from "rxjs/operators";
+import { fields } from "./annotations.schema.json";
+
+@Component({
+  selector: "baw-request-annotations",
```
atruskie, line 22: `baw-download-annotations`

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+  template: `
+    <baw-form
+      submitLabel="Download Annotations"
+      [model]="model"
+      [fields]="fields"
+      [submitLoading]="loading || failure"
+      (onSubmit)="submit($event)"
+    >
+      <span id="subTitle">
+        Please select the timezone for the CSV file containing annotations for
```
atruskie, line 32: Let's take the chance to improve this copy:

> The annotations in the CSV will have all their dates and times set to a time zone of your choice.
> The default time zone is the local time for the site where the audio was recorded.
>
> For example, annotations created for audio from Brisbane will have dates and times set to AEST (+10).
>
> However, if you have recordings from Brisbane and Perth, which time zone do we choose for all downloaded events?
> AEST (+10) or AWST (+8) or UTC (+0)?
>
> It depends on how you want to work with your data, so the choice is yours.
> Please select the time zone you wish to use:

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+    notifications: ToastrService,
+    route: ActivatedRoute,
+    router: Router
+  ) {
+    super(
+      notifications,
+      route,
+      router,
+      () =>
+        "Successfully downloaded annotations. Please accept pop-ups if they are disabled.",
```
atruskie, line 62: Pop-ups should not be required

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+      false
+    );
+  }
+
+  public ngOnInit() {
+    super.ngOnInit();
+
+    if (this.failure) {
+      return;
+    }
```
atruskie, line 73: This seems like a specific case - what is going on here?

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const { timezoneName, projectId } = this.route.snapshot.queryParams;
```
atruskie, line 75: maybe be region or site id specified as well

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (timezoneName) {
+      this.model.timezone = timezoneName;
+    }
```
atruskie, line 79: This information should be pulled from the project/site/region id's

For projects: select the most frequent timezone from the sites within it

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (this.user) {
+      this.modelName = this.user.userName;
+      return;
+    }
```
atruskie, line 84: why is a username needed?

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (this.region || this.site) {
+      this.modelName = (this.region ?? this.site).name;
```
atruskie, line 87: what is `modelName` for?

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+})
+export class AnnotationsComponent
+  extends SimpleFormTemplate<Model>
+  implements OnInit {
+  public fields = fields;
+  @Input() public user: User;
+  @Input() public region: Region;
+  @Input() public site: Site;
+  public modelName: string;
+  private project: Project;
```
atruskie, line 46: Document these

src/app/components/data-request/annotations/annotations.component.ts
```diff
@@ -0,0 +1,135 @@
...
+      );
+    } else {
+      return this.siteApi.downloadAnnotations(
+        this.site,
+        this.project,
+        model.timezone
+      );
+    }
+  }
+}
```
atruskie, line 131: should just be forming a URL

src/app/components/data-request/data-request.component.ts
```diff
@@ -1,79 +1,151 @@
...
+
+    <!-- Display error if exists -->
+    <ng-container *ngIf="error">
+      <baw-error-handler [error]="error"></baw-error-handler>
+    </ng-container>
+
+    <!-- Display spinner if loading -->
+    <ng-container *ngIf="loadingModel">
+      <baw-loading></baw-loading>
+    </ng-container>
```
atruskie, line 58: let's split this into two separate pages - there's no need for them to be on the same page

src/app/components/data-request/data-request.menus.ts
```diff
@@ -6,10 +6,12 @@ import { StrongRoute } from "@interfaces/strongRoute";
...
   (params: Params) => ({
-    userId: params?.userId,
-    projectId: params?.projectId,
+    // Handle params from workbench-client and baw-client
+    userId: params?.userId ?? params?.selected_user_id,
+    projectId: params?.projectId ?? params?.selected_project_id,
     regionId: params?.regionId,
-    siteId: params?.siteId,
+    siteId: params?.siteId ?? params?.selected_site_id,
+    timezoneName: params?.timezoneName ?? params?.selected_timezone_name,
```
atruskie, line 14: I appreciate that you added support for this, but this link will never be generated by the old-client, and the server will never need to redirect a user to this page with the old style QSPs, so redundant?

src/app/components/shared/formly/custom-inputs.module.ts
```diff
@@ -19,9 +19,24 @@ import { PasswordConfirmationInputComponent } from "./password-confirmation-inpu
...
+  {
+    name: "timezone",
+    component: TimezoneInputComponent,
+    defaultOptions: {
+      /*
+       * Sets default timezone selection to location of server. This is
+       * a solution to bypass an exception thrown by ngx-formly:
+       * https://github.com/ngx-formly/ngx-formly/issues/1410
+       */
+      defaultValue: Intl.DateTimeFormat().resolvedOptions().timeZone,
```
atruskie, line 37: Not the server? This is client code?

src/app/helpers/advancedTypes.ts
```diff
@@ -34,3 +34,9 @@ export type KeysOfType<T, TProp> = {
 
 /** Sets type to either be T or null */
 export type Option<T> = T | null;
+
+export interface FixedLengthTuple<T extends any, L extends number>
```
atruskie, line 38: A tuple by definition is fixed length

src/app/services/baw-api/account/accounts.service.ts
```diff
@@ -67,9 +68,25 @@ export class AccountsService extends StandardApi<User> {
...
+    if (!isInstantiated(localUserId)) {
+      return throwError(
+        new Error("Unauthorized access, please log in first.")
+      ).pipe(catchError(this.handleError));
+    }
+
+    return this.apiDownload(
+      endpoint(localUserId, emptyParam),
+      selectedTimezone
+    );
```
atruskie, line 86: why is a user id needed

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -251,6 +252,25 @@ export class BawApiService<Model extends AbstractModel> {
...
+    const downloadPath = `${path}audio_events/download?selected_timezone_name=${selectedTimezone}`;
+    return this.http
+      .get(this.getPath(downloadPath), { responseType: "arraybuffer" })
+      .pipe(
+        map((data: ArrayBuffer) => new Blob([data], { type: "text/plain" })),
+        map((blob: Blob) => downloadBlob(blob, "annotations.csv")),
+        catchError(this.handleError)
+      );
+  }
+
```
atruskie, line 255: ...should not be needed with other suggested changes

src/app/services/baw-api/region/regions.service.ts
```diff
@@ -61,6 +61,17 @@ export class RegionsService extends StandardApi<Region, [IdOr<Project>]> {
...
+  public downloadAnnotations(
+    model: IdOr<Region>,
+    project: IdOr<Project>,
+    selectedTimezone: string
+  ): Observable<void> {
+    return this.apiDownload(
+      endpoint(project, model, emptyParam),
+      selectedTimezone
+    );
+  }
```
atruskie, line 74: this should just be a route to the download annotations page

src/app/services/baw-api/site/sites.service.ts
```diff
@@ -98,6 +98,17 @@ export class SitesService extends StandardApi<Site, [IdOr<Project>]> {
...
+  public downloadAnnotations(
+    model: IdOr<Site>,
+    project: IdOr<Project>,
+    selectedTimezone: string
+  ): Observable<void> {
+    return this.apiDownload(
+      endpoint(project, model, emptyParam),
+      selectedTimezone
+    );
+  }
```
atruskie, line 111: a above, should not be a service call

atruskie (changes requested):
I'm not convinced you need the separate form component. If you don't want the form to navigate, why not just don't navigate by default?

I'm quite confused with how the annotation download component works. It looks like it takes a combination of `@input` bindings and QSPs? AFAIK it should only take QSP...

I think this stems from jamming two forms into one page. Split the pages and this should be cleaned up.

Default timezone for download annotations is the local time zone of the `site`, or fallback to local timezone of the browser.

The download annotations page should just generate a href button link. Doesn't need to be a form, doesn't need an API service (currently)

Allcharles:
Closed in favour of #1044

PR #996, opened 2021-04-21
Fix create/update API requests

### LINKED ISSUE #966: Remove id from update requests
Baw-server will reject requests to update a model if the models id is part of the body. To remove this issue, update the `apiUpdate` command to remove the `id` property of a model

Allcharles:
# Fix create/update api requests

API requests for create/update models failed due to a change in the strictness of the baw-server. This update gives us greater control over what attributes of a model will be sent on specific types of requests.

## Changes

-Added `update` and `create` properties to `bawPersistAttr` and related attribute decorators

## Problems

- Some of the updated permissions may be wrong, will have to validate as we use the website

## Issues

Fixes #966 

## Final Checklist


src/app/models/AbstractModel.ts
```diff
@@ -138,3 +154,5 @@ export function getUnknownViewUrl(errorMsg: string) {
   console.warn(errorMsg);
   return unknownViewUrl;
 }
+
+export type ToJsonOptions = XOR<{ create: boolean }, { update: boolean }>;
```
Allcharles: Rename to `toJSONOptions`?

src/app/helpers/advancedTypes.ts
```diff
@@ -34,3 +34,6 @@ export type KeysOfType<T, TProp> = {
 
 /** Sets type to either be T or null */
 export type Option<T> = T | null;
+
+/** Removed readonly flag from all properties in an object */
+export type Writeable<T> = {-readonly [P in keyof T]: T[P]}
```
Allcharles: Highlight that this only affects the types, and does not change a read-only property to a writeable one

src/app/models/AbstractModel.ts
atruskie: Since this is a type I think the adverb `To` aspect of this is the more incorrect aspect. The word `Json` is also overly specific.

Maybe `ApiPersistenceOptions` ?

src/app/helpers/advancedTypes.ts
atruskie: Also, readonly in the model, may not be the same as readonly in the API

src/app/models/AbstractModel.ts
```diff
@@ -60,18 +67,27 @@ export abstract class AbstractModel<Model = Record<string, any>> {
...
    * Convert model to JSON
    */
-  public toJSON() {
-    const output = {};
-    this[AbstractModel.attributeKey].forEach((attribute) => {
+  public toJSON(opts?: ToJsonOptions): Partial<this> {
+    const output: Partial<Writeable<this>> = {};
+    let keys: string[];
+    if (!opts) {
+      keys = Object.keys(this).filter((key) => key !== "injector");
```
atruskie, line 74: does typescript have a `nameof()` construct for `"injector"`

src/app/models/AbstractModel.ts
atruskie: Upon understanding the context more: `ModelSerializationOptions`?

src/app/models/AttributeDecorators.ts
```diff
@@ -2,23 +2,42 @@ import { Id, Ids, ImageSizes, ImageUrl } from "@interfaces/apiInterfaces";
 import { DateTime, Duration } from "luxon";
 import { AbstractModel } from "./AbstractModel";
 
+function persistAttr(model: AbstractModel, key: string, opts: boolean | BawPersistAttributeOptions) {
+    model[AbstractModel.createAttributesKey] ??= [];
+    model[AbstractModel.updateAttributesKey] ??= [];
```
atruskie, line 7: could this be a static/prototypical property?
Or else done on initialization in constructor?

src/app/models/AbstractModel.ts
atruskie: It also seems like this type (although cool) is overly complex.

can't the parameter just be a string literal of `"create" | "update"`?

atruskie (approved):
Great change.

Only questions and cosmetic changes in review.

src/app/models/AbstractModel.ts
Allcharles: It appears that only types have the ability to use string literals. Afaik, it wont be a feature until at least TS4.4: https://stackoverflow.com/a/65879197/1937822

src/app/models/AttributeDecorators.ts
Allcharles: Is it not already statically typed?

src/app/models/AbstractModel.ts
Allcharles: Not sure what you mean

src/app/models/AttributeDecorators.ts
atruskie: Static as in an attribute of the class/type/prototype and not the instance

src/app/models/AbstractModel.ts
atruskie: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof

https://schneidenbach.gitbooks.io/typescript-cookbook/content/nameof-operator.html

src/app/models/AttributeDecorators.ts
Allcharles: Will be part of #1005

Issue #988, opened 2021-04-19
Increased test failure rate in CI


Allcharles:
Number of tests randomly failing on the CI has increased. There is a test which is writing to localstorage asynchronously and not being removed which is causing issues for other tests. Attempt to track it down and fix it.

Allcharles:
Need to update the Register page, unit tests are reliant on localstorage and can sometimes fail

PR #1048, opened 2021-05-13
Production builds

### LINKED ISSUE #1043: Build pipeline
Need to implement some additional steps to the build pipeline to prepare for deployments. Goals include the following:

- [x] Build versions of website should display version (footer)
- [x] Build versions of website should display git hash (footer)
- [x] Docker container should contain labels (https://github.com/QutEcoacoustics/audio-analysis/blob/master/build/Dockerfile#L7-L16)
- [x] Docker container should contain tags (https://github.com/QutEcoacoustics/audio-analysis/blob/master/build/build_docker.ps1#L18-L21)
  - May not need to add `latest` tag
  - May want to replace `stable` with `beta` until workbench-client is the official version
- [x] Each new version should generate a git tag pointing to the commit
- [ ] (optional) Automatically generate CHANGELOG

Versioning of the builds should follow calendar versioning as highlighted in #414

Allcharles:
# Production Builds

Enhance the production builds to include calendar version (calver)

## Changes

- Insert version into footer of website
- Updated docker container to contain labels and tags
- Tag each commit attached to a docker container

## Problems

- Have not yet validated the full workflow will function. Will have to wait for push to master to see if things work

## Issues

Fixes #1043 
Working on #988 

### Website Footer

image

### Github Tag

image

## Final Checklist


Dockerfile
```diff
@@ -1,5 +1,19 @@
 FROM node:current-alpine
 
+ARG GIT_COMMIT
+ARG WORKBENCH_CLIENT_VERSION="latest"
+
+LABEL maintainer="Charles Alleman <alleman@qut.edu.au>" \
+      description="Production environment for workbench client server" \
+      version="3.0" \
```
atruskie: how is `version` different from `org.ecosounds.version`

Dockerfile
```diff
@@ -1,5 +1,19 @@
 FROM node:current-alpine
 
+ARG GIT_COMMIT
+ARG WORKBENCH_CLIENT_VERSION="latest"
```
atruskie: This should not default to `latest` - latest is a tag (alias) for a version, not a version itself

build/build_docker.ps1
```diff
@@ -0,0 +1,25 @@
...
+
+# builds and pushes a docker file to docker hub
+# currently only builds the 'beta' tag, which is applied to whatever version
+# is supplied to this script.
+
+# get the current git commit
+$GIT_LONG_COMMIT = git rev-parse --long HEAD
+$GIT_SHORT_COMMIT = git rev-parse --short HEAD
+$SHORT_VERSION = docker run --rm relizaio/versioning -s YYYY.0M.0D.Calvermodifier -i beta
+$LONG_VERSION = docker run --rm relizaio/versioning -s YYYY.0M.0D.Calvermodifier+Metadata -i beta -m $GIT_SHORT_COMMIT
```
atruskie: document the options used in the format string inline - or at least link to the documentation

Also, should quote the format string

build/build_docker.ps1
```diff
@@ -0,0 +1,25 @@
...
+  -t qutecoacoustics/workbench-client:$SHORT_VERSION `
+  . `
+  --build-arg GIT_COMMIT=$GIT_LONG_COMMIT `
+  --build-arg WORKBENCH_CLIENT_VERSION=$LONG_VERSION
+
+# docker run -p 4000:4000 -v "$(pwd)/src/assets/environment.json:/environment.json" qutecoacoustics/workbench-client:beta
+docker push --all-tags qutecoacoustics/workbench-client
+
+git tag -a $SHORT_VERSION -m "Build $SHORT_VERSION"
+git push origin $SHORT_VERSION
```
atruskie: should do the tag before building container

src/environments/environment.ts
```diff
@@ -1,9 +1,8 @@
 import { Configuration } from "@helpers/app-initializer/app-initializer";
-import { version } from "../../package.json";
 
-export const environment: Configuration = {
+export const environment = ({
   production: false,
-  version,
-} as any;
+  version: "dev",
```
atruskie: Might add a note that this value gets templated so it doesn't get moved/formatted into a manner that sed no longer recognises.

Or:
```suggestion
  version: "<<REPLACE_WHEN_BUILT>>",
```

use a token that makes it easy to find/replace, then your sed regex can be less complex and work in more cases.

atruskie (approved):
Great work. Only documentation changes suggested.

build/build_docker.ps1
atruskie: the `beta` tag should not be included int the versions - it is not a true beta.

build/build_docker.ps1
Allcharles: No point adding a tag if the docker build fails for some reason?

Dockerfile
Allcharles: Was just following the template here: https://github.com/QutEcoacoustics/audio-analysis/blob/master/build/Dockerfile#L7-L16. No clue what `org.ecosounds.version` is, and what its relevance is to

atruskie: in which case, if they're the same, set them both the version build arg

Allcharles: The matching labels are `version` and `org.ecosounds.schema-version`. `org.ecosounds.version` is slightly different. That said, unless i know what the labels are for, it seems like almost all of them are useless and we should instead only set the following:

- `maintainer`
- `description`
- `version`
- `name`
- `vendor`
- `url`
- `ref`

build/build_docker.ps1
atruskie: true

the git metadata changes based on whether a tag is present or not

git tag first

then build image

then push tag

then push image

atruskie (approved):

Issue #1108, opened 2021-06-01
SSR should be fast to render first page


atruskie:
**Is your feature request related to a problem? Please describe.**

We use SSR to render the first page load to make the client look fast. The problem is that it makes requests to the API and often seems slower because of it.

**Describe the solution you'd like**

- All API requests done when in the SSR environment should have a very small timeout. Less than one second?
- Possibly: no API requests should be allowed at all - all should be deferred until client runs the requests
- Or we should audit and tag approved and fast requests (e.g. that don't rely on authentication or that don't change state)

**Describe alternatives you've considered**

- We could also try just not doing SSR rendering. 

**Additional context**

Allcharles:
As far as i can tell, on a slow connection SSR is still faster than waiting for the bundle to load. Using the slow 3G preset on firefox i was getting consistent SSR load times of around 3 seconds. This can be significantly faster than waiting for the bundle to load. On fast connection, the system just skips to the larger bundle so the 3 seconds are ignored. I have implemented a 1 second timeout on api requests from the SSR, and have disabled the CMS pages (instead only showing the loading animation)

atruskie:
Where these measurements before or after the changes?

atruskie:
To add some context: I got 3.61 seconds with no cache just now

atruskie:
More research into this: https://github.com/angular/universal/issues/842#issuecomment-343702098
Our route resolvers might actually be the issue.

I think you actually need to profile server side to see what the delay is. 

For context the root page of the old website loads in ~60ms

PR #1138, opened 2021-06-09
SSR adjustments

### LINKED ISSUE #1106: Handle invalid environment variables
Currently if the `apiRoot` is invalid, the website will hang and display no error message to both developers and users. Catching this error in runtime should be fairly easy to implement and will help with debugging steps.

I would also like to add some runtime checking to at least determine the `siteRoot` and `siteDir` match the location of the website. This should also help with debugging bad environment files.

Allcharles:
# Adjustments to Server Side Rendering

Attempting to speed up SSR bundles, and fix some of the SSR only errors which occur in the website.

## Changes

- ~Set 1 second timeout on SSR API requests~
- Adjusted home page to handle mobile device screens
- Set default background colour for home page when assets fail to load
- Prevent use of line-trunction on server because of https://github.com/DiZhou92/ngx-line-truncation/issues/49
- Prevent use of google maps on server
- Disabled CMS on SSR because insertion requires access to `document`
- Enhanced configuration file checking
- Removed edge tests as they should produce the same results as chrome
- Forced `git-lfs` files to download to hopefully fix assets not being available in docker container
- Updated readme with documentation on server side rendering
- Removed snazzy-info-window dependency
- Added fontawsome icon styles to `angular.json` styles to fix large icons bug when initially rendered. Example of bug shown below:
image
- Cache SSR files in express server
- Fixed bug where `ngx-loading-bar` was causing the server side renderer to wait 3 seconds before returning to the client
- Set max timeouts for HTTP requests for both website (10 seconds) and server (100ms)
- Cache HTTP requests on SSR

## Results

Tests run on latest version of Firefox, using the throttling settings from developer tools

#### Regular 3G

**Home Page**

image

**Projects Page**

image

#### Regular 3G

**Home Page**

image

**Projects Page**

image

## Issues

Addresses #1108 
Closes #1106 

### Mobile home title

image

### Normal home title

image

### Default home background

image

## Final Checklist

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -45,6 +45,7 @@ export function app(path: string): express.Express {
   server.engine(
     "html",
     // Cache results as they should not change often for unauthenticated users
+    // Cache the last hour of results, or 100 most resent requests
     withCache(
       new LRUCache({ maxAge: 10 * 60_000, maxSize: 100 }),
       ngExpressEngine({
@@ -92,7 +93,7 @@ function run(configPath: string): void {
   // Start up the Node server
   const server = app(configPath);
   server.listen(port, "0.0.0.0", () => {
-    console.log(`Node Express server listening on http://localhost:${port}`);
+    console.log(`Node Express server listening on http://0.0.0.0:${port}`);
   });
 }
 
--- a/server.ts
+++ b/server.ts
@@ -17,6 +17,8 @@ import { ngExpressEngine } from "@nguniversal/express-engine";
 import { assetRoot } from "@services/config/config.service";
 import express from "express";
 import { environment } from "src/environments/environment";
+import { LRUCache } from "@ngx-ssr/cache";
+import { withCache } from "@ngx-ssr/cache/express";
 import { AppServerModule } from "./src/main.server";
 
 // The Express app is exported so that it can be used by serverless Functions.
@@ -42,11 +44,15 @@ export function app(path: string): express.Express {
   // Our Universal express-engine (found @ https://github.com/angular/universal/tree/master/modules/express-engine)
   server.engine(
     "html",
-    ngExpressEngine({
-      bootstrap: AppServerModule,
-      // a similar provider exists in main.ts
-      providers: [apiConfig],
-    })
+    // Cache results as they should not change often for unauthenticated users
+    withCache(
+      new LRUCache({ maxAge: 10 * 60_000, maxSize: 100 }),
+      ngExpressEngine({
+        bootstrap: AppServerModule,
+        // a similar provider exists in main.ts
+        providers: [apiConfig],
+      })
+    )
   );
 
   server.set("view engine", "html");
@@ -86,7 +92,7 @@ function run(configPath: string): void {
   // Start up the Node server
   const server = app(configPath);
   server.listen(port, "0.0.0.0", () => {
-    console.log(`Node Express server listening on http://0.0.0.0:${port}`);
+    console.log(`Node Express server listening on http://localhost:${port}`);
   });
 }
 
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,9 +1,7 @@
 import { Component, OnInit, ViewEncapsulation } from "@angular/core";
 import { Title } from "@angular/platform-browser";
 import { PageInfo } from "@helpers/page/pageInfo";
-import { LoadingBarService } from "@ngx-loading-bar/core";
 import { Observable } from "rxjs";
-import { delay, map, withLatestFrom } from "rxjs/operators";
 import { withUnsubscribe } from "./helpers/unsubscribe/unsubscribe";
 import { ConfigService } from "./services/config/config.service";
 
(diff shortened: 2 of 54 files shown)


src/app/helpers/embedGoogleMaps/embedGoogleMaps.ts
```diff
@@ -11,12 +11,16 @@ let node: HTMLScriptElement;
...
  * @param key Google maps API key
  */
-export async function embedGoogleMaps(key?: string) {
+export async function embedGoogleMaps(key?: string): Promise<void> {
   let googleMapsUrl = googleMapsBaseUrl;
   if (key) {
     googleMapsUrl += "?key=" + key;
   }
 
+  if (!document) {
```
atruskie: not the isSSR function?

atruskie (approved):
The `isSSR` function seems like it could be defined... yup there we go: https://angular.io/api/common/isPlatformBrowser

Also, a question: did you try lower timeouts?  From a short profile of API requests, it seems most requests average 30 - 60 ms. Can we get our timeout lower?

It'd be good to put these timeout values in our environment.json so we can tune our servers.

If a request fails during server side rendering, will it be attempted again when the client is hydrated in the browser?

Allcharles:
The downside to using the `isPlatformBrowser()` function is that it relies on the angular dependency injection system. The function I wrote achieves the same goal without having to be run from injectable components. While it may not achieve the exact same level of functionality, it does tell me what I'm interested in, which is that `window` and `document` are undefined. Potentially I could change the name to `hasWindowAndDocument`?

atruskie:
@Allcharles if DI is really the concern here, then the issue really boils down to a passive detection method.
We're fully in control of both DI and server environment variables when doing SSR, we should be explicit about that.

"Is window defined?" is not the right question.

"Are we rendering on the server?" is the right question. In which case a better solution would defining some kind of global constant in `server.ts` that we can test againt.

But, I'd also like to see a good reason for avoiding DI? ~~You've only used it once in a non-DI context (configuration validation).~~ Nope, it is in your `isConfiguation` function which is run inside the DI context.

server.ts
```diff
@@ -86,7 +92,7 @@ function run(configPath: string): void {
   // Start up the Node server
   const server = app(configPath);
   server.listen(port, "0.0.0.0", () => {
-    console.log(`Node Express server listening on http://0.0.0.0:${port}`);
+    console.log(`Node Express server listening on http://localhost:${port}`);
```
atruskie: given we intend for this to be in a container, it should almost always be 0.0.0.0. Is there a reason it's not?

server.ts
```diff
@@ -42,11 +44,15 @@ export function app(path: string): express.Express {
...
   server.engine(
     "html",
-    ngExpressEngine({
-      bootstrap: AppServerModule,
-      // a similar provider exists in main.ts
-      providers: [apiConfig],
-    })
+    // Cache results as they should not change often for unauthenticated users
+    withCache(
+      new LRUCache({ maxAge: 10 * 60_000, maxSize: 100 }),
```
atruskie, line 50: nit: add comment about units and calculated size. I assume it is 10 minutes?

src/app/app.component.ts
```diff
@@ -26,24 +24,13 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
...
   public ngOnInit() {
     this.title.setTitle(this.config.values.brand.name);
     this.fullscreen = true;
-
-    // Delay showing loading bar
-    this.delayedProgress$ = this.loader.value$.pipe(
-      delay(3000),
-      withLatestFrom(this.loader.value$),
-      map((v) => v[1])
-    );
```
atruskie, line 46: avoiding the problem altogether, smart

src/app/components/shared/cms/cms.component.ts
```diff
@@ -41,6 +46,11 @@ export class CmsComponent extends withUnsubscribe() implements OnInit {
   public ngOnInit() {
     this.loading = true;
 
+    // Don't attempt to load CMS data if currently we are in SSR
+    if (this.isServer) {
+      return;
+    }
+
```
atruskie, line 53: with ssr request cache then we may not need to not do this. It would be better if we could serve CMS content.

What the perf difference here, with this on/off/with cache?

src/app/services/baw-api/api.interceptor.service.ts
```diff
@@ -127,6 +128,15 @@ export class BawApiInterceptor implements HttpInterceptor {
         message: response.error.meta.error.details,
         info: response.error.meta.error?.info,
       };
+    } else if (response.status === 0) {
+      // Timeout library sets status to 0 if timed out
+      // https://github.com/IKatsuba/ngx-ssr/issues/397
```
atruskie, line 133: its non-standard but conventionally cancelled requests are returned with 0

https://stackoverflow.com/a/29907525/224512
https://www.ibm.com/support/pages/what-causes-status-code-0-indicating-reqcancelled

a very detailed answer: https://stackoverflow.com/a/26451773/224512

atruskie (approved):

src/app/components/shared/cms/cms.component.ts
Allcharles: The issue with CMS is not loading data, but the actual way in which we insert the content into the page. A solution would require a rewrite of how it works

Allcharles: I have made an issue tracking this here: #1163

Issue #1184, opened 2021-07-02
Real time website themeing


Allcharles:
It would be nice to have the ability to change the theme of the website in real time. This will allow us to make the theme of the website something which can changed in the environment, changed on the server, or even changed depending on user preferences.
Currently bootstrap 4 and 5 do not support the use of CSS variables to change colours in real time, instead relying on SASS' compile time variables. Thus to implement things in a usable manner until this feature is implemented, we need to override the themes for specific components as required. An example for the primary theme button can be seen below:

```scss
@import "~bootstrap/scss/variables";
@import "~bootstrap/scss/buttons";

:root {
  --baw-primary-red: 255;
  --baw-primary-green: 255;
  --baw-primary-blue: 255;
    
  --baw-primary: rgb(
    --baw-primary-red,
    --baw-primary-green,
    --baw-primary-blue
  );
  --baw-primary-inverse: calc(
    (
      (
        (
          (var(--baw-primary-red) * 299) +
          (var(--baw-primary-green) * 587) +
          (var(--baw-primary-blue) * 114)
        ) / 1000
      ) - 128
    ) * -1000
  );
  
  --baw-primary-border: rgb(255, 255, 255);
  --baw-primary-lighter: rgb(255, 255, 255);
  --baw-primary-lightest: rgb(255, 255, 255);
  --baw-primary-darker: rgb(255, 255, 255);
  --baw-primary-darkest: rgb(255, 255, 255);
}

.btn-primary {
  @extend .btn-primary;
  background-color: var(--bs-primary, $primary);
}
```

The red, green, and blue for each theme (primary, secondary, info, etc) will be inserted by an angular service. This update should also be included with upgrading to bootstrap 5, and using the beta build of ngx-bootstrap.

PR #1215, opened 2021-07-13
Bootstrap CSS vars

Allcharles:
# Bootstrap CSS variables

Override the bootstrap styles to enable use of CSS variables on most coloured components

## Changes

- Update to bootstrap 5
- Created the following variants for each colour type (`primary`, `secondary`, `info`, etc). Using primary as an example:
  - `--baw-primary-hue`
  - `--baw-primary-saturation`
  - `--baw-primary-lightness`
  - `--baw-primary`
  - `--baw-primary-contrast`
  - `--baw-primary-lighter`
  - `--baw-primary-lighter-contrast`
  - `--baw-primary-lightest`
  - `--baw-primary-lightest-contrast`
  - `--baw-primary-darker`
  - `--baw-primary-darker-contrast`
  - `--baw-primary-darkest`
  - `--baw-primary-darkest-contrast`
- Overrode the styles for the following bootstrap elements
  - buttons
  - nav-links
  - backgrounds
  - text-colors

## Problems

- Would like to have updating of CSS variables possible from an angular service. This would allow for styles to be defined/updated at anytime during runtime or possibly even build time.
- The following colour combinations fail to pass contrast ratios:
  - `--baw-success-lighter` and `--baw-success-lighter-contrast` provides a contrast of 2.74 (needs 4.5)
  - `--baw-info-darker` and `--baw-info-darker-contrast` provides a contrast of 3.06 (needs 4.5)
  - `--baw-warning-darker` and `--baw-warning-darker-contrast` provides a contrast of 2.42 (needs 4.5)

## Issues

Relevant to #1184 

## Visual Changes

image

## Final Checklist


src/app/components/shared/header/header.component.html
```diff
@@ -1,25 +1,28 @@
-<nav class="container-fluid navbar navbar-expand-lg navbar-dark fixed-top">
+<!--
+  Note: Using [routerLink] instead of [strongRoute] so that routerLinkActive will function.
+  Otherwise uses will not have a visual clue for which page they are on
```
atruskie: sounds like there's a deficiency in strongRoute? 

File an issue?

src/app/styles/_media-mixin.scss
```diff
@@ -1,10 +1,11 @@
-@import "../../../node_modules/include-media/dist/include-media";
-@import "./variables";
+@import "~include-media/dist/include-media";
```
atruskie, line 1: it's unnerving seeing a `~` without a trailing slash. Is this correct?

src/app/styles/bootstrap/_root.scss
```diff
@@ -0,0 +1,79 @@
...
+  --baw-contrast-threshold: 50%;
+
+  @include create-variable("primary", 216, 98%, 52%);
+  @include create-variable("secondary", 208, 7%, 46%);
+  @include create-variable("success", 152, 69%, 31%);
+  @include create-variable("info", 190, 90%, 50%);
+  @include create-variable("warning", 45, 100%, 51%);
+  @include create-variable("danger", 354, 70%, 54%);
+  @include create-variable("light", 210, 17%, 98%);
+  @include create-variable("dark", 210, 11%, 15%);
```
atruskie: are these all bootstrap defaults? 

maybe we should referencing their original variables?

src/app/styles/bootstrap/components/_buttons.scss
```diff
@@ -0,0 +1,34 @@
+@mixin button($property) {
```
atruskie: ok, it's unclear to me on a casual read how much is being over-ridden here.

Are we totally replacing the bootstrap `button` mixin?

atruskie (changes requested):
Technically, a great change.

I think anywhere we override something I want to make it plainly clear what is being overridden and why. Document heavily.

I also think, for the sake of clarity, we should namespace non-overridden but custom mixins and functions. 

e.g.

- if we are replacing (overriding) a bootstrap thing then it needs to have the same name
- if we have our own mixing function, prefixed with `baw-` then we can tell easily we're not overriding something

We may roll back some theming changes, but we'll see how it looks after a deploy.

Changes required: documentation of overrides

src/app/styles/_media-mixin.scss
Allcharles: yes it is correct

src/app/styles/bootstrap/_root.scss
Allcharles: They are bootstrap defaults converted to hsl, AFAIK there is no easy way to convert the rgb values of bootstrap to hsl. I have however just discovered a potential way of handling all of the colour calculations in rgb however if we wanted to use bootstrap values as defaults. I ruled it out as a possibility as you would need to set transparency values on the colours (which would be bad on things like image backgrounds), however it appears i could use the following solution to bypass that: https://gist.github.com/tqc/2564280. Colour contrast calculations may be slightly better/worse as well

src/app/styles/bootstrap/components/_buttons.scss
Allcharles: We are not modifying the bootstrap mixin at all, this has no reference to the way bootstrap sets the button properties, and instead just overrides the end styling. Its effectively just adding styling on top of the bootstrap styling similar to the following:

```css
.btn-primary {
  // bootstrap styling
}

.btn-primary {
  // custom styling
}

File: src/app/components/shared/header/header.component.html (hunk 1 above)
Comment (Allcharles) [reply src/app/components/shared/header/header.component.html]: Tracking issue here: https://github.com/QutEcoacoustics/workbench-client/issues/1216

File: src/app/styles/bootstrap/components/_buttons.scss [hunk 5]
```diff
@@ -0,0 +1,48 @@
...
+  background-color: unset;
+  color: var(--baw-#{$property});
+}
+
+@each $theme, $default in $theme-colors {
+  .btn-#{$theme} {
+    @include baw-button($theme);
+  }
+
+  .btn-outline-#{$theme} {
```
atruskie, line 45: this is the point where the override (extension technically?) actually happens

atruskie (approved):

src/app/styles/bootstrap/components/_buttons.scss
Allcharles: correct

PR #1246, opened 2021-07-19
Fix production baw-client linking

Allcharles:
# Fix production baw-client linking

This PR changes the behaviour of the `BawClientComponent` to make it functional with the production versions of workbench-client. Previously baw-client was hosted in the build files of workbench-client, which is incompatible with the deployment environment.  This is related to the following PR: https://github.com/QutEcoacoustics/baw-private/pull/176

## Changes

- Changed url linking iframe of baw-client from assets folder, to api root

## Problems

- Some visual bugs occur during development due to the baw-client dist packages existing on a different domain. A comment has been left in the code to explain this to future developers

## Final Checklist

### DIFF
--- a/src/app/components/shared/baw-client/baw-client.component.ts
+++ b/src/app/components/shared/baw-client/baw-client.component.ts
@@ -95,9 +95,11 @@ export class BawClientComponent extends withUnsubscribe() implements OnInit {
   }
 
   protected updateUrl(url: string) {
+    const { oldClientOrigin, oldClientDir } = this.config.endpoints;
+
     // Bypass angular default security
     this.url = this.sanitizer.bypassSecurityTrustResourceUrl(
-      this.config.getBawClientUrl(url)
+      oldClientOrigin + oldClientDir + url
     );
   }
 }
--- a/src/app/components/shared/baw-client/baw-client.component.ts
+++ b/src/app/components/shared/baw-client/baw-client.component.ts
@@ -95,11 +95,11 @@ export class BawClientComponent extends withUnsubscribe() implements OnInit {
   }
 
   protected updateUrl(url: string) {
-    const { oldClientOrigin, oldClientDir } = this.config.endpoints;
+    const { oldClientOrigin, oldClientBase } = this.config.endpoints;
 
     // Bypass angular default security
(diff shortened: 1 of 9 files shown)


src/app/components/shared/baw-client/baw-client.component.ts
```diff
@@ -66,13 +71,15 @@ export class BawClientComponent extends withUnsubscribe() implements OnInit {
     }
 
     // Update iframe size every 250 milliseconds
-    interval(250)
+    interval(this.iframeUpdateIntervalMs)
```
atruskie, line 74: I noticed a bug in this with the version before this patch: if the iframe does not load the size just keeps increasing indefinitely.

Can you reproduce?

src/assets/environment.json
```diff
@@ -3,7 +3,9 @@
     "environment": "development",
     "apiRoot": "https://api.staging.ecosounds.org",
     "clientOrigin": "http://development.ecosounds.org:4200",
-    "clientDir": "/"
+    "clientDir": "/",
+    "oldClientOrigin": "https://api.staging.ecosounds.org",
+    "oldClientDir": "/listen_to/index.html#"
```
atruskie: `index.html#` is technically not a directory. Could call it `base` instead of `dir`, could also just put index.html# in the code

atruskie (approved):

src/app/components/shared/baw-client/baw-client.component.ts
Allcharles: I did know about the bug, but fixing it was kinda tricky and it only occurs if the baw client fails to load. The main issue is that if the client fails to load, i have no way of knowing. I also have no way of knowing if the previous size increase was because the iframe grew, or because the last iframe interval caused the change, Also, the amount the iframe grows by each step is browser dependent (50px in firefox, 57 for chrome/edge/opera).

Allcharles: I could make a solution, but there's not much i can think of which doesn't add a potential edge case which affects the successful deploy. Especially as this only occurs on a fault deploy, and should be fairly easy to identify.

atruskie: yeah, fine to ignore

PR #1247, opened 2021-07-21
Project permissions issues

Allcharles:
# Fixed broken project permission issues

There was an issue where users with writer/owner access were not able to access the project/site/point settings.

## Changes

- Added multiple user support to `UserBadgeComponent`
- Renamed `isProjectOwnerPredicate` to `isProjectEditorPredicate`
- Fixed permission checking on `isProjectEditorPredicate`

## Problems

- Didn't add testing for multiple users in `UserBadgeComponent` 

### Multiple users in menu

image

## Final Checklist

### DIFF
--- a/src/app/models/Project.ts
+++ b/src/app/models/Project.ts
@@ -15,13 +15,7 @@ import {
 import { assetRoot } from "@services/config/config.service";
 import { Card } from "@shared/cards/cards.component";
 import { AbstractModel } from "./AbstractModel";
-import {
-  creator,
-  deleter,
-  hasMany,
-  hasOne,
-  updater,
-} from "./AssociationDecorators";
+import { creator, deleter, hasMany, updater } from "./AssociationDecorators";
 import {
   bawCollection,
   bawDateTime,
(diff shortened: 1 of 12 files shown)


src/app/models/Project.ts
```diff
@@ -107,6 +101,13 @@ export class Project extends AbstractModel<IProject> implements IProject {
     };
   }
 
+  /**
+   * Returns true if user has the permissions to edit this model
+   */
+  public get canEdit(): boolean {
+    return [AccessLevel.owner, AccessLevel.writer].includes(this.accessLevel);
+  }
+
```
atruskie, line 110: did we set up a faked capabilities API? If so this method should fit into it.

atruskie (approved):
ensure the `isProjectOwner` predicate is not used for any other use cases. Owners do have different permissions than writers in the general case.

Why not test multiple owner badges?

src/app/models/Project.ts
Allcharles: I remember some work being done on it, but I'm unsure where and if it resulted in something we can use. I believe accessLevel was what we ended up with

PR #1268, opened 2021-07-27
Dynamic theme editor

### LINKED ISSUE #1184: Real time website themeing

Allcharles:
# Dynamic theme editor

This creates an admin tool for editing the theme in application and viewing the changes. It is also the part of a greater change allowing each instance of workbench-client to have its own theme set by the config.

## Changes

- Created `ThemeService`
- Created `ThemeTemplateComponent` page for viewing theme changes in real time
  - This page can be accessed without admin permissions, but only if given the direct route. Potentially it falls in the category of a psuedo-admin user such as designers who may wish to adjust the CMS or Theme
- Fixed excessively long re-compile times

## Problems

- It appears that an issue is occurring where SSR is not including the styles.scss styles in the initial render (only during production). This causes the website to look broken until the rest of the app is transferred. This needs to be investigated, potentially part of https://github.com/angular/universal/issues/974. This is tracked in issue #1300 

## Issues

Closes #1184 

### Light Toggle

image

### Dark Toggle

image

### Disabled Button Toggle

image

## Final Checklist


src/app/components/admin/theme-template/theme-template.component.ts
```diff
@@ -0,0 +1,111 @@
...
+  styleUrls: ["theme-template.component.scss"],
+})
+class AdminThemeTemplateComponent
+  extends PageComponent
+  implements AfterViewInit
+{
+  @ViewChildren("themeBox") private themeBoxes!: QueryList<ElementRef>;
+
+  public currentPallette: string[] = [];
+  public themes = bawThemes;
```
atruskie: I haven't finished review yet but it's unclear why we have multiple themes.

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,202 @@
+import { DOCUMENT } from "@angular/common";
+import { Inject, Injectable } from "@angular/core";
+import { IS_SERVER_PLATFORM } from "src/app/app.helper";
+
+export const bawThemes = [
+  "highlight",
```
atruskie, line 7: Okay I see the confusion. A theme is a collection of colours, a palette. 

Multiple themes would be multiple different palettes. The scope of the required change is only one theme

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,202 @@
...
+export const bawThemeVariants = [
+  "",
+  "-lighter",
+  "-lightest",
+  "-darker",
+  "-darkest",
+] as const;
+
+export type BawTheme = typeof bawThemes[number];
+export type BawThemeVariant = typeof bawThemeVariants[number];
```
atruskie: Publically, I don't think these colour variables need to be exposed as component strings.

I.e. we should never see --darker by itself.

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export interface RGB {
```
atruskie: Conventionally such a value would be called Rgb24

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,202 @@
...
+    @Inject(IS_SERVER_PLATFORM) private isServer: boolean
+  ) {}
+
+  /**
+   * Set the base colour for a theme
+   *
+   * @param theme Theme to modify
+   * @param color New base colour for theme
+   */
+  public setTheme(theme: BawTheme, color: string): void {
```
atruskie: This should be named something more like _setColorVariable_. A _setTheme_ function should update all the colours.

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  /**
+   * Update a theme by either inserting, or modifying, a style in the header
+   *
+   * @param theme Theme to update
+   * @param styles CSS styles to apply
+   */
+  private updateTheme(theme: BawTheme, styles: string): void {
```
atruskie: Annoying that a literal style element is needed. Under what condition would it be missing?

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Convert rgb hex string into an object containing the hue, saturation and
+   * lightness of the colour
+   *
+   * @param color RGB hex string to convert
+   * @returns HSL colour equivalent to rgb
+   */
+  private rgb2hsl(color: string): HSL {
```
atruskie: Really don't want to see this function in our code. External impl available?

atruskie (changes requested):
Some terminology needs work, and I'm concerned by the need of a style element. 

I thought one could update style variables directly?

src/app/services/theme/theme.service.ts
Allcharles: I'm unsure of what solution you are looking for. Often I want to use the palette by itself. I also often use it in combination with the variant (whilst not knowing which theme is being applied). Ie. I will apply `-darker` to a theme, without knowing which theme is in use.

src/app/services/theme/theme.service.ts
Allcharles: Could you potentially link to some relevant examples? I can't seem to find much on this convention, and as it will likely apply to a lot of the data I'm dealing with, it'd be nice to use standard names.

src/app/services/theme/theme.service.ts
Allcharles: I was under the impression you disliked colour libraries because of the bulk they add. I'll attempt to find a lightweight library

src/app/services/theme/theme.service.ts
atruskie: That's an interesting use case. 

In which case:  a lookup function is more useful?

`darker("primary") => "--primary-darker"`

src/app/services/theme/theme.service.ts
atruskie: I am of two minds on this, because it is not strictly true, unless we were using a UInt8 clamped array 

Anyway: web versions: https://github.com/Automattic/node-canvas#image-pixel-formats-experimental

src/app/services/theme/theme.service.ts
atruskie: as long as we have tree shaking, we should be able to import just the functions we need, without the full library

atruskie: last i checked the latest d3 libs are ultra portable and we can import singular functions

https://github.com/d3/d3-color

src/app/services/theme/theme.service.ts
Allcharles: A viable solution, however it assumes that we a) will only have `base`, `lighter`, `lightest`, etc (bootstrap may move to a 100, 200, 300, ..., 900 solution in the future), and b) will never want to generate an array of all palettes and variants (ie. the theme-template page), as I would have to make this array anyway. It seems like a more cumbersome solution

atruskie: Ok the distinct issue I have is we should not be exposing fragments of variables names from a public API.

If you want to expose groups of colours, then make a structure that reflects that.

If you want to expose variants of a colour, then make a helper that does that.

src/app/components/admin/theme-template/theme-template.component.scss
```diff
@@ -0,0 +1,21 @@
+input[type="color"] {
+  height: 38px;
```
atruskie: why `38`? I tend to always prefer units that scale with line size. i.e. `2em`

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
+import { DOCUMENT } from "@angular/common";
+import { Inject, Injectable } from "@angular/core";
+import * as Color from "color";
+
+/** List of palette options for website theming */
+export enum BawPalettes {
```
atruskie:
```suggestion
export enum BawThemeColors {
```

Symmetry with https://getbootstrap.com/docs/5.0/utilities/colors/#map

I don't believe this should be a public type, in which name spacing and specificity are not required

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+enum BawVariants {
+  base = "",
+  lighter = "-lighter",
+  lightest = "-lightest",
+  darker = "-darker",
+  darkest = "-darkest",
+}
+
+/** List of css variables associated with a variant of a palette */
+interface BawThemeVariant {
```
atruskie: why is this a type

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+  success = "success",
+  info = "info",
+  warning = "warning",
+  danger = "danger",
+  light = "light",
+  dark = "dark",
+}
+
+/** List of variants which exist for each palette */
+enum BawVariants {
```
atruskie:
```suggestion
enum BawColorVariants {
```

I don't believe this should be a public type, in which name spacing and specificity are not required

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/** List of css variables associated with a variant of a palette */
+interface BawThemeVariant {
+  color: {
+    base: string;
+    hue: string;
+    saturation: string;
+    lightness: string;
+  };
```
atruskie: Remove HSL structure. We're going to constrain colors to their string representation.

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+    function getCssVariables(palette: BawPalettes, variant: BawVariants) {
+      const colorBase = `--baw-${palette}${variant}`;
+      const contrastBase = `${colorBase}-contrast`;
+
+      return {
+        color: {
+          base: colorBase,
+          hue: `${colorBase}-hue`,
+          saturation: `${colorBase}-saturation`,
+          lightness: `${colorBase}-lightness`,
```
atruskie: this information is not needed

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+        },
+        contrast: {
+          base: contrastBase,
+          hue: `${contrastBase}-hue`,
+          saturation: `${contrastBase}-saturation`,
+          lightness: `${contrastBase}-lightness`,
```
atruskie: not needed

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Returns an object containing all of the css variables used by the website
+   * theme
+   */
+  public get themeVariables(): BawThemeVariables {
+    function getCssVariables(palette: BawPalettes, variant: BawVariants) {
+      const colorBase = `--baw-${palette}${variant}`;
+      const contrastBase = `${colorBase}-contrast`;
+
+      return {
```
atruskie: This structure could be simplified to `{color: '', contrast: ''}`.

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+      saturation = color.saturationl();
+      lightness = color.lightness();
+    } catch (e) {
+      console.warn(
+        `Invalid theme color given for ${palette} detected: ${colorString}`
+      );
+      return;
+    }
+
+    const root = this.document.documentElement;
```
atruskie: move this outside of function, it never should change after service initialization

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+@Injectable({
+  providedIn: "root",
+})
+export class ThemeService {
+  /** Tracks config changes to theme */
+  private theme: BawTheme;
+  /** Tracks all theme variables */
+  private _themeVariables: BawThemeVariables;
+  /** Track modified palettes to allow us to handle resetting styles */
+  private modifiedPalettes: Set<BawPalettes> = new Set();
```
atruskie: I don't think you need to track modified palettes. 

Just unset all base variables

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+        `Invalid theme color given for ${palette} detected: ${colorString}`
+      );
+      return;
+    }
+
+    const root = this.document.documentElement;
+    const baseColor = this.themeVariables[palette].base.color;
+    root.style.setProperty(baseColor.hue, `${hue}deg`);
+    root.style.setProperty(baseColor.saturation, `${saturation}%`);
+    root.style.setProperty(baseColor.lightness, `${lightness}%`);
```
atruskie: for simplicity I'd inline the variable construction here:

```
 root.style.setProperty(`--baw-${palette}-hue`, `${hue}deg`);
```

Amendment:  on seeing you use the variables elsewhere; I'd simplify this differently. 

Perhaps a function?

```
setCssColorProperty(name, color: Color) {
 root.style.setProperty(`--baw-${palette}-hue`, `${color.hue}deg`);
 root.style.setProperty(`--baw-${palette}-saturation`, `${saturation}%`);
 root.style.setProperty(``--baw-${palette}-lightness`, `${lightness}%`);
}
```

Then either choose `setProperty` or `removeProperty` based on whether `color` is `null`, or add a second function

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+          base: contrastBase,
+          hue: `${contrastBase}-hue`,
+          saturation: `${contrastBase}-saturation`,
+          lightness: `${contrastBase}-lightness`,
+        },
+      };
+    }
+
+    // Generate list of theme variables on first access
+    if (!this._themeVariables) {
```
atruskie: I don't think there's any reason this has to be lazy generated

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,213 @@
...
+    return this._themeVariables;
+  }
+
+  /**
+   * Set the base colour for a palette
+   *
+   * @param palette Palette to modify
+   * @param colorString New base colour for palette (accepts most standards)
+   */
+  public setPalette(palette: BawPalettes, colorString: string): void {
```
atruskie:
```suggestion
  public setPalette(palette: BawPalettes, color: string): void {
```

src/assets/environment.json
```diff
@@ -46,6 +46,10 @@
           }
         ]
       }
-    ]
+    ],
+    "theme": {
+      "highlight": "#304f3b",
+      "primary": "#2c6fbb"
```
atruskie: why do we modify primary by default?

src/app/components/admin/theme-template/theme-template.component.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.updateColorDescriptions();
+    this.resetSelection();
+  }
+
+  private updateColorDescriptions(): void {
+    this.colorBoxes.forEach((colorBox, index) => {
+      const color = rgb(
+        getComputedStyle(colorBox.nativeElement).backgroundColor
+      ).hex();
+      colorBox.nativeElement.querySelector("#color").innerText = color;
```
atruskie: this ID selector is concerning. IDs should be globally unique

src/app/components/admin/theme-template/theme-template.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  public ngOnInit(): void {
+    const variables = this.theme.themeVariables;
+
+    // Create list of theme palettes
+    this.palettes = Object.keys(BawPalettes) as BawPaletteType[];
+    for (const palette of this.palettes) {
+      this.texts.push(`text-${palette}`);
+      this.spinners.push(`text-${palette}`);
+      this.buttons.push(`btn-${palette}`, `btn-outline-${palette}`);
```
atruskie: it seems to be these just need to be part of the `ngFor`s in the view template. Redundant to pre-generate them

src/app/components/admin/theme-template/theme-template.component.ts
```diff
@@ -0,0 +1,137 @@
...
+      this.theme.resetPalette(BawPalettes[paletteKey]);
+    }
+    this.updateColorDescriptions();
+    this.resetSelection();
+  }
+
+  private updateColorDescriptions(): void {
+    this.colorBoxes.forEach((colorBox, index) => {
+      const color = rgb(
+        getComputedStyle(colorBox.nativeElement).backgroundColor
```
atruskie: `.getPropertyValue` on the `getComputedStyle` result should achieve the same thing and be semantically clearer

src/app/components/admin/theme-template/theme-template.component.html
```diff
@@ -0,0 +1,71 @@
...
+    *ngFor="let color of colors"
+    class="color-box"
+    [style.backgroundColor]="'var(' + color.color + ')'"
+    [style.color]="'var(' + color.contrast + ')'"
+  >
+    <p class="mt-3">
+      {{ color.palette | titlecase }}:{{ color.variant | titlecase }}
+    </p>
+    <p>Sample Text</p>
+    <p id="color">#000</p>
```
atruskie: ID's should be globally unique within a DOM tree. Use

src/app/components/admin/theme-template/theme-template.component.ts
atruskie: ditto below

src/app/components/admin/theme-template/theme-template.component.html
```diff
@@ -0,0 +1,71 @@
...
+        class="form-control"
+        [value]="selected.color"
+        (change)="onColorChange($event)"
+      />
+      <div class="input-group-append">
+        <button type="button" class="btn btn-primary" (click)="resetColors()">
+          Reset
+        </button>
+      </div>
+    </div>
```
atruskie: your provided screenshot was for the first batch of changes, but it looks like these two inputs are associated. Reset should have some spacing away from the color input

atruskie (changes requested):
I'm not keen on the `palette` terminology. 

I think this changeset is still overly complicated. 

There is a lot of:
- dom manipulation which is unnecessary 
- leaky abstractions
- pre-caching of values, that in the common case make things more complex
- manual updates (like `updateColorDescriptions`) that could just work with angular's normal change detection system

I also still think none of the CSS variables should be leaking out of the theme service.

Theme should expose:
- color names
- variant names

The rest of your theme page can use those values and apply classes. 
I know I suggested `themeVariables` but I did so we weren't constructing CSS variable names  in a module that really shouldn't know anything about CSS variables. However, the net result is the complexity seen in this recent changeset.

src/app/services/theme/theme.service.ts
Allcharles: This is required, not all base variables are tracked by the theme (ie. default values set in styles.scss). If we delete all palettes, it'll set those values to black.

src/assets/environment.json
Allcharles: The contrast colour we auto-generate isnt the nicest, so i picked a better colour that is very similar but handles better.

src/app/components/admin/theme-template/theme-template.component.ts
Allcharles: They do not get the same thing. `getPropetyValue` returns the css variable (ie. `"var(--baw-highlight)"`)

src/app/components/admin/theme-template/theme-template.component.html
Allcharles: Part of me thinks that i should turn this into two buttons. Reset palette, and reset theme

src/app/components/admin/theme-template/theme-template.component.ts
Allcharles: >ditto below

Is this referring to the `colors` array, or just the `texts`, `spinners`, and `buttons`

src/app/components/admin/theme-template/theme-template.component.scss
Allcharles: This is to make the input form match the height of the other input form. Bootstrap did not change the color input forms to match the styling of the rest of their forms for some reason. I have changed this to 100% which is a better approach

src/app/services/theme/theme.service.ts
atruskie: Example? If I'm wrong here I'd like to understand why.

Of the variables we could change (`--baw-{colour}-{hue|saturation|lightness}`) which are not tracked in styles.css?

Allcharles:
> I'm not keen on the `palette` terminology.
> 

Can we agree on a terminology? I previously proposed the following which you seemed to have little opposition to:

- `color`: Individual color of an element (ie. `primary-lighter`)
- `variant`: A variant of the base colour for a pallete (ie. `lighter`)
- `palette`: A collection of colours (ie. `primary`)
- `theme`: A collection of palettes

As this will be the third time I am renaming how I address things, I would like to get it done right. Its making my ability to use consistent naming across the code difficult as I have known each of the elements as various things, and I am forced to find everywhere I've named something whenever we change the convention.

> I think this changeset is still overly complicated.
> 
> There is a lot of:
> 
>     * dom manipulation which is unnecessary
> 

This has hopefully been addressed in my recent changes

>     * leaky abstractions
> 

I'm unsure what you are referring to here? Leaky in what way?

>     * pre-caching of values, that in the common case make things more complex
> 

There was no distinct method made to cache any values. In the case of `themeVariables` I have changed it to be generated on construction, and the generation of css class names in the theme-template was an attempt to make a cleaner approach to displaying multiple bootstrap components. It was fairly useful as I have had to rename and rework the variables which generate these class names multiple times. Hopefully this address the concern of creeping complexity

>     * manual updates (like `updateColorDescriptions`) that could just work with angular's normal change detection system
> 

This comment is a bit misleading at first. I suspect you are referring to how `updateColorDescriptions` sets the `innerText` of the color box. I have addressed this in my latest changes.

> 
> I also still think none of the CSS variables should be leaking out of the theme service.
> 
> Theme should expose:
> 
>     * color names
> 
>     * variant names
> 
> 
> The rest of your theme page can use those values and apply classes.

Quite frankly, to my understanding what you want is literally impossible and I suspect stems from miscommunication about what the CSS variable names are being used for. The theme-template page is a way to view the colours of these CSS variables, these variables are not directly used in any javascript throughout the entire application. It is merely a debugging/theming tool to allow admins the ability to view the full colour palette and how well the text contrast will work. The variables themselves are not used outside of the css styles, something which is somewhat unrelated to this change-set. If the theme-template page did not exist, the ThemeService would only allow inputs which set the theme. The original PR reflected this idea by having the list of colors and variants as something separate from the ThemeService.

The concern I see here is that my styles will have to reference these CSS variables (ie.  [example](https://github.com/QutEcoacoustics/workbench-client/blob/3853a7884471118de5dbdbef0019ee918ce5c781/src/app/components/home/home.component.scss#L48-L49)). Because this is effectively string typing, it is not amazing; however AFAIK there is no better way to type support my CSS variables.

Your comment indicating that i should just expose the color and variant names will only lead back to my initial approach to this feature which you disliked. I suspect a call is needed.

> I know I suggested `themeVariables` but I did so we weren't constructing CSS variable names in a module that really shouldn't know anything about CSS variables. However, the net result is the complexity seen in this recent changeset.

src/assets/environment.json
atruskie: okay, this should be the default in sass then. 

Anything in the config is essentially an override of the default theme

src/app/components/admin/theme-template/theme-template.component.ts
atruskie: `colorBoxes`

src/app/services/theme/theme.service.ts
Allcharles: True, you are correct. I could re-generate the `--baw-{color}` using its hue, saturation, and lightness values and do the following:

```typescript
const color = `--baw-${color}`;
this.style.setProperty(color , `hsl(var(${color}-hue), var(${color}-saturation), var(${color}-lightness))`);
```

Allcharles: Actually no, i was correct the first time. We are modifying the hue, saturation, and lightness values. Effectively this is what happens. The styles.scss creates a list of values:

```css
--baw-primary: blue;
--baw-secondary: grey;
```

Then, when the application starts up we set theme theme overrides:
```css
--baw-primary: lightblue;
--baw-primary: blue;
--baw-secondary: grey;
```

If we delete `--baw-primary` using `removeProperty` everything is fine because we only remove the lightblue version, and the original remains. But if we delete `--baw-secondary`, there is no backup value, and it will be left unset

atruskie: Discussed offline, but basic browser DOM tests indicate this should be possible:
image

There is a difference somewhere we don't yet understand

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  public themeVariants = themeVariants;
+
+  /** Tracks config changes to theme */
+  private theme: BawTheme;
+  /** Stored styles of the website */
+  private style: CSSStyleDeclaration;
+
+  public constructor(@Inject(DOCUMENT) private document: Document) {
+    // Read the root style of the website
+    this.style = this.document.documentElement.style;
```
atruskie: hmm, it would not have been my first instinct to store the style map over the dom element reference.

Is there a reason?

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Set the base colour for a colour
+   *
+   * @param color Colour to modify
+   * @param value New base value for colour (accepts most colour standards)
+   */
+  public setColor(color: ThemeColor, value: string): void {
```
atruskie: My personal preference:
```suggestion
  public setColor(colorName: ThemeColor, color: string): void {
```

src/app/services/theme/theme.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  providedIn: "root",
+})
+export class ThemeService {
+  /** List of colour options in theme */
+  public themeColors = themeColors;
+  /** List of variants for each colour option in theme */
+  public themeVariants = themeVariants;
+
+  /** Tracks config changes to theme */
+  private theme: BawTheme;
```
atruskie, line 41: ~~Is there a reason this service needs to be stateful?~~

OK, I see, hard `resetTheme` sets the values to the last set theme.

src/app/styles/_panels.scss
```diff
@@ -0,0 +1,22 @@
...
+  "info",
+  "warning",
+  "danger",
+  "light",
+  "dark"
+);
+
+$theme-color-variants: ("", "-lighter", "-lightest", "-darker", "-darkest");
+
+@each $color in $theme-colors {
```
atruskie, line 21: since we talked about the triviality of these classes, might be worth leaving a comment that they're only used in one place currently

atruskie (approved):
OK, this is much cleaner.

I think there does not need to be the concept of a hard reset.
`resetColor` should basically be:

```
// if there is an override in the config, use it, otherwise unset (use null)
 color = this.theme[colorName]
 this.setCssColorProperty(colorName, color);
```

Doing this would obviate the need for `customizeInstance`

src/app/services/theme/theme.service.ts
Allcharles: I never use the dom element, so everywhere in my code was being written as `root.style`. Potentially that provides more semantic information, but i found it repetitive.

src/app/services/theme/theme.service.spec.ts
```diff
@@ -1,16 +1,170 @@
...
+    expect(style.getPropertyValue(lightness)).toBe(`${readHslValue(color.l)}%`);
+  }
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
-    service = TestBed.inject(ThemeService);
+    spec = createService();
+
+    // Styles are not reset between tests, so we reset here
+    spec.service.resetTheme();
```
atruskie, line 63: might be worth resetting afterAll ?

src/app/services/theme/theme.service.ts
```diff
@@ -41,109 +42,103 @@ export class ThemeService {
...
-    this.setCssColorProperty(color, null);
+  public resetColor(colorName: ThemeColor): void {
+    this.setCssColorProperty(colorName, null);
 
-    if (!hardReset) {
-      // Re-add instance changes to color
-      this.customizeInstance(color);
-    }
+    // Re-add instance changes to color
+    this.setColor(colorName, this.theme[colorName]);
```
atruskie, line 76: `setColor` itself calls `setCssProperty`, is there any reason for the additional `setColorProperty` above?

atruskie (approved):

PR #1337, opened 2021-08-12
General Unit Test Cleanup

Allcharles:
# General Unit Test Cleanup

## Changes

- Updated test model generators to accept a partial version of the model. This is far more flexible than just editing the `id`, and reduces the amount of code which replicates this functionality
- Removed instances of the following code `new Project({...defaultProject, name: "custom name"})` and other variants. This was a potential bug in our unit tests which would trigger if the persistent values for the model changed. Ie. If a test used the above pattern, and the component depended on reading the `createdAt` date for a model, removing the `createdAt` property from the list of persistent properties for the model would cause the test to fail
- Updated some component and service tests to use spectator library
- Enabled some disabled tests
- Updated `mockActivatedRoute` to be compatible with spectator library
- Created `Errorable` type

## Final Checklist

### DIFF
--- a/src/app/components/profile/pages/sites/my-sites.component.ts
+++ b/src/app/components/profile/pages/sites/my-sites.component.ts
@@ -51,6 +51,10 @@ class MySitesComponent extends PagedTableTemplate<TableRow, Site> {
     return this.models[userKey] as User;
   }
 
+  public hasViewUrl(site: Site): boolean {
+    return site.projectIds.size > 0;
+  }
+
   public resolveHighestAccessLevel(projects: Project[]): string {
     if ((projects ?? []).length === 0) {
       return "Unknown";
--- a/src/app/helpers/advancedTypes.ts
+++ b/src/app/helpers/advancedTypes.ts
@@ -1,3 +1,5 @@
+import { ApiErrorDetails } from "@baw-api/api.interceptor.service";
+
 /**
  * Like Partial<T> but require certain properties.
  */
@@ -46,3 +48,5 @@ export interface Tuple<T extends any, L extends number> extends Array<T> {
   0: T;
   length: L;
 }
+
+export type Errorable<T> = T | ApiErrorDetails;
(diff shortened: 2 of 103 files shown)


src/app/components/profile/pages/sites/my-sites.component.ts
```diff
@@ -51,6 +51,10 @@ class MySitesComponent extends PagedTableTemplate<TableRow, Site> {
     return this.models[userKey] as User;
   }
 
+  public hasViewUrl(site: Site): boolean {
+    return site.projectIds.size > 0;
+  }
+
```
atruskie, line 57: Is this for filtering out orphan sites?

src/app/helpers/advancedTypes.ts
```diff
@@ -46,3 +48,5 @@ export interface Tuple<T extends any, L extends number> extends Array<T> {
   0: T;
   length: L;
 }
+
+export type Errorable<T> = T | ApiErrorDetails;
```
atruskie, line 52: Personally, I'd prefer to see proper monadic types (and names) used. E.g. https://github.com/sniptt-official/monads#resultt-e

It's fine for the once off case, but if this becomes a common pattern, then I'd insist.

atruskie (approved):
the shortening of `spectator` to `spec` confused me at first - From name alone I thought it was the test/spec object you were referring too.

src/app/components/profile/pages/sites/my-sites.component.ts
Allcharles: Yeah, stops us from creating links on orphaned sites. We tested for this, but i suspect the old method of generating sites was causing the test to run improperly, so I had to update the behaviour. Potentially i should add this logic on the Site model side of things

src/app/helpers/advancedTypes.ts
Allcharles: I would like to start using this in a lot of tests so i will look into your link

PR #1360, opened 2021-08-17
Implemented Statistics Page

Allcharles:
# Implemented Statistics Page

## Changes

- Added website statistics page to website
- Implemented statistics service
- Implemented statistics model

## Problems

- Deployed baw server client does not currently expose statistics route
- Baw server has a bug where statistics route breaks when logged in

### Logged in User View

image

### Guest User View

image

## Final Checklist


src/app/components/statistics/pages/statistics.component.ts
```diff
@@ -1,128 +1,123 @@
...
+            value: fileSize(summary.audioRecordingsTotalSize, { round: 2 }),
+          },
+          { index: 6, value: summary.audioRecordingsRecent },
+          {
+            index: 7,
+            value: toRelative(summary.audioRecordingsTotalDuration, {
+              largest: 2,
+            }),
+          }
+        );
```
atruskie: ok this seems brittle. 

Can I suggest either:

- a arrow func on each definition such that with an updated value you can just `groupOne.map(x => x.new(updatedValue))`
- making the definitions funcs only (should be efficient since update only triggered by observable)
- or inverting the dictionary so that the `name` is the key and you can definitely assign in this step

src/app/models/Statistics.ts
```diff
@@ -0,0 +1,105 @@
...
+    throw Error();
+  }
+}
+
+export interface IStatistics {
+  summary: StatisticsSummary | IStatisticsSummary;
+  recent: StatisticsRecent | IStatisticsRecent;
+}
+
+export class Statistics extends AbstractModel {
```
atruskie: Hmm, these models are not really descendants of our normal API structure.  It's worth thinking about if AbstractModel is the right base class.

For example, they do not have an `id` field, unlike all other models

src/app/services/baw-api/statistics/statistics.service.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+import { MockAppConfigModule } from "@services/config/configMock.module";
+import { generateStatistics } from "@test/fakes/Statistics";
+import { validateApiShow } from "@test/helpers/api-common";
+import { StatisticsService } from "./statistics.service";
+
+type Model = Statistics;
+type Params = [];
+type Service = StatisticsService;
+
+describe("UserService", function () {
```
atruskie:
```suggestion
describe("StatsService", function () {
```

src/assets/environment.json
```diff
@@ -1,7 +1,7 @@
 {
   "endpoints": {
     "environment": "development",
-    "apiRoot": "https://api.staging.ecosounds.org",
+    "apiRoot": "http://development.ecosounds.org:3000",
```
atruskie: intentional?

atruskie (approved):
Excellent. Mostly comments, the only bit that needs work is the stats extraction.

src/app/models/AbstractModel.ts
```diff
@@ -40,14 +45,7 @@ export abstract class AbstractModel<Model = Record<string, any>> {
...
 
-  /**
-   * Model ID
-   */
-  public readonly id?: Id;
-
-  /**
-   * Model Identifier
-   */
+  /** Model Identifier */
```
atruskie: `ID` is an abbreviation of identifier. I'd say this doc string is misleading. Maybe ` Model type name` ? Or something like that.

atruskie (approved):

PR #1382, opened 2021-08-21
HideProject config setting

### LINKED ISSUE #1216: Links using strongRoute directive are non-functional with routerLinkActive
When a link uses the `[strongRoute]` directive, the `routerLinkActive` property no longer functions on the link. This affects the menus which are unable to highlight themselves automatically if the current route matches the strong route. Currently this is bypassed by using `[routerLink]` instead for the header and footer components, and a custom solution for the menus. This issue should be investigated to determine if it is an error in the `StrongRouteDirective` class, or the `RouterLinkWithHref` class, and dealt with accordingly (either by fixing, or creating a bug report).

Allcharles:
# Handle project/region/site/point renaming between website instances

This creates a config setting which will show/hide the concept of projects from the website

## Changes

- Added `hideProject` setting to config
- Created sites (region models) list page
- Created active link directives for `UrlDirective` and `StrongRouteDirective` addressing (#1216)

`RouterLinkWithHref`|`RouterLinkActive`|Use Case|Example
-|-|-|-
`UrlDirective`|`UrlActiveDirective`|Allows links to routes which includes both route parameters and query paramters|`<a [bawUrl]="/sites/1?page=1" bawUrlActive="active">link</a>`
`StrongRouteDirective`|`StrongRouteActiveDirective`|Allows the use of the strong route variables to strongly type links on the website, however route and query parameters must be passed to this directive (route parameters are extracted from the existing url if they exist)|`<a [strongRoute]="strongRoute" strongRouteActive="active">link</a>`

- Updated home page to switch content based on `hideProject` setting
- Updated header to switch links between "projects" and "sites"

## Problems

- Still need to address hiding projects from the menu system (if wanted)

## Issues

Closes #1216 

### Home Page (`hideProject: true`)

image

### Home Page (`hideProject: false`)

image

### Sites Page

image

## Final Checklist


src/app/components/home/home.component.ts
```diff
@@ -62,17 +79,24 @@ class HomeComponent extends PageComponent implements OnInit {
...
       )
       .subscribe(
-        (cards) => (this.projectList = cards),
-        () => (this.projectList = List([]))
+        (cards) => {
+          this.viewMore.list = cards;
+          this.viewMore.loading = false;
+        },
+        () => {
+          this.viewMore.loading = false;
```
atruskie: do observables have something equivalent to a `finally` operator?

src/app/directives/strongRoute/strong-route-active.directive.ts
```diff
@@ -0,0 +1,49 @@
...
+} from "@angular/router";
+import { StrongRouteDirective } from "./strong-route.directive";
+
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: "a[strongRouteActive]",
+})
+export class StrongRouteActiveDirective extends RouterLinkActive {
+  @ContentChildren(StrongRouteDirective, { descendants: true })
+  public linksWithHrefs!: QueryList<StrongRouteDirective>;
```
atruskie, line 25: is the bang (`!`) significant ?

src/app/directives/strongRoute/strong-route-active.directive.ts
```diff
@@ -0,0 +1,49 @@
...
+  IsActiveMatchOptions,
+  Router,
+  RouterLink,
+  RouterLinkActive,
+} from "@angular/router";
+import { StrongRouteDirective } from "./strong-route.directive";
+
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: "a[strongRouteActive]",
```
atruskie, line 21: Are we creating another string route directive? Why?

src/app/directives/strongRoute/strong-route.directive.ts
```diff
@@ -48,10 +52,19 @@ export class StrongRouteDirective
...
+        ...this.angularRouteParams,
+        ...this.routeParams,
+      });
+    } else {
+      this.routerLink = this.strongRoute?.toRouterLink({
+        ...this.angularRouteParams,
+        ...this.routeParams,
+      });
+    }
+
```
atruskie: Ok, this seems like a major violation of separation of concerns. Why does strong route need to know about the project hide setting?

src/app/directives/url/url-active.directive.ts
```diff
@@ -0,0 +1,49 @@
...
+  RouterLink,
+  RouterLinkActive,
+} from "@angular/router";
+import { UrlDirective } from "./url.directive";
+
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: "a[bawUrlActive]",
+})
+export class UrlActiveDirective extends RouterLinkActive {
```
atruskie, line 23: I think I'm going to need a summary table of these different directives

src/app/models/Region.ts
```diff
@@ -47,6 +48,8 @@ export interface IRegion extends HasAllUsers, HasDescription {
  * A region model.
  */
 export class Region extends AbstractModel<IRegion> implements IRegion {
+  public static defaultProjectId = 1;
```
atruskie: 100% rejected

src/assets/environment.json
```diff
@@ -24,6 +24,7 @@
       "sourceRepository": "https://github.com/QutEcoacoustics/workbench-client",
       "sourceRepositoryIssues": "https://github.com/QutEcoacoustics/workbench-client/issues"
     },
+    "hideProjects": true,
```
atruskie: is there a reason for this default?

atruskie (changes requested):
Looks like a good start.

The multiple directives we have for routes are starting to get confusing. Can you present a table with their differences and use cases?

DefaultRegionId and strong route knowing about the hideProject settings are major violations of the leaky abstraction principle.

src/app/directives/strongRoute/strong-route-active.directive.ts
Allcharles: Somewhat, its typescript syntax thats required if i ever want to enable `strict` type checking for the project. It basically just means that this value will exist eventually. Otherwise in strict mode, it is treated as `QueryList | undefined`

src/app/components/home/home.component.ts
Allcharles: Not in the way you are hoping where we could have a function run when the observable is complete. Observables have the concept of a `complete` operator which is similar to the `finally` operator, but it does not run on failures. The other alternative is the `finalize` pipe, but it would require rewriting tests, and isn't a pretty solution either.

src/app/directives/strongRoute/strong-route-active.directive.ts
Allcharles: I could make a directive that combines the functionality of `StrongRouteActiveDirective` and `UrlActiveDirective` if you wanted, however I'm unsure what selector to use. I could also theoretically override the `RouterModule` with my own version of `RouterLinkActive`, however that seems dangerous and brittle

src/app/directives/strongRoute/strong-route.directive.ts
Allcharles: I am going to have to insert default project ids into strong routes somewhere in the app, AFAIK there is no alternative to this idea without remaking the menu system to allow components to have multiple menu routes (which'll take too long). This was the optimal place i could find where i can update any links to include a default project id if needed when the `hideProjects` setting is active. If you have an alternative to suggest, i would love to hear

src/app/directives/url/url-active.directive.ts
Allcharles: Updated PR description to hopefully give you what you are looking for, not sure i've achieved that goal tho

src/app/models/Region.ts
Allcharles: 2 immediate alternatives i see that would continue down the road ive set:

1. Store this value in the config sent by the server (ie. `defaultProject: 1`)
2. During bootstrap, make a request for projects if `hideProjects` is set, and extract the project id for the only project (assuming that is true)

src/assets/environment.json
Allcharles: Mostly leftover from testing, this will be changed to `false` on completion assuming i remember. Only a dev environment setting so it shouldnt matter too much

src/app/directives/strongRoute/strong-route.directive.ts
atruskie: I'm not convinced there is a problem. We should probably step through this in a call.

Essentially though:
- load all regions
- regions have a project_id
- when a region is navigated too, use that project id in the route
- everything else should be normal

You may need a new page, a shallow regions list, to kick this process off.

src/app/models/Region.ts
atruskie: plan outlined in comment above, but, yeah, option 2 essentially - except not assuming there is a single project

src/app/directives/url/url-active.directive.ts
atruskie: Add: example usage, input type, how qsps are used.

Also: why are there separate `*Active` directives?

atruskie (approved):
Excellent

PR #1473, opened 2021-09-15
Region Images

Allcharles:
# Enable Region Images

Previous implementation of region images was flawed, and this was exposed when images were finally added to the regions

## Changes

- Added image to region details page
- Updated region model to return image
- Updated `bawImage` decorator to handle relative links
- Updated `ImageDirective` to display a grey background placeholder while images are loading

## Problems

- Need to implement tests

## Visual Changes

Loading Animation:

Image Loading 2

Region Details Page

image

Project Details Page

image

## Final Checklist

### DIFF
--- a/src/app/components/projects/components/site-card/site-card.component.scss
+++ b/src/app/components/projects/components/site-card/site-card.component.scss
@@ -17,6 +17,11 @@ li.list-group-item {
   }
 }
 
+img {
+  width: 60px;
+  height: auto;
+}
+
 .nav > li > a:hover {
   text-decoration: none;
   background-color: #eee;
@@ -33,9 +38,4 @@ li.list-group-item {
   span {
     background-color: var(--baw-highlight);
   }
-}
-
-#image {
-  width: 60px;
-  height: auto;
 }
--- a/src/app/components/projects/components/site-card/site-card.component.scss
+++ b/src/app/components/projects/components/site-card/site-card.component.scss
@@ -17,16 +17,6 @@ li.list-group-item {
   }
 }
 
-.image-wrapper {
-  min-height: 60px;
-  min-width: 60px;
-
-  img {
-    max-width: 60px;
-    max-height: 60px;
-  }
-}
-
 .nav > li > a:hover {
   text-decoration: none;
   background-color: #eee;
--- a/src/app/components/projects/components/site-card/site-card.component.scss
+++ b/src/app/components/projects/components/site-card/site-card.component.scss
@@ -34,3 +34,8 @@ li.list-group-item {
     background-color: var(--baw-highlight);
   }
 }
+
+#image {
+  width: 60px;
+  height: auto;
+}
--- a/src/app/components/projects/components/site-card/site-card.component.scss
+++ b/src/app/components/projects/components/site-card/site-card.component.scss
@@ -17,9 +17,14 @@ li.list-group-item {
   }
 }
 
-img {
-  max-width: 60px;
-  height: auto;
+.image-wrapper {
+  min-height: 60px;
+  min-width: 60px;
+
+  img {
+    max-width: 60px;
+    max-height: 60px;
+  }
 }
 
 .nav > li > a:hover {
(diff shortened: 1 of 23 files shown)


src/app/components/projects/components/site-card/site-card.component.scss
```diff
@@ -17,11 +17,6 @@ li.list-group-item {
   }
 }
 
-img {
-  max-width: 60px;
-  height: auto;
-}
-
```
atruskie: why remove css definition?

src/app/directives/image/image.directive.ts
```diff
@@ -12,6 +12,12 @@ import { ImageSizes, ImageUrl } from "@interfaces/apiInterfaces";
 import { assetRoot } from "@services/config/config.service";
 import { OrderedSet } from "immutable";
 
+export const notFoundImage: ImageUrl = {
+  url: `${assetRoot}/images/404.png`,
+  height: 221,
```
atruskie: why 221?

src/app/directives/image/image.directive.ts
```diff
@@ -63,29 +73,40 @@ export class AuthenticatedImageDirective implements OnChanges {
...
+      this.imageEl.style.backgroundColor = "lightgrey";
+
+      // Clear background color when loaded
+      this.imageEl.onload = () => {
+        // Prevent overriding of 'this'
+        // Make background transparent, use a wrapper to set a different background
+        this.imageEl.style.backgroundColor = "unset";
+        // Change height to automatic sizing once it is loaded
+        this.imageEl.style.height = this.height ?? "auto";
+      };
```
atruskie: This should all be CSS

atruskie (changes requested):
ok, I really dislike setting attributes for image dimensions.

These should be really CSS styles unless there's a good reason for them not to be? We can see the DRY principle violated everywhere you have inlined those dimensions.

The direct element referencing for image load needs work. Seems like overkill to be interacting with the DOM directly. What problem were you attempting to solve?

src/app/directives/image/image.directive.ts
Allcharles: That is the height of the image in pixels

src/app/components/projects/components/site-card/site-card.component.scss
Allcharles: This was no longer needed as we tell the image directive that the image should be 60px, and it automatically sets the height to auto

src/app/directives/image/image.directive.ts
atruskie: two issues:

- the 404 image may not always be these dimensions
- why is it not 220? Why is it not square?

src/app/directives/image/image.directive.ts
```diff
@@ -63,29 +69,37 @@ export class AuthenticatedImageDirective implements OnChanges {
...
+        this.imageEl.classList.add("loaded-image");
+      };
     }
 
     // Re-enable use of current src
-    if (this.usedUrls.count() > 0) {
-      this.usedUrls = this.usedUrls.delete(this.usedUrls.last());
+    if (this.usedImages.count() > 0) {
+      this.usedImages = this.usedImages.delete(this.usedImages.last());
     }
```
atruskie: random thought; no changes required: this could just be a queue? I.e. pop off the array

src/app/models/AttributeDecorators.ts
```diff
@@ -60,25 +61,45 @@ export function bawImage<Model>(
...
     const imageBSize = b.height * b.width;
     return imageASize === imageBSize ? 0 : imageASize > imageBSize ? -1 : 1;
-  }
+  };
 
   function missingDefault(images: ImageUrl[]): boolean {
     return !images.find((image) => image.size === ImageSizes.default);
   }
 
+  function sanitizeUrl(apiRoot: string, url: string): string {
```
atruskie: nit: this is more normalization than sanitization (cleaning for malicious values)

atruskie (approved):

src/app/directives/image/image.directive.ts
Allcharles: Off the top of my head, i don't remember why i chose this approach. A queue may be a better approach

Allcharles: Had a dig and figured out the problem i originally ran into, however i figured out a bypass which should be a bit cleaner

PR #1587, opened 2021-10-18
Iframe post messages

Allcharles:
# IFrame post messages

This is part of a collection of changes which allows the baw-client to communicate when it has successfully loaded, and the height of its containing content to the parent website. Purpose of this fix is to allow the workbench-client to properly handle loading the baw-client inside an iframe even when the domains may not match.

The changes to the config have already been enabled in the following commit on baw private: https://github.com/QutEcoacoustics/baw-private/commit/c8021b3b324733ad67c1a2d95b0abd2076490f11. The sister PR to this can be located here: https://github.com/QutEcoacoustics/baw-client/pull/353

## Final Checklist

### DIFF
--- a/src/app/components/shared/baw-client/baw-client.component.ts
+++ b/src/app/components/shared/baw-client/baw-client.component.ts
@@ -51,6 +51,11 @@ export class BawClientComponent extends withUnsubscribe() implements OnInit {
 
   @HostListener("window:message", ["$event"])
   private onBawClientMessage(event: MessageEvent): void {
+    // Ignore events which are not from the old client
+    if (event.origin !== this.config.endpoints.oldClientOrigin) {
+      return;
+    }
+
     try {
       const meta: { height: number } = JSON.parse(event.data);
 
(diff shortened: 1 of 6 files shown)


src/app/components/shared/baw-client/baw-client.component.ts
```diff
@@ -39,11 +45,28 @@ export class BawClientComponent extends withUnsubscribe() implements OnInit {
...
+
+    try {
+      const meta: { height: number } = JSON.parse(event.data);
+
+      if (meta.height > 0) {
+        this.loading = false;
+        this.updateIframeSize(meta.height);
+      }
+    } catch (e: any) {
+      // Ignore message
```
atruskie, line 63: should log - never suppress totally

atruskie (approved):
Good, clean solution

src/app/components/shared/baw-client/baw-client.component.ts
Allcharles: The Baw client posts other messages which may not relate to this use case. For example, webpack posts messages about its state for some reason. I could log, but most of it should theoretically be useless

Issue #1163, opened 2021-06-21
SSR CMS Content


Allcharles:
As of #1138, CMS content will be disabled during SSR. A better solution would instead be to load the CMS content on the server, and then transfer the content using angular state. This was disabled because the current solution for inserting CMS content on the website relies on using DOM features which are not supported during SSR. This will need to be re-written using a different method which is compatible with SSR, or a polyfill will need to be added.

The benefits of this solution is that the CMS content should be loaded extremely fast on the server, and shipping the data through the angular state should be faster for the end client when compared with making another request for the data after page load. The hope is that this has no significant performance penalty on the load times for SSR.

atruskie:
https://github.com/QutEcoacoustics/workbench-client/blob/2131539cca3ee05c980787c864d4031a4380f623/src/app/components/shared/cms/cms.component.ts#L48-L60

So, the issue we had was inserting any CMS content into the client is a security concern for the browser, particularly dynamic client-side insertion of script tags. In a SSR context, we can just use `innerHTML` , or the more correct, `setProperty` function from Angular's Renderer2 since as far as the browser is concerned nothing changed since it downloaded the page.

https://angular.io/api/core/Renderer2#setProperty

PR #1631, opened 2021-11-01
Deployment cleanup

### LINKED ISSUE #1514: SSR API requests failing
Currently, all SSR API requests are failing with an Unknown Error:

HTTP Request
```
{
  request: HttpRequest {
    url: 'https://api.staging.ecosounds.org/projects/filter',
    body: { paging: [Object], filter: {} },
    reportProgress: false,
    withCredentials: true,
    responseType: 'json',
    method: 'POST',
    headers: HttpHeaders {
      normalizedNames: [Map],
      lazyUpdate: null,
      headers: [Map],
      lazyInit: null
    },
    context: HttpContext { map: Map(0) {} },
    params: HttpParams {
      updates: null,
      cloneFrom: null,
      encoder: HttpUrlEncodingCodec {},
      map: Map(0) {}
    },
    urlWithParams: 'https://api.staging.ecosounds.org/projects/filter'
  }
}
```

HTTP Response
```
{
  response: HttpErrorResponse {
    headers: HttpHeaders {
      normalizedNames: Map(0) {},
      lazyUpdate: null,
      headers: Map(0) {}
    },
    status: 0,
    statusText: 'Unknown Error',
    url: 'https://api.staging.ecosounds.org/projects/filter',
    ok: false,
    name: 'HttpErrorResponse',
    message: 'Http failure response for https://api.staging.ecosounds.org/projects/filter: 0 Unknown Error',
    error: ProgressEvent {
      type: 'error',
      target: [XMLHttpRequest],
      currentTarget: [XMLHttpRequest],
      lengthComputable: false,
      loaded: 0,
      total: 0
    }
  }
}
```

My current guess is that this is some form of CORS issue, however without debugging tools (https://github.com/angular/universal/issues/2198) isolating and fixing the bug appears to be fairly difficult

Allcharles:
# Deployment Cleanup

This is a collection of fixes for bugs which were exposed during the staging testing of workbench-client

## Changes

- Cleanup header component unit tests
- Removed pill colouring from (none) values when showing tags associated with an annotation in statistics page
- Hide WIP wrapped components in production build
- Fixed transition issue between SSR and browser where images would still have the loading css class applied after loading
- Fixed issue where SSR was unable to make API requests
- Extracted timeout values for API requests to environment file which can be more easily changed
  - Will need to check if current production timeout values for SSR are long enough to be viable
- Implemented `RehydrationInterceptorService` which will cache the requests made in SSR, and send them to the browser
- Minor changes to express server
  - Using compression middleware to compress data and reduce loading times (TODO validate this improves things)
  - Set X-Frame-Options header to SAMEORIGIN to reduce chances of click-jacking

## Problems

- Need to implement tests for `RehydrationInterceptorService`
- There is a (hopefully) dev only problem where baw-client throws an error attempting to access any of its assets because the local path does not exist

## Issues

Fixes #1514 
#1163 Should now be possible to implement properly

### Statistics Page Before

image

### Statistics Page After

image

## Final Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,12 @@ LABEL maintainer="Charles Alleman <alleman@qut.edu.au>" \
   ref=${GIT_COMMIT} \
   schema-version="1.0"
 
+USER root
+
+# Add ability to make https wget requests
+RUN apk upgrade libssl1.0 --update-cache && \
+    apk add wget ca-certificates
+
 # drop privileges
 USER node
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,8 +44,6 @@ LABEL maintainer="Charles Alleman <alleman@qut.edu.au>" \
   ref=${GIT_COMMIT} \
   schema-version="1.0"
 
-USER root
-
 # Add ability to make https wget requests
 RUN apk upgrade libssl1.0 --update-cache && \
     apk add wget ca-certificates
--- a/server.ts
+++ b/server.ts
@@ -18,6 +18,7 @@ import { ngExpressEngine } from "@nguniversal/express-engine";
 import { assetRoot } from "@services/config/config.service";
 import express from "express";
 import { environment } from "src/environments/environment";
+import * as compressionModule from "compression";
 import { AppServerModule } from "./src/main.server";
 
 // The Express app is exported so that it can be used by serverless Functions.
@@ -50,8 +51,20 @@ export function app(path: string): express.Express {
     })
   );
 
+  // Gzip static assets
+  server.use(compressionModule());
+
   server.set("view engine", "html");
   server.set("views", distFolder);
+
+  /*
+   * This allows us to reduce the chances of click-jacking by ensuring that the
+   * site cannot be embedded into another site
+   */
+  server.get("*", (_, res, next) => {
+    res.setHeader("X-Frame-Options", "SAMEORIGIN");
+    next();
+  });
 
   // special case rendering our settings file - we already have it loaded
   server.get(`${assetRoot}/environment.json`, (request, response) => {
--- a/server.ts
+++ b/server.ts
@@ -4,8 +4,8 @@
  * the application bundle downloads.
  */
 
-import "zone.js/node";
 import "reflect-metadata";
+import "zone.js/node";
 
 import { existsSync, readFileSync } from "fs";
 import { join } from "path";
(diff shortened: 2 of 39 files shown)


Dockerfile
```diff
@@ -44,6 +44,12 @@ LABEL maintainer="Charles Alleman <alleman@qut.edu.au>" \
   ref=${GIT_COMMIT} \
   schema-version="1.0"
 
+USER root
```
atruskie: this should not be necessary - you haven't dropped privileges yet

server.ts
```diff
@@ -50,9 +51,21 @@ export function app(path: string): express.Express {
     })
   );
 
+  // Gzip static assets
+  server.use(compressionModule());
```
atruskie: We use nginx for this in production. 

We should be ensuring we're using x-sendfile though to serve assets (https://www.nginx.com/resources/wiki/start/topics/examples/xsendfile/) and follow up in our deploy playbooks. For another time though.

src/app/helpers/app-initializer/app-initializer.ts
```diff
@@ -80,7 +80,16 @@ export interface Keys {
  */
 export interface Configuration {
   kind: "Configuration";
+  /** Is the current environment running in production mode */
   production: boolean;
+  /** Timeout for web requests in server side renderer */
+  ssrTimeout: number;
+  /** Timeout for web requests in browser */
+  browserTimeout: number;
```
atruskie, line 88: nit: specifically _API_ requests only

atruskie (changes requested):
I think the compression should be removed - nginx should be doing that for us.

I think the rehydration module is the wrong solution to a problem. If the problem being fixed is:

- user sees SSR data
- then loading spinner replaces page
- then data actually loads

Then the solution IMO is to get rid of the obnoxious loading spinner and let the page stream updates as the happen (ideal). Or to rephrase: if data rendered, spinner size := 10% && position := to the side somewhere. (or something similar... our loading bar on the old client was a bar that ran across the top of the page, not to obnoxious). We really only want a large spinner when the area is completely empty.


As pointed out in our discussion: for this to work properly and not force users to some kind of navigate/ refresh to see context-correct data (i.e. API query for logged in state) then the hydrate state feature is broken as is. 
We could ensure express is forwarding the cookie when SSR makes requests, but as we haven't formalised our auth process yet, it seems this is not the right time to make such a decision.

On review: I think save the hydration interceptor for another time and focus on the UI looking nice for now.

src/app/helpers/app-initializer/app-initializer.ts
Allcharles: This affects all requests which go through the angular HttpClient library, does not specifically discriminate against API requests

atruskie (approved):

PR #1648, opened 2021-11-05
Image Upload Fix

### LINKED ISSUE #608: Formly image input
Currently the custom input for formly images is disabled in all forms. This is because of a combination of issues where the image input does not work, and the API has not been updated to support images.

Allcharles:
# Image Upload Fix

Enable the ability for users to upload images for projects, regions, and sites

## Changes

- Rename `image` attribute to `images` for models (`image` was a the converted version of `imageUrls` with accurate url paths, and default values)
- Used `image` attribute for model property when uploading images
- Updated services to handle image uploading using multipart requests

## Issues

Fixes #608 

## Final Checklist


src/app/components/shared/formly/image-input.component.ts
```diff
@@ -24,11 +24,20 @@ import { asFormControl } from "./helper";
...
 export class ImageInputComponent extends FieldType {
   public asFormControl = asFormControl;
+  public readFile(): void {
+    // File input returns a list of files, grab the first file and set it as
+    // the value of this field
+    const imageFiles = this.formControl.value;
+    if (imageFiles instanceof FileList && imageFiles.length > 0) {
+      this.formControl.setValue(imageFiles.item(0));
+    }
+  }
```
atruskie: i wonder if this should throw if more than one file is present

if file `input` does support multi-select  then picking the first file would be really mis-representing the UI's intent

src/app/models/AbstractModel.ts
```diff
@@ -62,32 +74,36 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
-      } else {
-        output[attribute] = this[attribute] as any;
+  /** Determine if model will return FormData */
+  public hasFormData(opts?: ModelSerializationOptions): boolean {
+    for (const attr of this.getModelAttributes({ ...opts, formData: true })) {
+      if (isInstantiated(this[attr])) {
+        return true;
       }
-    });
+    }
```
atruskie: personally i'd express this with `some`

```
this.getModelAttributes({ ...opts, formData: true })).values().some(isInstantiated)
```

src/app/models/AbstractModel.ts
```diff
@@ -62,32 +74,36 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+
+    for (const attr of Object.keys(data)) {
+      /*
+       * Do not surround attribute name in quotes, it is not a valid input and
+       * baw-server will return a 500 response. Attributes are in the form of
+       * model_name[attribute_name]
+       */
+      const snakeCaseAttr = snakeCase(attr);
+      output.append(`${this.kind}[${snakeCaseAttr}]`, data[attr]);
+    }
```
atruskie: Something is off with the naming here. 

Technically, every field (except notes - and even that is solvable with effort) could be sent via form data. I'd expect `toFormData` to work exactly the same as `toJson`, it's just that the data would be serialised differently.

I think we want to de-generalize this a bit. 

- if we're only using `toFormData` for image uploads then it should be a mixin, not defined on `AbstractModel`
- i think we should be a little more specific in naming with intent
  - method name could be something like `getImageAsFormData`
  - `getModelData`'s option could be `images` over `formData`

Needs some more thought, but unless there are cases I haven't considered, emphasis on being more specific and less general

src/app/models/AbstractModel.ts
```diff
@@ -106,12 +122,59 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
 
   /** Get hidden model metadata */
   public getMetadata(): Meta {
-    return this[AbstractModel.metaKey];
+    return this[AbstractModel.keys.meta];
+  }
+
+  /**
+   * Converts the model and a list of keys, into an object containing each key
+   * value pair
```
atruskie: A better description might be:

```
Converts the model and its values into a simpler object representation which is suitable for serialization. Only the `keys` specified are included in the result.
```

src/app/models/AbstractModel.ts
```diff
@@ -106,12 +122,59 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+   *
+   * @param keys List of attributes to retrieve
```
atruskie:
```suggestion
   * @param keys List of attributes to extract
```

src/app/models/AbstractModel.ts
```diff
@@ -106,12 +122,59 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+   */
+  private toObject(keys: string[]): Partial<this> {
```
atruskie, line 152: Hmm: the return type `Partial<this>` is not really true is it? Several value types are stringified in the result object, which notionally makes it a different type

src/app/models/AbstractModel.ts
```diff
@@ -106,12 +122,59 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+  }
+
+  /**
+   * Retrieves a list of all attributes associated with the model which are
+   * not the injector
+   */
+  private getModelAttributes(opts?: {
+    create?: boolean;
+    update?: boolean;
+    formData?: boolean;
```
atruskie, line 176: See previous comment, but the `formData` is really not expressing the intent well. You want to extract image related fields only, so you can send those via form data.

```suggestion
    images?: boolean;
```

atruskie, line 176: personally i think these `opts` could just be parameters... there are no additional arguments needed to justify an options object

src/app/models/AbstractModel.ts
```diff
@@ -106,12 +122,59 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+  }): string[] {
+    if (opts?.create) {
+      const attribute: keyof typeof AbstractModel.keys.create = opts.formData
+        ? "formDataAttributes"
+        : "jsonAttributes";
+      return this[AbstractModel.keys.create[attribute]] ?? [];
+    } else if (opts?.update) {
+      const attribute: keyof typeof AbstractModel.keys.create = opts.formData
```
atruskie: should this not be `.update` ?

src/app/models/AbstractModel.ts
```diff
@@ -106,12 +122,59 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+   */
+  private getModelAttributes(opts?: {
+    create?: boolean;
+    update?: boolean;
+    formData?: boolean;
+  }): string[] {
+    if (opts?.create) {
+      const attribute: keyof typeof AbstractModel.keys.create = opts.formData
+        ? "formDataAttributes"
+        : "jsonAttributes";
```
atruskie: this seems wrong... again, almost all attributes can be encoded in either json or form data

src/app/models/AttributeDecorators.ts
```diff
@@ -4,37 +4,55 @@ import { Id, Ids, ImageSizes, ImageUrl } from "@interfaces/apiInterfaces";
...
-    if (!opts) {
-      return;
-    }
-    model[AbstractModel.createAttributesKey].push(key);
-    model[AbstractModel.updateAttributesKey].push(key);
+  // Ensure all storage types have an empty array
+  model[createKeys.jsonAttributes] ??= [];
+  model[updateKeys.jsonAttributes] ??= [];
+  model[createKeys.formDataAttributes] ??= [];
+  model[updateKeys.formDataAttributes] ??= [];
```
atruskie: see main review comment: but this feels untenable

src/app/models/AttributeDecorators.ts
```diff
@@ -236,18 +254,27 @@ function createDecorator<Model>(
...
+ */
+type BawPersistAttributeOptions =
+  | boolean
+  | {
+      /** Include attribute in create requests for the model */
+      create?: boolean;
+      /** Include attribute in update requests for the model */
+      update?: boolean;
+      /** Serve this value in a form data request, instead of the usual json */
+      formData?: boolean;
```
atruskie: again most values could be submitted to API as form data. 

This parameter would be better named `formDataOnly` (or see the `supprortedFormats`  suggestion from main review).

src/app/models/Project.ts
```diff
@@ -44,19 +45,21 @@ export interface IProject extends HasAllUsers, HasDescription {
...
   public readonly descriptionHtml?: Description;
   public readonly descriptionHtmlTagline?: Description;
-  public readonly imageUrls?: string;
+  public readonly imageUrls?: ImageUrl[];
   @bawImage<IProject>(`${assetRoot}/images/project/project_span4.png`, {
     key: "imageUrls",
   })
-  public readonly image: ImageUrl[];
+  public readonly images: ImageUrl[];
+  @bawPersistAttr({ create: true, update: true, formData: true })
```
atruskie: a great example where arguments with default values would make a nicer api over an opts object

src/app/services/baw-api/baw-api.service.spec.ts
```diff
@@ -46,13 +47,14 @@ class IMockModel {
 }
 
 class MockModel extends AbstractModel implements IMockModel {
+  public kind = "MockModel";
```
atruskie, line 51: i think camel case is incorrect, but if you convince me it's not then this field is inconsistent with the others

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -208,23 +217,58 @@ export class BawApiService<Model extends AbstractModel> {
...
    * @param body Request body
    */
-  protected apiUpdate(path: string, body: AbstractModel): Observable<Model> {
-    return this.httpPatch(path, body.toJSON?.({ update: true }) ?? body).pipe(
+  protected apiCreateMultipart(
+    createPath: string,
+    updatePath: (model: Model) => string,
+    body: AbstractModel
+  ) {
+    const formData = body.toFormData({ create: true });
```
atruskie: yeah this is where this naming issue really comes into it.

This method is really better named as `apiUpdateForFormDataOnlyAttributes`  since that is how it is used.

And `body.toFormData` is better named... well i don't have a better name, but it's intent is 'get only the fields that can't be updated as JSON and format a formData response`

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -208,23 +217,58 @@ export class BawApiService<Model extends AbstractModel> {
...
+      const formData = body.toFormData?.({ update: true });
+      return this.httpPatch(path, jsonData ?? body).pipe(
+        map(this.handleSingleResponse),
+        mergeMap(() => this.httpPut(path, formData, multiPartHeaders)),
+        map(this.handleSingleResponse)
+      );
+    } else {
+      return this.httpPatch(path, jsonData ?? body).pipe(
+        map(this.handleSingleResponse)
+      );
```
atruskie: this can probably be flattened into one chain

atruskie (changes requested):
As mentioned in chat: I think `ImageUrls` values should be normalised and updated and I don't think we need to alias the field to something else like `Images` (especially since the field is read-only).

I'm not fond of the naming in some of the `AbstractModel` methods. I think the correct naming depends on use cases which may not be visible in this PR, so open to discussion. However, you should split intent (get these things) out from serialization format (json/formdata) and not mix them.

An alternate design might actually be to store more metadata about attributes - rather than just the name.  i.e. `BawPersistAttr` would (rather than inserting a string into a number of different arrays) insert an object of type 

```typescript
{
  name: string,
  create: boolean,
  update: boolean,
  supportedFormats: Array<"json"|"formData">
} 
```

This second proposal would probably make it easier to get attributes for these different cases too.  `getFormDataCreateAttributes = (x) => x.create && x.supportedFormats.contains("formData") ; this.attributes.map(getFormDataCreateAttributes )`


Why did all the model `kind` fields get converted to _snake case_? The kind field is a type script only concept, and thus a client only concept, and should be in _camel case_ along with everything else.


So I think I want to see this flow readily visible in the services layer:

1. Get json serializable model
2. update/create
3. see if there are any formData only supported dirty attributes on the model
4. if yes, get the form data payload for form-data-only attributes
5. issue a second request, which will always be an update request, and which also will be multipart

This model should remove the need for `apiCreateMultipart` - `apiCreate` will just handle it when needed.

src/app/services/baw-api/baw-api.service.ts
Allcharles: Tried to, but I'm unsure of a good method to maintain type support. Potentially a const array, or maybe a function

src/app/components/shared/formly/image-input.component.ts
Allcharles: It should not be possible to select multiple files, file input is set to only accept a single file

src/app/models/AbstractModel.ts
Allcharles: Technically it still is. The values are the same format as when the data came in which is defined in the Interface for the model. Seeing as this class extends its interface, a `Partial<this>` is correct. Ie. `IProject` defines `siteIds` as a `Id[] | Set<Id>`. In `toObject` we just convert the value from `Set<Id>` back to `Id[]`.

src/app/services/baw-api/baw-api.service.ts
Allcharles: Nvm, figured out how

atruskie (approved):

Issue #984, opened 2021-04-16
Implement audio recording downloading and list views


atruskie:
**Is your feature request related to a problem? Please describe.**

We previously wanted to abstract the notion that discrete audio recordings underlaid the stream of audio that is our site.

This is not useful to people who are actually trying to work with or download audio (especially audio that is open access).

**Describe the solution you'd like**

Implement new views:

1. [x] An audio recording list table (with sorting, paging, and filtering)
1. [x] Audio recording details page
1. [x] Download audio links for original files
   - subject to licence and citation agreement
   - on
     - listen page
     - audio recording list / details views
1. [x]  A batch download page
   - with a form to select date ranges + scope to site/point
   - instructions on how to use downloaded script (install powershell, open prompt, cd to working dir, run script)
1. [x] Option to disable/enable downloads for a project on the project details page

**Describe alternatives you've considered**

We were considering distributing scripts that will let people download en-masse but we need a nice point and click solution as well.

**Additional context**

We'll need several new API features. Issues for the server need to be done a well.

atruskie:
Related to https://github.com/QutEcoacoustics/baw-server/issues/538

Allcharles:
Some notable features of the batch download page which should be achieved:

- [ ] User selectable project/region/site (using typeahead)
- [ ] Ability to multiselect the above models
- [ ] Using asciicast for displaying terminal commands
- [ ] Show preview of audio recordings which will be downloaded by this command
- [ ] Give basic summary of what will be downloaded
  - [ ] Total size
  - [ ] Number of recordings
  - [ ] Number of sites/regions/projects
  - [ ] Length of audio data
- [ ] Add a base route which does not pre-populate any data from the route params

PR #1710, opened 2021-11-30
Audio recording pages

Allcharles:
# Audio recording pages

This implements some generic list and detail pages for audio recordings. This page handles showing the audio recordings for the following scenarios:

- all recordings
- all project recordings
- all region recordings
- all site recordings
- all point recordings

## Changes

- Created list page for audio recordings
- Created details page for audio recordings
- Fixed styling for disabled links
- Fixed issue where menu items with query parameters would not be highlighted in the secondary menu on access
- Fixed secondary menu not including query parameters in links
- Updated `hasResolvedSuccessfully` to handle optional resolvers

## Problems

- Permissions shield on audio recording details page uses the site permissions instead of the audio recording permissions, this will need to be addressed in another PR
- Admin pages for audio recordings are now redundant, and should likely be removed at some point
- (TODO) Need to implement unit tests
- Site links in audio recording details page can sometimes point to a different project from the one specified in the query parameters
- Missing filtering functionality, only sorting and paging right now

## Issues

Relates to #984

### Site/Point List page

image

### All other List pages

image

### Details page

image

## Final Checklist


src/app/components/audio-recordings/audio-recording.menus.ts
```diff
@@ -0,0 +1,39 @@
...
+  parent: audioRecordingsMenuItem,
+});
+
+export const downloadAudioRecordingMenuItem = menuLink({
+  disabled: true,
+  icon: ["fas", "download"],
+  label: "Download",
+  tooltip: () => "(UNDER CONSTRUCTION) Download audio recording",
+  uri: () => "not_implemented",
+});
```
atruskie: batch download is under construction

the download link for a single file should be functional

it's not clear from the description here which this applies to - batch or single

src/app/components/audio-recordings/audio-recording.menus.ts
```diff
@@ -0,0 +1,39 @@
+import { Category, menuLink, menuRoute } from "@interfaces/menusInterfaces";
+import { StrongRoute } from "@interfaces/strongRoute";
+
+export const audioRecordingsRoute = StrongRoute.newRoot().add(
+  "audio_recordings",
+  ({ siteId, regionId, projectId }) => ({ siteId, regionId, projectId })
+);
```
atruskie: I wasn't thinking a shallow route would be needed - i think always nested under project/site/point

you may have done that in the relevant modules though so maybe ignore this comment

src/app/components/audio-recordings/pages/list/list.component.html
```diff
@@ -0,0 +1,135 @@
...
+    <ngx-datatable-column name="Recorded">
+      <ng-template let-row="row" let-value="value" ngx-datatable-cell-template>
+        <!-- Use resolver site -->
+        <span
+          *ngIf="site"
+          [ngbTooltip]="site.tzinfoTz ?? 'Unknown timezone'"
+          placement="right"
+        >
+          {{ value }}
+        </span>
```
atruskie: We'll want a time component so we can do this in a consistent way across the site

src/app/models/AudioRecording.ts
```diff
@@ -86,9 +90,33 @@ export class AudioRecording
...
+
+  public get playUrl(): string {
     return listenRecordingMenuItem.route.format({ audioRecordingId: this.id });
   }
 
+  public get downloadUrl(): string {
+    // TODO Add download url when download page built
+    throw new Error("not implemented");
+  }
+
```
atruskie: you mean this to be the view url to the download batch page right?
It should be named batchDownloadUrl then

src/app/components/audio-recordings/pages/list/list.component.html
```diff
@@ -0,0 +1,135 @@
...
+          placement="left"
+          [bawUrl]="value.playUrl"
+        >
+          <fa-icon [icon]="['fas', 'play']"></fa-icon>
+        </a>
+        <a
+          class="btn btn-sm btn-secondary mx-1"
+          ngbTooltip="(UNDER CONSTRUCTION) Download the audio recording"
+          placement="left"
+        >
```
atruskie: single file download should work

atruskie (changes requested):
UI feedback:

- for list:
   - the _download_ menu item should read something like _batch download_
   - the secondary class on the download button for each item might be confused as "disabled" - maybe just make it a normal button?
- details:
  - unknown user is fine but linking to it seems weird, and showing a negative id is weirder
      - maybe add ℹ️ with help that says "user is unavailable because you're not logged in" or "user account is deleted
   -   can we add thousands separators for large numbers (sample rate, data length bytes)
   - why is the download menu item gray coloured?
    
Ok the most important change is: the `site` (_site_ or _point_) page should have a menu action item that says "show recordings".  The route should be `/projects/1/regions/1/sites/1/audio_recordings` and `/projects/1/regions/1/sites/1/audio_recordings/1`.


The global audio recordings page can remain, but it should be hidden (not on the main toolbar).

You also seem to think the original download endpoint is not implemented - but it is, you added a link to it in the old-client last week.

src/app/components/audio-recordings/audio-recording.menus.ts
```diff
@@ -19,21 +38,57 @@ export const audioRecordingsMenuItem = menuRoute({
...
+  uri: ({ audioRecordingId }) =>
+    `/audio_recordings/${audioRecordingId}/original`,
+});
+
+export const batchDownloadAudioRecordingMenuItem = menuLink({
+  disabled: true,
+  icon: ["fas", "download"],
+  label: "Batch Download",
+  tooltip: () =>
+    "(UNDER CONSTRUCTION) Download groups of audio recordings in a single batch",
```
atruskie:
```suggestion
    "(UNDER CONSTRUCTION) Download multiple audio recordings",
```

src/app/components/audio-recordings/pages/details/details.component.ts
```diff
@@ -70,7 +72,19 @@ class DetailsComponent extends PageComponent implements OnInit {
   }
 }
 
-DetailsComponent.linkComponentToPageInfo({
+@Component({
```
atruskie, line 75: effective hack; document as temporary - maybe also add a permalink to the relevant issue so we know to come back here

src/app/components/shared/menu/link/link.component.ts
```diff
@@ -109,13 +109,8 @@ export class MenuLinkComponent implements OnChanges {
...
-
-    /*
-     * TODO This is a potential future point of failure if the website and the api
-     * do not exist on the same domain
-     *
-     * Prepend apiRoot to relative links
-     */
+    // Redirect relative routes to api
+    // ! This seems unintuitive and likely needs to be changed in the future
     this.href = uri.startsWith("/") ? this.apiRoot + uri : uri;
```
atruskie: agreed this is dodgy. We should probably ensure an url returned by an API object gets transformed before it gets to the model so that there are no relative links

src/app/interfaces/menusInterfaces.ts
```diff
@@ -132,6 +132,8 @@ export interface MenuLink extends MenuItem {
   kind: "MenuLink";
   /**
    * The URL or fragment this link points to
+   * TODO Add input for different root urls so that Href can choose where to
+   * route
```
atruskie: Remove: I don't think that's the right design - there should never be any ambiguity in a URI, as in this should be normalized by controllers/models before this point

src/app/models/AudioRecording.ts
```diff
@@ -113,7 +116,11 @@ export class AudioRecording
   }
 
   public get downloadUrl(): string {
-    // TODO Add download url when download page built
+    return downloadAudioRecordingMenuItem.uri({ audioRecordingId: this.id });
```
atruskie: seems like we should be pointing to a strongroute instance?

atruskie (approved):
I've left some comments that can be addressed in later changes.

I think there should not be a ghost-user component - it is just an error state of a user component. I.e. We should never be choosing to instantiate a ghost-user in a template, only users.

src/app/models/AudioRecording.ts
Allcharles: The original download route is an external route to the api server. StrongRoute is for internal links only. Unless im missing something

src/app/components/audio-recordings/pages/details/details.component.ts
Allcharles: Tracking issue #1711

atruskie (approved):

PR #1722, opened 2021-12-06
Audio recording Page Cleanup

Allcharles:
# Cleanup of some threads left in #1710 

This is intended to re-enable a lot of the disabled tests in #1710, as well as cleanup some of the design criticism.

## Changes

- Re-enabled abstract model related detail view tests
- Separate components for each entry point into audio recordings action
- Changed `UserLinkComponent` into `ModelLinkComponent` and made more generic
- Fixed bug where download link on audio recording list page was accidentally disabled
- Changed action item colours in audio recording list table
- Cleanup detail view and render value components/tests

## Problems

- Some tests still need to be implemented

### Audio Recording Base Page (projects/regions)

image

### Audio Recording Project Page

image

### Audio Recording Site Page

image

### Audio Recording Region Page

image

### Audio Recording Point Page

image

## Final Checklist


src/app/components/audio-recordings/pages/details/details.component.ts
```diff
@@ -72,41 +71,74 @@ class DetailsComponent extends PageComponent implements OnInit {
...
+@Component({
+  selector: "baw-audio-recording-project",
+  templateUrl: "./details.component.html",
+})
+class ProjectDetailsComponent extends DetailsComponent {}
+@Component({
+  selector: "baw-audio-recording-projects",
+  templateUrl: "./details.component.html",
+})
+class ProjectsDetailsComponent extends DetailsComponent {}
```
atruskie: Ok this is very confusing. Some points:

- we should not be seeing a mix of terms in here. We should see `projects/sites/points` or `projects/regions/sites` but not a mix of both. Probably the latter set too. The _point_ term is a UI concept only... we should basically only ever see it rendered. Imagine it like a different language was installed.

- it might help clear this up if you left a comment detailing which route these components are for

- these are all details components; I don't understand from this code why you have duplicates for different pluralities. There should never be a `ProjectDetailsComponent` and only ever a `ProjectsDetailsComponent`

- I understand the components are namespaced but the naming in this file is confusing. This is not a suggestion, because it is worse, but it would be less ambiguous with naming like `AudioRecordingDetailsForProjectsComponent`

src/app/components/audio-recordings/pages/list/list.component.html
```diff
@@ -86,26 +86,32 @@ <h1>Audio Recordings</h1>
...
+          <fa-icon [icon]="actions.play.icon"></fa-icon>
         </a>
         <a
-          class="btn btn-sm btn-primary mx-1"
-          ngbTooltip="(UNDER CONSTRUCTION) Download the audio recording"
+          class="btn btn-sm btn-default mx-1"
           placement="left"
+          [ngbTooltip]="actions.download.tooltip()"
+          [href]="
+            apiRoot + actions.download.uri({ audioRecordingId: value.id })
```
atruskie: is there a reason this isn't normalised elsewhere?

src/app/components/audio-recordings/pages/list/list.component.ts
```diff
@@ -66,6 +74,21 @@ class ListComponent
...
+    const menuRouteKey = Object.entries(audioRecordingMenuItems.list).find(
+      (value) => pageData.route === value[1].route
+    )[0];
+
+    this.actions = {
+      play: listenRecordingMenuItem,
+      details: audioRecordingMenuItems.details[menuRouteKey],
+      download: downloadAudioRecordingMenuItem,
+    };
+  }
```
atruskie: the component shouldn't be updating it's links, the links should update themselves. Is there a reason this code is not in the actions?

src/app/components/audio-recordings/pages/list/list.component.ts
```diff
@@ -114,37 +137,67 @@ interface TableRow {
...
+@Component({
+  selector: "baw-audio-recordings-project",
+  templateUrl: "./list.component.html",
+})
+class ProjectListComponent extends ListComponent {}
+@Component({
+  selector: "baw-audio-recordings-projects",
+  templateUrl: "./list.component.html",
+})
+class ProjectsListComponent extends ListComponent {}
```
atruskie: same feedback as for details components

src/app/components/audio-recordings/audio-recording.menus.ts
```diff
@@ -1,80 +1,107 @@
...
-  queryParameters
-);
+export const audioRecordingsRoutes = {
+  base: StrongRoute.newRoot().add(baseRoutePath),
+  site: makeRoute(siteMenuItem),
+  point: makeRoute(pointMenuItem),
+  region: makeRoute(regionMenuItem),
+  regions: makeRoute(shallowRegionsMenuItem),
+  project: makeRoute(projectMenuItem),
+  projects: makeRoute(projectsMenuItem),
```
atruskie: this route  is essentially the same as the base route and is redundant

atruskie: list out in comments the routes you are making here

src/app/components/audio-recordings/pages/details/details.component.ts
atruskie: OK I've worked it out.... I think... once you add some comments for the routes you expect these components to be served under it will be clearer what you intend

src/app/components/shared/detail-view/render-field/render-field.component.ts
```diff
@@ -17,18 +17,17 @@ import { DateTime, Duration } from "luxon";
...
 import { takeUntil } from "rxjs/operators";
 
-// TODO Simplify this component by splitting into sub components
 @Component({
   selector: "baw-render-field",
   template: `
     <!-- Display plain text -->
     <dl *ngIf="styling === fieldStyling.plain">
-      <p id="plain" class="m-0">{{ display }}</p>
+      <p id="plain" class="m-0" [innerText]="display"></p>
```
atruskie, line 25: is there a reason for using innerText here?

src/app/components/shared/detail-view/render-field/render-field.component.ts
```diff
@@ -44,29 +43,10 @@ import { takeUntil } from "rxjs/operators";
...
-        >
-          {{ model }}
-        </a>
-        <ng-template #noViewUrl>
-          <span id="model">{{ model }}</span>
-        </ng-template>
-      </ng-template>
+      <baw-model-link [model]="model">
+        <span id="model" [innerText]="model.toString()"></span>
+        <span id="ghost" [innerText]="model.toString()"></span>
```
atruskie, line 48: again, why innerText?

src/app/components/shared/model-link/model-link.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  selector: "baw-model-link",
+  template: `
+    <!-- Loading text -->
+    <ng-container *ngIf="model | isUnresolved; else resolved">
+      <ng-content select="#unresolved"></ng-content>
+    </ng-container>
+
+    <ng-template #resolved>
+      <!-- Normal model with view url -->
+      <a *ngIf="!isGhostUser && hasViewUrl" [bawUrl]="model.viewUrl">
```
atruskie: inappropriate variable name `isGhostUser` in the case that this now a general component

src/app/components/shared/model-link/model-link.component.ts
```diff
@@ -0,0 +1,65 @@
...
+    <ng-template #modelTemplate>
+      <ng-content select="#model"></ng-content>
+    </ng-template>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ModelLinkComponent implements OnChanges {
+  @Input() public model: AbstractModel;
+
+  public isGhostUser: boolean;
```
atruskie: remove in the general case

src/app/components/shared/model-link/model-link.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      (this.model as User).isUnknown || (this.model as User).isDeleted;
+    this.hasViewUrl = this.safelyDetermineViewUrl();
+  }
+
+  private safelyDetermineViewUrl() {
+    // Some viewUrl methods return errors
+    try {
+      return isInstantiated(this.model.viewUrl);
+    } catch (err) {
+      console.log("No viewURL");
```
atruskie: this is not such a useful console.log - add context (e.g. this.model ) or remove

src/app/models/AudioRecording.ts
```diff
@@ -94,21 +90,6 @@ export class AudioRecording
...
-    project?: IdOr<Project>,
-    region?: IdOr<Region>,
-    site?: IdOr<Site>
-  ): string {
-    return audioRecordingMenuItem.route.format(
-      { audioRecordingId: this.id },
-      { projectId: id(project), regionId: id(region), siteId: id(site) }
-    );
-  }
-
```
atruskie: seems inconsistent with other models to remove this.

Without context, I think it would be fine to route this to the base route.

atruskie (changes requested):
Ok I need a solid reason for the 14 components this creates... 

Can we start by listing out all the routes (in comments or response) and choosing better names for the components.

I think the changing of the user-link component to the model-link component shows a fundamental misunderstanding of the utility of the class you had. I think there is room for both but ultimately this boils down to premature optimisation.

To summarize: a wrapper component that has to be styles with slots - ng-containers and their content in angular parlance - is not that much more useful than a bunch of `ngIf` statements. 
The user link component was worth making because there were two error states (as opposed to one). How the error states are rendered and how the user are rendered never change, thus the component didn't need slots/ng-content to work. 

Also this is a regularly scheduled reminder that we should not have "session user" and "user" models in our app - they are one and the same. We should maybe have a `session` model which contains a user id reference, but it is not a sub-type of user.

src/app/components/shared/detail-view/render-field/render-field.component.ts
Allcharles: Using `{{ display }}` adds random whitespace to the contents which is annoying for tests

src/app/components/shared/model-link/model-link.component.ts
Allcharles: I'm unsure what a more appropriate name is? `isUnknownOrDeletedUserModel`? `isGhostUser` identifies that this is a user model of type ghost. From your other comment below, it seems you just don't want this state being tracked

src/app/components/audio-recordings/pages/list/list.component.ts
Allcharles: This is the actions inside the datatable, not the actions in the actions menu

src/app/components/audio-recordings/pages/list/list.component.html
Allcharles: Currently I'm not sure of a good long term solution for this. I have cleaned it up in my changes, but its still not ideal

src/app/components/audio-recordings/pages/list/list.component.html
```diff
@@ -34,12 +34,8 @@ <h1>Audio Recordings</h1>
...
     <ngx-datatable-column name="Uploader">
       <ng-template let-value="value" ngx-datatable-cell-template>
-        <!-- Create link to user -->
-        <baw-model-link [model]="value.uploader">
-          <span id="model">{{ value.uploader.userName }}</span>
-          <span id="ghost">Unknown User</span>
-          <span id="unresolved"><baw-loading size="sm"></baw-loading></span>
-        </baw-model-link>
+        <baw-user-link [user]="value.uploader"></baw-user-link>
+        <baw-ghost-user-hint [user]="value.uploader"></baw-ghost-user-hint>
```
atruskie: Why do you need both the user link and the ghost user link?

src/app/components/profile/pages/profile/their-profile.component.ts
```diff
@@ -60,7 +60,7 @@ class TheirProfileComponent extends MyProfileComponent implements OnInit {
     this.user = accountModel.model;
     this.updateUserProfile(this.user);
 
-    if (this.user.isDeleted || this.user.isUnknown) {
+    if (this.user.isGhost) {
```
atruskie, line 63: Does isGhost cover deleted and unknown?

atruskie (changes requested):
Looks good. Collapse the user-link and the ghost-user-link components into one and you're good

src/app/components/profile/pages/profile/their-profile.component.ts
Allcharles: Yeah

atruskie (approved):

Issue #458, opened 2020-09-11
Fullscreen  menu system


Allcharles:
Currently, fullscreen pages do not have any method of accessing the secondary/action menu. Implement a hamburger menu which is moved to the sides to allow the user to maintain access to the buttons/actions.

Allcharles:
I would like to potentially take advantage of the offcanvas feature introduced in bootstrap 5 for this feature: https://getbootstrap.com/docs/5.1/components/offcanvas/

PR #1725, opened 2021-12-09
Multi route components

### LINKED ISSUE #1711: Page components with multiple routes
Currently classes which extend `PageComponent` are only able to link a single set of page info and menu route data. This means that when a component is needed for multiple pages, it must either retrieve its information through query parameters, or be declared multiple times. This can be fixed by updating `PageComponent` to accept multiple pairs of page info and menu route values.

Allcharles:
# Updated PageComponents to handle multiple routes

This allows a single component to be utilised in multiple route locations, with multiple page infos.

## Changes

- Removed extra audio recording list components
- Removed extra audio recording detail components
- Removed point components (handled by site components)
- Updated PageComponent to handle multiple page infos and menu routes:

```typescript
@Component()
class Component { }

Component
  .linkToRouterWith(pageInfo1, menuItem1)
  .linkToRouterWith(pageInfo2, menuItem2);

export { Component }
```

## Problems

- Fullscreen value for a component is read only from the first page info found. This may be a future problem, but will be fixed by implementing #458
- Still need to update profile pages to the new system, fairly large change, will be a separate PR

## Issues

Closes #1711 

## Final Checklist


src/app/app.component.ts
```diff
@@ -42,7 +45,7 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
    * component rendered inside the router-outlet changes.
    */
   public updatePageLayout(component: any) {
-    this.fullscreen = !!(component?.constructor?.pageInfo as PageInfo)
-      ?.fullscreen;
+    const pageComponent = component?.constructor as IPageComponentStatic;
+    this.fullscreen = !!getPageInfos(pageComponent)?.[0]?.fullscreen;
```
atruskie: potential solutions:

- page infos should have a route, can we match to current route?
- also: could just use a `.some(x => x.fullscreen)`. Whenever we have multiple routes to the same component, we are showing the _same component_. So if that component is meant to be shown full screen... then it is probably meant to be full screen despite the route is attached to. To counter: I could contrive an example where full-screen-ness varies by route, but we currently don't have that need, so let's not cater for the hypothetical

src/app/helpers/page/pageComponent.ts
```diff
@@ -5,44 +5,34 @@ import { withUnsubscribe } from "../unsubscribe/unsubscribe";
...
-      throw new Error(
-        "AndMenuRoute must be called after LinkComponentToPageInfo"
-      );
-    }
-
-    this._pageInfo.setMenuRoute(this, menu);
+  public static linkToRouterWith(
+    info: IPageInfo,
+    menu: MenuRoute
+  ): typeof PageComponent {
```
atruskie: Naming discussion: would this be better named `linkToRoute` - because that's really what's happening here, we are telling the router which component to render for a route.

Also: might be worth considering if we need the entire menu item in this interface? I don't know, but in the interests on minimalism it would be great if we only accepted a strongRoute here over a `menuRoute`.

Actually, on looking at the `pageInfo.SetMenuRoute` call below, why are we even accepting another argument. Could we just pass the `pageInfo` to this method since pageInfo can store a route?

There might have been a historical reason (probably the way decorators worked) that we needed to pass the route separately from the pageinfo, but I can't think of a good reason now?

atruskie (changes requested):
Excellent work. Lots of great removal of boilerplate, a nice and tight change.

I suspect we can make the call site to the `linkWithRouter` function even simpler; the route is part of the pageInfo (each page should have it's own pageInfo, even if it is only slightly modified for components hosted at multiple routes).

src/app/helpers/page/pageComponent.ts
Allcharles: Looking through the code, we do actually use the information stored in the menu route for the secondary menu. Otherwise the other changes should be feasible and i will attempt to do so

src/app/components/home/home.module.ts
```diff
@@ -8,6 +8,8 @@ import { homeRoute } from "./home.menus";
 const components = [HomeComponent];
 const routes = homeRoute.compileRoutes(getRouteConfigForPage);
 
+console.log(routes);
+
```
atruskie:
```suggestion
```

src/app/helpers/page/pageInfo.ts
```diff
@@ -41,17 +42,12 @@ export class PageInfo implements IPageInfo {
...
 
-  /**
-   * Set Menu Route for PageInfo. This also modifies the menu route to include
-   * the associated target component.
-   */
-  public setMenuRoute(target: Type<PageComponent>, menu: MenuRoute) {
-    this.pageRoute = menu;
+  public setComponent(target: Type<PageComponent>): void {
     this.component = target;
     this.route.pageComponent = target;
```
atruskie, line 56: might be good to add some validation - `menuRoute` is optional in the interface which means it could be omitted. That would be an error?

src/app/helpers/page/pageInfo.ts
```diff
@@ -29,7 +30,7 @@ export interface IPageInfo extends Data {
  * Page info class
  */
 export class PageInfo implements IPageInfo {
-  public pageRoute: MenuRoute & { pageComponent?: Type<PageComponent> };
+  public menuRoute: MenuRoute;
```
atruskie: I'm going to suggest this name change is not good.

The point of this field is to store the route for the page. `pageRoute` was descriptive of that function.

I'm still skeptical we even need the whole `MenuRoute` type here. Where else do we depend on the information in `pageInfo.menuRoute` except for getting the route in `get route()` below?

atruskie (changes requested):

src/app/helpers/page/pageInfo.ts
atruskie: Validation in the caller (`linkToRoute`) might also be acceptable

src/app/helpers/page/pageInfo.ts
Allcharles: The `SecondaryMenuComponent` makes use of the `MenuRoute` stored in the pageInfo. This is required because it needs to know what the current routes `MenuRoute` is, and all parent `MenuRoute`s, so that it can add them to the menu. Here is the relevant code: https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/shared/secondary-menu/secondary-menu.component.ts#L54-L79. `ActionMenuComponent` doesn't need this because we provide the list of menu items in the `menu` value.

atruskie: Ok; it still seems weird but I won't block on that. Maybe in the future we should consider how we've coupled the menu and route. Perhaps it is best that `route` and `menu` are stored as separate fields, even if they contain overlapping information.

I think the reasoning for naming holds though. Agree?

Allcharles: I don't mine changing to pageRoute

atruskie (approved):

PR #1732, opened 2021-12-16
toHaveComputedStyle custom matcher

Allcharles:
# toHaveComputedStyle custom matcher

This adds a custom jasmine matcher for computed styles in a html element

## Final Checklist

### DIFF
--- a/src/app/test/matchers/toHaveComputedStyle.ts
+++ b/src/app/test/matchers/toHaveComputedStyle.ts
@@ -1,0 +1,76 @@
+import CustomMatcherFactories = jasmine.CustomMatcherFactories;
+import CustomMatcher = jasmine.CustomMatcher;
+import CustomMatcherResult = jasmine.CustomMatcherResult;
+
+/**
+ * Custom matchers use a combination of the following resources:
+ * https://stackoverflow.com/questions/42956195/create-custom-jasmine-matcher-using-typescript
+ * https://jasmine.github.io/tutorials/custom_matcher
+ *
+ * Remember to update matcher-types.d.ts when making changes
+ */
+
+type StyleValue = string | number;
+
+function validateStyles(
+  actual: HTMLElement,
+  expected: Partial<CSSStyleDeclaration>,
+  callback: (
+    key: string,
+    actualStyle: StyleValue,
+    expectedStyle: StyleValue
+  ) => CustomMatcherResult
+): CustomMatcherResult {
+  const computedStyle = getComputedStyle(actual);
+
+  /*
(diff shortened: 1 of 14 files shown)


src/app/test/matchers/toHaveComputedStyle.ts
```diff
@@ -0,0 +1,76 @@
...
+function validateStyles(
+  actual: HTMLElement,
+  expected: Partial<CSSStyleDeclaration>,
+  callback: (
+    key: string,
+    actualStyle: StyleValue,
+    expectedStyle: StyleValue
+  ) => CustomMatcherResult
+): CustomMatcherResult {
+  const computedStyle = getComputedStyle(actual);
```
atruskie, line 44: silly question: where does this function become available? Is it not imported? Are we pulling it off a global scope?

atruskie, line 44: I'm not sure how much and what checking is done by getComputedStyle.

Assuming it is is the browser default; then we probably want to make some assertions about the `actual` target here, like not null, is a DOM Element, is attached, etc..

src/app/test/matchers/toHaveComputedStyle.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /*
+   * For each expected value, compare to computed value, and return if
+   * miss match detected
+   */
+  Object.entries(expected).forEach(([key, value]) => {
+    const computedValue = computedStyle[key];
+    const result = callback(key, computedValue, value as StyleValue);
+    if (!result.pass) {
+      return result;
```
atruskie: Another question: does this not just return the lambda? 

Does this matcher ever fail?

atruskie: Also: 

We either want to return a set of aggregate assertions (more useful IMO)

Or stop running all assertions after the first failure. Currently all assertions are run no matter the fail state which makes the if condition useless.

atruskie (changes requested):
Nice addition. 

I suspect (may be wrong) that your logic for assertions might be wrong - it is essentially only ever returning `{pass: true}`.

src/app/test/matchers/toHaveComputedStyle.ts
Allcharles: Made available here: https://github.com/QutEcoacoustics/workbench-client/blob/bdeeb1863ff4ca2e2efd373238baa76be92596c1/src/test.ts#L17-L21

atruskie (approved):

Issue #346, opened 2017-02-08
Clear up ambiguities about the timezone field in the website


atruskie:
For sites and users, advanced users get confused by the difference in timezone offset used for recordings and the timezone field used by the website.

Clear it up.

- small info tip
- potential help topic

Allcharles:
Related to #345

PR #1736, opened 2021-12-17
Timezone component

### LINKED ISSUE #345: Systematic check of dates & times, ensuring timezone formatting
Time zones must not be browser dependent, and instead constant to when the recordings were taken, so that people from around the world can evaluate audio data according to when it was recorded
- The challenge is displaying time in a format that is intuitive to users
- Most UI avoid showing times in other timezones at **all** costs.
- There are a few screens where times in multiple timezones may need to be displayed
- More feedback appreciated on this issue in particular from other designers/researchers how they show multiple timezones in a user interface!

Related to https://github.com/QutBioacoustics/baw-server/issues/279

Allcharles:
# Timezone Component

This wraps the logic of connecting a timezone to its relavent site, extracting the original timezone, and converting the value to show the time relative to the site. This also adds a on hover tooltip showing which timezone the value belongs to.

## Changes

- Updated render-view to utilize timezone component
- Updated audio recording list and detail views to utilize timezone component

## Problems

- Broken tests (awaiting architecture approval)
- Update everywhere in the app which makes reference to timezones

## Issues

Relates to #346
Fixes #345

## Visual Changes

image

## Final Checklist


src/app/components/shared/timezone/timezone.component.ts
```diff
@@ -0,0 +1,58 @@
...
+import { DateTimeTimezone } from "@interfaces/apiInterfaces";
+import {
+  AbstractModel,
+  isUnresolvedModel,
+  UnresolvedModel,
+} from "@models/AbstractModel";
+import { Site } from "@models/Site";
+
+@Component({
+  selector: "baw-timezone",
```
atruskie: I think the name is inappropriate. We are showing a date here. A date is either local or not (has a timezone) but we're still showing the date.

`<baw-date>` would be my preference. I think `<baw-time>`  is too narrow and `<baw-datetime>` is unnecessarilly verbose.

src/app/components/shared/timezone/timezone.component.ts
```diff
@@ -0,0 +1,58 @@
...
+      span {
+        text-decoration: underline dotted;
+        cursor: help;
+        text-decoration-skip-ink: none;
+      }
+    `,
+  ],
+})
+export class TimezoneComponent implements OnChanges {
+  @Input() public dateTime: DateTimeTimezone;
```
atruskie, line 28: I think this should just be called `value` - like any other input we'd see. The name of the component already explains what it does.

src/app/components/shared/timezone/timezone.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Input() public site: UnresolvedModel | Site;
```
atruskie, line 29: I think this should be called `timezone` and should accept a timezone

Given how often the `site` model will be providing the timezone I'd be happy if we accepted a model as well, or as another property, but perhaps it should be decoupled and duck typed: using typescript we can accept any model with a property called `tzInfoTz`

src/app/components/shared/timezone/timezone.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Input() public dateFormat: (date: DateTimeTimezone) => string = (date) =>
```
atruskie, line 30: Good escape hatch thinking.

I'm wondering though if we can just accept one of a few predefined types? e.g. `datetime`, `date`, `time`.
That way we'd be sure times are only shown in our formats, but we can still change format for suitability to the current use of the component

src/app/components/shared/timezone/timezone.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  public timezone: string;
+
+  public ngOnChanges(): void {
+    if (!this.dateTime) {
+      this.formattedDate = "Unknown";
+      return;
+    }
+
+    if (isUnresolvedModel(this.site)) {
+      this.timezone = "loading, hover over text again to refresh";
```
atruskie:
```suggestion
      this.timezone = "loading...";
```

atruskie (changes requested):
Ok, there are two types of times we show on this site:

- (A) the time that something in the real world happened, at a location, we should show in the local time of that event
   - an audio recording was recorded at some date, at some place, at 18:00 AEDT
   - an audio event occurred at 03:30 Burma time
- (B) the time that a digital record on the site was modified
   - a user logged in
   - a project /any other stamped model was created/updated/deleted
   
For every type A event on the site, we want to show the date in the timezone of real word location where it happened.

For every type B event, they should be shown in the user's current local date time.

For both types, they should be shown in the human friendly RFC3399 format (space delimited date/time components).

For both types they should have variable precision displays and should show the full time on hover/tooltip.

For both types they should show the time zone name and UTC offset on hover/tooltip.

However, they should be visibly distinguishable from each other. I was thinking type A dates might have a different colored border, or an icon indicating their difference.
Maybe fa-globe, or fa-clock, or both overlaid to indicate time/place difference.

Given this spec, I think a second draft of this PR is in order.

You've got a great start going here given you had a crack at this before receiving a spec for it.

atruskie:
Make sure we update audio recording column in audio recording list view to use this component as well. 

Related issues: https://github.com/QutEcoacoustics/workbench-client/issues/1815

atruskie:
Make sure we update usage in upload list page as well. See #1736

PR #1766, opened 2022-01-11
Fullscreen Layout Redesign

### LINKED ISSUE #458: Fullscreen  menu system

Allcharles:
# Fullscreen layout redesign

This PR updates the fullscreen layouts to give users access to the secondary and action menus.

## Changes

- Added menu toggle to full-screen page headers
- Added side-nav component
- Extracted secondary and action menus from router
- Extracted header links to `PrimaryMenuComponent`
- Updated styling of `CardsComponent`

## Problems

- Need to complete tests after architecture approval

## Issues

Closes #458 

### Menu System

Menu System

### Card resizing

Cards

## Final Checklist


src/app/app.component.html
```diff
@@ -1,24 +1,45 @@
...
-    </div>
+  <div
+    id="container"
+    [ngClass]="{ 'container-fluid': fullscreen, container: !fullscreen }"
+    [class.fullscreen]="fullscreen"
+    [class.menu-layout]="!fullscreen"
+  >
+    <ng-container *ngIf="fullscreen; else menuLayout">
+      <div id="menu">
+        <baw-off-canvas>
```
atruskie: given `canvas` is an actual concept (dom element, JS API etc...) i'd suggest this name is bad

Maybe `baw-modal` or `baw-dynamic-menu`? I'll look at the actual component definition later and maybe come up with a better name

src/app/app.component.html
```diff
@@ -1,24 +1,45 @@
...
-      <router-outlet name="secondary"></router-outlet>
-    </div>
+  <div
+    id="container"
+    [ngClass]="{ 'container-fluid': fullscreen, container: !fullscreen }"
+    [class.fullscreen]="fullscreen"
+    [class.menu-layout]="!fullscreen"
+  >
+    <ng-container *ngIf="fullscreen; else menuLayout">
+      <div id="menu">
```
atruskie: should this element be a `nav`?

atruskie: additionally this seems like a redundant wrapper container, why not give baw-off-canvas the id and remove the wrapper?

src/app/app.component.scss
```diff
@@ -10,105 +10,72 @@
...
  * takes up any room left by header and footer.
  */
+
 #app {
   display: flex;
   flex-direction: column;
   min-height: 100vh;
 
-  // Fullscreen Layout
-  > .container-fluid {
```
atruskie, line 19: this class is removed but still referred to in the template

src/app/components/shared/header/header.component.html
```diff
@@ -1,41 +1,54 @@
...
-  <div class="container">
-    <!-- Brand Logo -->
-    <a
-      class="navbar-brand"
-      strongRouteActive="active"
-      [strongRoute]="routes.home.route"
-      [innerText]="title"
-    ></a>
+<nav class="fixed-top">
+  <div id="navbar" class="navbar navbar-expand-lg navbar-dark">
```
atruskie: again I'm wondering if we can collapse useless wrapper divs

src/app/components/shared/header/header.component.html
```diff
@@ -1,41 +1,54 @@
...
+        <ul class="navbar-nav me-auto">
+          <ng-container *ngFor="let header of headers">
+            <baw-header-dropdown
+              *ngIf="!isNavigableMenuItem(header)"
+              [links]="header"
+            ></baw-header-dropdown>
+            <baw-header-item
+              *ngIf="isNavigableMenuItem(header)"
+              [link]="header"
+            ></baw-header-item>
```
atruskie: not a comment strictly related to this changeset, but these components seem overly specific

src/app/components/shared/off-canvas/off-canvas.component.ts
```diff
@@ -0,0 +1,51 @@
+import { ChangeDetectorRef, Component, Input, OnInit } from "@angular/core";
+import { MenuType, LeftOrRight } from "@helpers/generalTypes";
+import { withUnsubscribe } from "@helpers/unsubscribe/unsubscribe";
+import { MenuService } from "@services/menu/menu.service";
+import { takeUntil } from "rxjs";
+
+@Component({
+  selector: "baw-off-canvas",
```
atruskie: still think we can have a better name. `baw-menu-container`?

src/app/interfaces/menusInterfaces.ts
```diff
@@ -199,12 +195,7 @@ export function menuAction<T extends Omit<MenuAction, "kind">>(
  * Any Menu Item discriminated union. This is used to store any of the
  *  menu item types together in a collection.
  */
-export type AnyMenuItem =
-  | MenuAction
-  | MenuLink
-  | MenuRoute
-  | MenuModalWithoutAction
-  | MenuModal;
```
atruskie, line 207: Hmm  why was `MenuModal` removed from being a valid menu item?

atruskie (approved):
Looks fine. Small suggestions made but I think the only real feedback I can give is look for ways to simplify and reduce. 

- remove wrapper divs unless needed
- remove components that are only used once

etc...

src/app/app.component.html
Allcharles: `baw-side-nav`?

src/app/app.component.scss
Allcharles: container-fluid is a bootstrap class

src/app/interfaces/menusInterfaces.ts
Allcharles: There is no current scenario where i need a type that allows both `MenuModalWithoutAction` and `MenuModal`. Often i need a type which combines `MenuLink|MenuRoute|MenuAction` with either `MenuModalWithoutAction` or `MenuModal`. Potentially i need a better naming scheme. `AnyMenuItem (MenuLink|MenuRoute|MenuAction|MenuModal)`/`AnyMenuItemWithoutModalAction(MenuLink|MenuRoute|MenuAction|MenuModalWithoutAction)`? Currently, I use `AnyMenuItem | MenuModalWithoutAction` or `AnyMenuItem | MenuModal` where needed

src/app/app.component.ts
```diff
@@ -27,11 +28,15 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
   public fullscreen: boolean;
   public delayedProgress$: Observable<number>;
 
-  public constructor(private config: ConfigService, private title: Title) {
+  public constructor(
+    public menu: MenuService,
```
atruskie, line 34: is this unused?

src/app/components/shared/menu/action-menu/action-menu.component.ts
```diff
@@ -0,0 +1,46 @@
...
+import { withUnsubscribe } from "@helpers/unsubscribe/unsubscribe";
+import { AnyMenuItem, LabelAndIcon } from "@interfaces/menusInterfaces";
+import { MenuModalWithoutAction, WidgetMenuItem } from "@menu/widgetItem";
+import { MenuService } from "@services/menu/menu.service";
+import { Set } from "immutable";
+import { takeUntil } from "rxjs";
+
+/**
+ * Action Menu Component.
+ * A menu on the right side of the page displaying actions available
```
atruskie:
```suggestion
 * A menu on located close the current subject of the page, displaying relevant actions available
```

src/app/components/shared/menu/link/link.component.ts
```diff
@@ -93,6 +107,21 @@ export class MenuLinkComponent implements OnChanges {
...
+     */
+    this.routeData.activatedRoute$
+      .pipe(
+        map((route): ActivatedRouteSnapshot => route.snapshot),
+        takeUntil(this.unsubscribe)
+      )
+      .subscribe(({ queryParams, params }) => {
+        this.queryParams = queryParams;
+        this.routeParams = params;
+      });
```
atruskie: seems like something `SharedActivatedRouteDataService` should be doing?

src/app/helpers/page/pageComponent.ts
```diff
@@ -29,6 +32,14 @@ export class PageComponent extends withUnsubscribe() implements IPageComponent {
...
     return this;
   }
+
+  public constructor(
+    public routeData: SharedActivatedRouteDataService,
+    public route: ActivatedRoute
+  ) {
+    super();
+    this.routeData.setActivatedRoute(route);
+  }
```
atruskie: yeah see my main comments - this is all totally unrelated the components

atruskie (commented):
This is probably an asinine thing to mention now, but `fullscreen` is a bit misleading. `fullwidth` layout might be more accurate? Take it or leave it.

Ok, I realise we chose the injection pattern so we wouldn't be filtering router events... but the amount of boilerplate injected in classes now is becoming excessive.

Do we have any alternatives? Are they worth it?

- observing and filtering router events could mean `SharedActivatedRouteDataService` is set its value once in one way
- or going the other way: https://devblogs.microsoft.com/premier-developer/angular-how-to-simplify-components-with-typescript-inheritance/ (using injector to resolve dependencies indirectly)
    - I think this is a bad choice since it completely break the notion of DI

I was indecisive about my above comments, but for every component that I looked at (at least as well as my memory serves me) the `routeData` and `route` injections are not used by the component. They're orthogonal code-wise, even if strictly speaking these things are related to each other via our concept of a page. It ends up just being a lot of noise and ceremony.

src/app/interfaces/menusInterfaces.ts
atruskie: I guess I just see a menu as a collection of menu items... I'd see no reason at all to ever be more specific about the types.

I don't understand the mutual exclusion for the two modal types. Explain to me on chat? Why can't we have a menu that has multiple modals, of both types?

Allcharles:
Currently idea for keeping track of the router state of the page component is to hook into the `activate` event on the router outlet (https://angular.io/api/router/RouterOutlet). This is already being done by the `AppComponent` (https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/app.component.ts#L47-L54), and should provide a simple way of getting the data we want. Something to merge into this in a separate PR would be the ability for the app component to hide the page component if an error is detected, and vise versa. This will remove the need to hide components when route resolution fails

src/app/components/shared/menu/link/link.component.ts
```diff
@@ -93,6 +107,21 @@ export class MenuLinkComponent implements OnChanges {
...
     }
+
+    /*
+     * Components outside of router-outlet are unable to read the query/route
+     * params of the page component. So we use this bypass, check the service
+     * for more details
+     */
+    this.sharedRoute.activatedRoute$
+      .pipe(
+        map((route): ActivatedRouteSnapshot => route.snapshot),
```
atruskie: i still think this can be packaged into `SharedActivatedRouteService`.

plus a name would make the intent a little more discernable... `this.sharedRoute.getSnapshot$`

src/app/services/shared-activated-route/shared-activated-route.service.ts
```diff
@@ -0,0 +1,131 @@
...
+    this.setActivatedRoute(route);
+    this.observeRouterEvents(router, route);
+  }
+
+  /** Triggers whenever activated route for the page changes */
+  public get activatedRoute$(): Observable<ActivatedRoute> {
+    return this._activatedRoute$;
+  }
+
+  public get data(): Observable<Data> {
```
atruskie, line 85:
```suggestion
  public get routeData(): Observable<Data> {
```
Maybe? Context seems a little unclear to me. Or just add some documentation?

src/app/services/shared-activated-route/shared-activated-route.service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+      // If page component found, return activated route
+      if ((_route.component as any)?.prototype instanceof PageComponent) {
+        return _route;
+      }
+
+      // Hit max depth, return undefined
+      if (depth === 5) {
+        return undefined;
+      }
```
atruskie: Why is this needed?

Arbitrary depths in traversals are a huge code smell for me

src/app/services/shared-activated-route/shared-activated-route.service.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  public get queryParams(): Observable<Params> {
+    return this._activatedRoute$.pipe(
+      map((route): Params => route.queryParams)
+    );
+  }
+
+  private observeRouterEvents(router: Router, route: ActivatedRoute): void {
+    const findPageComponent = (
```
atruskie: for clarity, i'd suggest moving this function outside of the current one

src/app/services/shared-activated-route/shared-activated-route.service.ts
```diff
@@ -0,0 +1,131 @@
...
+      }
+
+      // Recursively search children
+      //  if page component found
+      //    return activated route
+      //  else return undefined
+      return _route.children
+        .map((child) => findPageComponent(child, depth + 1))
+        .filter((child) => child instanceof ActivatedRoute)
+        .pop();
```
atruskie: hmmm, `.pop()` mutates the array, maybe `slice(-1)` is a better intent of what you want

I realise that .map and .filter always return new arrays but pop implies you want to mutate the array and keep the state

atruskie (changes requested):
Ok this is super close. 

I have some concerns about your recursive event filtering.

I think you want something that looks like this:
```
   router.events
      .pipe(
        filter((e): boolean => e instanceof NavigationEnd),
        takeUntil(this.unsubscribe)
        concatAll(expandChildren)
        filter(isPrimaryOutlet)
        first(isPageComponent)
      ).subscribe(...)
```

Also what is the reason for the arbitrary depth traversal limit?

atruskie (approved):
I'm a little concerned those observables will be re-created each time the properties are fetched, potentially leading to an explosion of observables after a few navigation steps. Maybe initialize them in the constructor for the service? Depends of how they're disposed of later on.

otherwise it's looking really nice.

PR #1775, opened 2022-01-24
Bulk download

Allcharles:
# Batch Downloading of Audio Recording Files

This adds a page which downloads the batch downloading script, as well as giving some instructions to the user.

## Changes

- Created batch download page
- Added model change detection to `FormComponent`
- Updated differ workflow to show ssr build files

## Problems

- TODO Test components
- Page could be taken further with the following improvements:
  - Preview of audio recordings which will be downloaded
  - Extract instructions to CMS?
  - Create a base page which allows users to select specific projects/regions/sites (multi-select)
  - Better support for timezones may be possible by trying to guess the timezone the user wants to sort by, with the ability to select relevant timezone if multiple are selected?

## Issues

Addresses #984 

### Site Page

image

### Site and Region Page

image

### Region Page

image

### Project Page

image

## Final Checklist


src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+import { DateTime } from "luxon";
+
+const projectKey = "project";
+const regionKey = "region";
+const siteKey = "site";
+
+interface Model {
+  projectId?: Id;
+  regionId?: Id;
+  siteId?: Id;
```
atruskie: you might find it prudent to make these arrays because that's what they could be in the future

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+  recordingStartedBefore?: string;
```
atruskie: why before? should this not be after?

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+  recordingFinishedBefore?: string;
```
atruskie: why not date types?

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+}
+
+@Component({
+  selector: "baw-download",
+  template: `
+    <ng-container *ngIf="!failure">
```
atruskie: so if there is a failure nothing shows? Might want to address that

src/app/helpers/encoding/encoding.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * TODO Finish comment (Base64 RFC 4648 §4???)
+ * Encode plain text into base64url following the TODO standard
+ *
+ * @param plainText Plain text to encode
+ */
+export function toBase64Url(plainText: string): string {
+  // Convert to base64, then replace unsafe symbols
```
atruskie:
```suggestion
  // Convert to base64, then replace symbols so we conform the base64url spec
```

src/app/helpers/encoding/encoding.ts
```diff
@@ -0,0 +1,38 @@
...
+  return window
+    .btoa(plainText)
```
atruskie: this will fail if called on a SSR rendered page.

src/app/services/baw-api/audio-recording/audio-recordings.service.ts
```diff
@@ -90,6 +93,25 @@ export class AudioRecordingsService extends ReadonlyApi<AudioRecording> {
...
+  }
+
+  /**
+   * Returns a link to the API which will download a templated script for downloading original audio files
+   *
+   * @param filter Audio recording filters
+   */
+  public batchDownloadUrl(filter: Filters<AudioRecording>): string {
+    const qsp = new URLSearchParams({
+      // Base64 RFC 4648 §4 encoding
```
atruskie:
```suggestion
      // Base64 RFC 4648 §5 encoding
```

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+        [model]="model"
+        [fields]="fields"
+        [submitLabel]="undefined"
+        (modelChange)="updateHref()"
+      ></baw-form>
+
+      <div class="clearfix">
+        <a
+          class="btn btn-primary float-end mt-3"
+          [class.disabled]="!href"
```
atruskie: why would this be disabled?

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+          equal to 0), you can skip to the next step. If you have no output, the
+          command fails, or the version does not match, read through
+          <a
+            href="https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-7.2"
+          >
+            this guide.
+          </a>
+        </p>
+
+        <h3>2. Setup terminal</h3>
```
atruskie: you're not really setting up a terminal... just opening it

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+        <ol>
+          <li>
+            Open the containing folder for the batch downloading script you just
+            downloaded in your file explorer
+          </li>
+          <li>
+            Click on the containing folder, and hit the "Copy path" button in
+            the Clipboard section of the context menu (top menu, Home tab). The
+            location path you have copied should look similar to
+            <code>C:/Users/username/Downloads</code>
```
atruskie: Windows uses backslashes

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+
+        <h3>4. Run the script</h3>
+
+        <p>
+          Finally we can run the script. Read through the documentation of the
+          script for more specific advice/use cases. To run the command, take
+          the terminal you setup in the previous steps, and run the following
+          command:
+          <code>./download_audio_files.ps1</code>
+        </p>
```
atruskie: consider just embedding an asciicast

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+
+        <h3>4. Run the script</h3>
+
+        <p>
+          Finally we can run the script. Read through the documentation of the
+          script for more specific advice/use cases. To run the command, take
+          the terminal you setup in the previous steps, and run the following
+          command:
+          <code>./download_audio_files.ps1</code>
```
atruskie: ~~this invocation is wrong - an auth token parameter is required~~

oh i see: you still missed out on telling them to paste the token

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+    return this.models[regionKey] as Region;
+  }
+  public get site(): Site {
+    return this.models[siteKey] as Site;
+  }
+
+  private getProjectFormInput(): FormlyFieldConfig {
+    return {
+      key: "projectId",
+      type: "select",
```
atruskie: just makes these labels if we're not going to support changing them in this version

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  private dateSelectionRange(): FormlyFieldConfig[] {
+    return [
+      {
+        key: "recordingStartedBefore",
+        type: "input",
+        templateOptions: {
+          label: "Date recording started before",
+          required: false,
+          type: "datetime-local",
```
atruskie: but it's not local by definition?

also it should just be a date

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+        disabled: true,
+        options: [{ value: this.site.id, label: this.site.name }],
+      },
+    };
+  }
+
+  private dateSelectionRange(): FormlyFieldConfig[] {
+    return [
+      {
+        key: "recordingStartedBefore",
```
atruskie:
```suggestion
        key: "recordingStartedAfter",
```

though personall I'd choose start and end

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+            "If you're unsure if your data will be included in a filter, expand your date range by one-day in either direction.",
+        },
+      },
+    ];
+  }
+
+  public updateHref(): void {
+    if (!this.project) {
+      return;
+    }
```
atruskie: this doesn't make sense from what i've seen so far? why is only project tested for presence?

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  private generateFilter(): Filters<AudioRecording> {
+    const filter: Filters<AudioRecording> = {};
+    const recordingStartedBefore = DateTime.fromISO(
+      this.model.recordingStartedBefore
+    );
+    const recordingFinishedBefore = DateTime.fromISO(
+      this.model.recordingFinishedBefore
+    );
```
atruskie: if they're already iso8601 strings, why decode them?

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -0,0 +1,389 @@
...
+
+    if (this.model.siteId) {
+      filter["sites.id"] = { equal: this.model.siteId };
+    } else if (this.model.regionId) {
+      filter["sites.regions.id"] = { equal: this.model.regionId };
+    } else if (this.model.projectId) {
+      filter["sites.regions.project.id"] = {
+        equal: this.model.projectId,
+      };
+    }
```
atruskie: feels dirty to have bits of filter query arbitrarily constructed here... remind me to get you to reinvent our QueryBuilder class

atruskie (changes requested):
Ok: with current implementation, the context selectors (project/site/point) look like drop downs? There's a little down arrow there?

If we're not making these interactive, then let's just leave them as labels for now. Disabled controls imply the could be made active.

The date inputs:
- do they open a date selector?
- Why are they not formatted as ISO8601?
- what are the hyphens doing there? If they're time placeholders why are they not using a letter place holder date like the dates?
- but really: just remove time selection altogether.

Lastly, radically simplify the instructions. Remove cruft and padding, provide simple steps and link to further instructions. As a rough guide, cut the copy by about 66%.

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: If there is a failure, the error component will be shown, and this component should show nothing

atruskie: this is not logic that belongs in this component. Whether or not to render a component should be determined by the container of the component

I suspect this is a larger pattern, make an issue to deal with it. Otherwise fix here.

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: This model is just the output of the form, without making the actual form inputs which would output arrays, i cant have this as an array. I can however convert these values to arrays when being passed to the service so that things are prepared for the future change

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: Date form inputs return strings, not dates. They are converted when passed to the api service

src/app/components/audio-recordings/pages/download/download.component.ts
atruskie: I mean, isn't the formly spec for these controls a `select`? so by adding the `multiple` attribute this would happen anyway?

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: Moved to here: https://github.com/QutEcoacoustics/workbench-client/pull/1766#issuecomment-1020798283

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: A project must exist for the script to be of any use, would you prefer for it to just return a script with no filled in values if project is not defined?

src/app/components/audio-recordings/pages/download/download.component.ts
atruskie: that's a problem with the inputs

src/app/components/audio-recordings/pages/download/download.component.ts
atruskie: Why must a project exist?

We need zero to many of projects, regions, sites. If we had 0 of all of them we'd be effectively downloading all recordings

src/app/components/audio-recordings/pages/download/download.component.ts
atruskie: Also inputs can return date values, see `valueAsDate` , see https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

to be clearer: if the model is representing a date, it should be a date, and there should be a value converter between our model and the input:

https://stackoverflow.com/questions/11616636/how-to-do-two-way-filtering-in-angularjs

https://zemke.io/angular-model-converter

and one for reactive forms: https://angular-schule.github.io/website-articles/blog/2016-10-angular-date-value-accessor/README.html

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: https://github.com/QutEcoacoustics/workbench-client/issues/984#issuecomment-1021892132

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: https://github.com/QutEcoacoustics/workbench-client/issues/1777

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+
+  <!-- TODO Extract to CMS? -->
+  <section>
+    <h2>Instructions</h2>
+
+    <p>
+      The above form will guide you to downloading a powershell script which can
+      be used to download the audio files in batches. For easiest use, save to
+      your downloads folder as the following instructions will use that as an
+      example. Please read through this explanation to understand how to run and
```
atruskie: No: the point is the script downloads to the directory where it is saved.  

- download the script
- move it to the directory where you want the downloaded audio to be stored

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+
+    <!-- TODO Show users how with asciicast -->
+
+    <h3>2. Open terminal</h3>
+
+    <p>
+      Next open a terminal in the same folder as your script. In most operating
+      systems the file manager has the option to open the current folder in a
+      terminal. If that is not possible, you will need to copy the folder path,
+      and use <code>cd "FOLDER_LOCATION_HERE"</code> in a powershell terminal.
```
atruskie:
```suggestion
      and use <code>cd "FOLDER_LOCATION_HERE"</code> in a PowerShell shell.
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+    <p>
+      For the various operating systems, the file path will look similar to the
+      following:
+    </p>
+
+    <ul>
+      <li>Windows: <code>C:\Users\username\Downloads</code></li>
+      <li>Linux: <code>/home/username/Downloads</code></li>
+      <li>MacOS: <code>/Users/username/Downloads</code></li>
+    </ul>
```
atruskie: I appreciate the example, but I really don't want people downloading this data into the downloads directory. Try a portable hdd demo path instead?

G:\\
/mnt/sdb
/Volumes/my-passport

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+      <a
+        href="https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-7.2"
+      >
+        this guide.
+      </a>
+    </p>
+
+    <!-- TODO Show users how with asciicast -->
+
+    <h3>2. Open terminal</h3>
```
atruskie:
```suggestion
    <h3>2. Open a terminal</h3>
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+    <ul>
+      <li>Windows: <code>C:\Users\username\Downloads</code></li>
+      <li>Linux: <code>/home/username/Downloads</code></li>
+      <li>MacOS: <code>/Users/username/Downloads</code></li>
+    </ul>
+
+    <p>
+      You can open a powershell terminal by running <code>pwsh</code> inside a
+      normal terminal.
+    </p>
```
atruskie: Just get them to open PowerShell from their default launcher

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+
+    <h3>3. Get Authentication Token</h3>
```
atruskie:
```suggestion
    <h3>3. Get an Authentication Token</h3>
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+
+    <p>
+      Every time you want to run this script, you will need to update the
```
atruskie:
```suggestion
      Every time you want to run this script, you will need to provide an
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+      authentication token (auth token). This is what will identify you to our
+      servers, and you can think of it as being similar to your username and
+      password.
+      <strong class="text-danger">
+        Treat this code like your password, and do not ever share it with
+        anyone!
```
atruskie:
```suggestion
        Treat this code like your password, and do not share it.
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+      </strong>
+      You can find the auth token in your
+      <a [strongRoute]="profile.route">{{ profile.label }}</a> page, under the
+      "Authentication Token" section. Copy the auth token, and keep it for the
```
atruskie:
```suggestion
      <i>Authentication Token</i> section. Copy the auth token, and keep it for the
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+      next step<br />
+      You should perform this step whenever you wish to run the script as the
+      auth token can expire if left for a long enough period of time.
```
atruskie:
```suggestion
```

You said this above already

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+    </p>
+
+    <!-- TODO Show users how with asciicast -->
+
+    <h3>4. Run the script</h3>
+
+    <p>
+      Finally we can run the script. Read through the documentation of the
+      script for more specific advice/use cases. To run the command, take the
```
atruskie:
```suggestion
      Finally we can run the script.  To run the command, in the
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+      terminal you setup in the previous steps, and run the following command
+      (replacing INSERT_AUTH_TOKEN_HERE with your auth token):
+      <code>
+        ./download_audio_files.ps1 -auth_token="INSERT_AUTH_TOKEN_HERE
```
atruskie:
```suggestion
        ./download_audio_files.ps1 -auth_token="INSERT_AUTH_TOKEN_HERE"
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -0,0 +1,167 @@
...
+    </p>
+
+    <!-- TODO Show users how with asciicast -->
+
+    <h3>4. Run the script</h3>
+
+    <p>
+      Finally we can run the script. Read through the documentation of the
+      script for more specific advice/use cases. To run the command, take the
+      terminal you setup in the previous steps, and run the following command
```
atruskie:
```suggestion
      terminal you opened in the previous steps, type in and then run the following command
```

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -218,151 +66,64 @@ class DownloadAudioRecordingsComponent
...
-      defaultValue: this.site.id,
-      templateOptions: {
-        label: hideProject ? "Site" : "Point",
-        // TODO Allow user to change selected site
-        disabled: true,
-        options: [{ value: this.site.id, label: this.site.name }],
-      },
-    };
+  public setRecordingStartedAfterDate(event: Event): void {
+    this.model.recordingStartedAfter = (event.target as HTMLInputElement).value;
```
atruskie: I believe ngForm has an event for whenever the form is updated.

Actually ngForm supports two-way data binding... why aren't you using ngForm?

src/app/helpers/encoding/encoding.ts
```diff
@@ -1,28 +1,43 @@
...
+  if (!window?.btoa) {
+    return "";
+  }
+
+  // Encode special characters (btoa only supports Latin1)
+  //  then convert to base64 (RFC 4648 §4),
+  //  then replace so we conform to the base64url spec (RFC 4648 §5)
   return window
-    .btoa(plainText)
+    .btoa(encodeURIComponent(plainText))
```
atruskie: HMMM this is not ok.

These won't be decoded server side.

The reason we didn't have to do this was because JSON.stringify handles encoding outside of our supported range. So using both stringify and btoa safely side steps the problem.

As far as I know btoa treats it's input as a series of bytes; there should be no case where we have to worry about this. Can you explain/show me why you made this change?

src/app/services/baw-api/audio-recording/audio-recordings.service.ts
atruskie: This suggestion was correct, base64url is section 5 of the RFC. Why did you reject it? https://datatracker.ietf.org/doc/html/rfc4648#section-5

atruskie (changes requested):

src/app/helpers/encoding/encoding.ts
Allcharles: As stated in the comment above, `btoa` only supports Latin1. While `JSON.stringify` handles encoding for us, it is not directly used in this function so cannot be taken as an assumption. We could make it take in an object as an input instead of text? Also, if `JSON.stringify` encodes the text, why will that be decoded server side, but not `encodeURIComponent`?

PR #1803, opened 2022-02-16
Better handling of auth state changes

### LINKED ISSUE #1640: Logout users with expired credentials
Currently, when a user with expired credentials attempts to make any API requests, they are met with an Unauthorized Access error message. Workbench-client should make a `GET /security` request to determine if the credentials are still valid, and if not, log out the user automatically.

Allcharles:
# Better handling of auth state changes

This PR is a rework of how authentication is handled, and addresses some of the desync that can occur between the cookie and the auth token where one can exist without the other.

## Changes

- Fixed broken login form on chrome when running on localhost (run in ssl fixes chrome blocking same site cookies)
- Added support for user data directory for chrome debugger which allows devs to run custom extensions/profiles: https://stackoverflow.com/questions/51725854/run-vscode-chrome-debugger-with-its-extensions
  - May want to add a section to the README explaining how to use this, ideally with information about the angular devtools extension: https://chrome.google.com/webstore/detail/angular-devtools/ienfalfjdbdpebioblfackkekamfmbnh
- Stopped logout button from refreshing page to reload content
- Updated show/filter/list functions in baw api service so that they will retrigger whenever the authenticated state changes. This better handles login/logout happening without a full page refresh
- Removed auth token from session storage
- Update auth token when the angular app loads. This means that the auth token should resync with the auth cookie whenever the page refreshes, and will function with the baw-client pages
  - On slow internet connections, this may add unnecessary delays (potentially store the value in session storage again and validate its not broken instead?)
- De-tangled SessionUser and User models so they are separate entities
  - TODO Move `SessionUser` to own file?

## Problems

- TODO Tests
- Need to validate everything works with SSR and baw-client pages
- Does not handle if cookie is expired, but token exists without a page refresh

## Issues

Addresses #1640 

## Final Checklist


src/app/helpers/custom-errors/baw-api-error.ts
```diff
@@ -0,0 +1,78 @@
...
+    message: string,
+    info?: Record<string, string>
+  );
+
+  public constructor(...args: any[]) {
+    super(args.length === 1 ? args[0].message : args[1]);
+
+    this.status = args.length === 1 ? args[0].status : args[0];
+    this._message = args.length === 1 ? args[0].message : args[1];
+    this.info = args.length === 1 ? args[0].info : args[2] ?? {};
```
atruskie: this is pretty messy IMO

as far as i can tell we have a maximum of three arguments... no need for variadic args. 

Just name the args? Or can we simplify the call sites?

src/app/helpers/custom-errors/baw-api-error.ts
```diff
@@ -0,0 +1,78 @@
...
+export function isApiErrorDetails(
+  errorResponse: any
+): errorResponse is ApiErrorDetails {
+  if (!errorResponse) {
+    return false;
+  }
+
+  const keys = Object.keys(errorResponse);
+  return (
+    keys.length <= 3 && keys.includes("status") && keys.includes("message")
```
atruskie: This seems prone to error. Why 3?

Document magic number, or better yet just use the key presence checks?

atruskie (changes requested):
I think I'd generally like to see a dedicated service being used to check login status (as opposed to the isLoggedIn property on the services' base class). This is mainly a readability nit at this point, but I can see value of properly decoupling session state from our root service classes if needed in the future.

I wrote the above with a weakly held conviction, but upon seeing `baw-api.service.ts` I am doubling down. There's a whole lot of code in there about managing session state, when really all a service needs to do is tack on an auth token when the time comes (if that! the interceptors handle that).  The setting state on static properties again is a smell that the state should be handled in another DI singleton service that is injected into baw-api.serice. Again, not a necessary change.

~~Otherwise, good to SessionUser removed - it was always weird to see it around.~~  Yeah I see it is not removed, but it would be nice if it was. There's no need for SessionUser. `Session`, sure, `User` sure, but why `SessionUser`?

The authTrigger and switchMap combo is brilliant.

src/app/components/home/home.component.ts
```diff
@@ -42,7 +42,7 @@ class HomeComponent extends PageComponent implements OnInit {
   public constructor(
     private regionApi: ShallowRegionsService,
     private projectApi: ProjectsService,
-    private securityApi: SecurityService,
+    private state: BawApiStateService,
```
atruskie: yeah state is as about a general term as exists. Let's choose a better name for these variables.

`session` or `sessionState`

src/app/services/baw-api/baw-api-state.service.ts
```diff
@@ -0,0 +1,57 @@
...
+type GuestUser = typeof guestUser;
+type GuestAuthToken = typeof guestAuthToken;
+
+export interface AuthTriggerData {
+  user: User | GuestUser;
+  authToken?: AuthToken | GuestAuthToken;
+}
+
+@Injectable()
+export class BawApiStateService {
```
atruskie:
```suggestion
export class BawSessionState {
```

We don't need _state_ and _service_ in a name.

atruskie (changes requested):
Referring to the session state as only `state` in used components is not acceptable.

PR #1804, opened 2022-02-18
Restructure API Services

Allcharles:
# Restructure API Services

This is a compliment PR to #1803 which brings greater changes to the underlying structure of how the API services work

## Changes

- API services no longer extend `BawApiService`, and instead use it through DI to handle the API requests
- Fixed various bugs introduced with `BawApiError`
- Implemented a more consistent approach to notifying users when an error has occurred. Notifications are now performed automatically by `BawApiService`, and don't need individual solutions whenever an API call is made 

## Problems

- TODO Tests

## Issues

Addresses #1640

## Final Checklist


src/app/helpers/custom-errors/baw-api-error.ts
```diff
@@ -62,17 +69,16 @@ export interface ApiErrorDetails {
...
+  if (!error || !(error instanceof Object)) {
     return false;
   }
 
-  const keys = Object.keys(errorResponse);
+  const keys = Object.keys(error);
+  keys.sort();
   return (
-    keys.length <= 3 && keys.includes("status") && keys.includes("message")
+    keys === ["message", "status"] || keys === ["info", "message", "status"]
```
atruskie: I've read this a few times and I don't really understand the logic.

If you used `includes` you wouldn't have to sort the keys and this would also be forwards compatible with additional fields added to the error info object.

Maybe add some docs here for future readers?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -71,113 +104,98 @@ export class BawApiService<Model extends AbstractModel> {
   }
 
   public constructor(
-    protected http: HttpClient,
     @Inject(API_ROOT) protected apiRoot: string,
-    @Inject(STUB_MODEL_BUILDER)
-    classBuilder: AbstractModelConstructor<Model>,
+    @Inject(IS_SERVER_PLATFORM) protected isServer: boolean,
```
atruskie, line 107: if i've understood this right, there are no child classes any more? so why `protected`?

atruskie (approved):
Nice work

src/app/services/baw-api/baw-api.service.ts
Allcharles: `BawFormApiService` extends `BawApiService`. Keeping this `protected` may be useful in the future

Allcharles: Although potentially it should import `BawApiService` instead to be more inline with everything else

atruskie: Form service seems like an actual sub-class of enhanced functionality. Keep as is

Issue #1790, opened 2022-02-11
Meeting Notes, 11/02/2022


Allcharles:
# Redesign of Navigation and Actions

Goal is to simplify the navigation and general layout of our website by pulling actions out of the their dedicated menu, and move them into the page contents. This should also better expose the hierarchy of our data and making the association between audio recordings and _points_/sites/regions/projects more apparent.

Major Goals:

- [ ] Move actions in the page body
- [ ] Remove links from panels
  - We may need to explore putting the links into a drop down shortcut menu on the panels. Some functionality such as new/edit/delete model should be easy to find
  - [ ] Add design ideas below
- [x] Finalize #1775 PR

Minor Goals:

- [x] Improve handling of "unknown users"
  - This may require some additional work in the baw-server so that guests are capable of viewing users without logging in
  - [ ] Add design ideas below
- [x] Remove created/updated by panels for audio recordings list page
- [ ] Sort lists by amount of audio or recent uploads
  - Currently on the A2O, a lot of data has yet to be processed resulting in users mostly finding empty sites/_points_
- [x] Remove "play" link from pages (specifically the link which shows the first recording of any group of recordings)
- [x] Have David update the outdated A2O data portal link in the wordpress site: https://acousticobservatory.org/data/find-data/
- [x] Remove link to parent route from the actions menu
- [x] Reinstate breadcrumb navigation
- [x] Add design ideas below
- [x] Make "visualize" link less prominent on the website, or at least removed on the A2O version of the website until it has been properly redesigned

Allcharles:
## Reinstate breadcrumb navigation

Current solution for breadcrumbs is to show location in the secondary menu:

![image](https://user-images.githubusercontent.com/3955116/153532868-07ec05e7-fb7d-4cb3-97a0-1e5c7723f3f3.png)

A nicer solution may be to move the breadcrumbs above the page component using bootstraps breadcrumb system:

![image](https://user-images.githubusercontent.com/3955116/153533097-6dc6c6d4-742a-408d-8f3a-19c24e5890ff.png)

Alternative:

![image](https://user-images.githubusercontent.com/3955116/153546669-4346d700-6280-48af-94f7-961b08db31fa.png)

Allcharles:
# Move actions in page body

## Current design

![image](https://user-images.githubusercontent.com/3955116/153973059-63a281c6-d4d0-4061-b561-9d6d18b1133f.png)

## Simple adjustments

[...]

PR #1807, opened 2022-02-24
Breadcrumbs and hidden secondary menu

Allcharles:
# Breadcrumbs and hidden secondary menu

Removed secondary menu from page, and added breadcrumb to the top of any nested page

## Changes

- Added breadcrumb component
- Extracted secondary menu to side nav component
- `hideProjects` environment setting now causes root route of website to be `shallowRegionsMenuRoute` instead of `projectsMenuItem` whenever looking at project descendant pages

## Problems

- The current implementation of the `breadcrumbResolve` and its use in the `.menus.ts` file causes circular dependencies. This is because the method requires the model for `instanceof` comparisons, and the model requires a `viewUrl` function which generally relates to the menu item for the model which is calling `breadcrumbResolve`. Some possible solutions are the following:
  - Remove the concept of `viewUrl` from the `AbstractModel`s
  - Dynamic imports?
  - Extract the `StrongRoute`s to a `.routes.ts` file, and use them directly instead of the `MenuItem`s for the `viewUrl` paths

## Issues

Relates to #1790 

## Top Route Pages

image

## Nested Route Pages

image
image
image

## Final Checklist


src/app/components/shared/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  ) {
+    super();
+  }
+
+  public ngOnInit(): void {
+    this.menu.menuUpdate
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe(({ breadcrumbs }): void => {
+        this.breadcrumbs = breadcrumbs;
+      });
```
atruskie: this is a good candidate for async pipes - the logic here only deals with unsubscribes  and ~~the state is not used elsewhere in the class~~ (ok it is, but that's what a query function is for right?)

src/app/services/menu/menu.service.ts
```diff
@@ -182,48 +220,53 @@ export class MenuService extends withUnsubscribe() {
...
   }
 
+  private rootToMenuRoute(menuRoute: MenuRoute): OrderedSet<MenuRoute> {
+    let menuRoutes = OrderedSet<MenuRoute>([menuRoute]);
+    let current: MenuRoute = menuRoute;
+    while (current.parent) {
+      current = current.parent;
+
+      if (
+        this.config.settings.hideProjects &&
```
atruskie: you might pull this value out of the loop and assign it to a const
```suggestion
        hideProjects &&
```

src/app/services/menu/menu.service.ts
```diff
@@ -182,48 +220,53 @@ export class MenuService extends withUnsubscribe() {
...
+        [projectMenuItem, projectsMenuItem].includes(current)
+      ) {
+        // Skip project menu items if the setting is enabled, duplicate menu
+        // items will be removed because this is a set
+        menuRoutes = menuRoutes.add(shallowRegionsMenuItem);
```
atruskie: more like _substitute_?

Also a subtly that I just realized is that `shallowRegionsMenuItem` would have no parent, and hence is  base case and would stop the route traversal. Might be worth mentioning?

atruskie (approved):
There should always be a breadcrumb for consistency. So for the root site list page, there should  be a `Sites` crumb - especially as that root crumb shows on other pages.

Solid, tight. Well done.

PR #1820, opened 2022-03-04
Fix audio recording recorded date

### LINKED ISSUE #1815: Wrong timezone given for audio recording recordedAt field
`AudioRecording` model has two fields; `recordedDate` and `recodedUtcOffset` which are of importance. The baw server returns the `recordedDate` using a UTC offset currently. This means that in places, we are displaying the wrong value, such as the audio recordings list pages.

Example:
https://data.acousticobservatory.org/projects/1/regions/16/points/61/audio_recordings
Audio recordings tooltips show that the timezone is in Australia/Perth, however the actual displayed values are in UTC.

Allcharles:
# Fix audio recording recorded date

Updated `AudioRecording` model to use the `recordedDateTimezone` so that references to `recordedDate` will have the correct value

## Changes

- Removed old admin panel for audio recordings
- Updated `AudioRecording` model
- Fixed audio recordings page

## Issues

Addresses #1815

## Final Checklist


src/app/models/AudioRecording.ts
```diff
@@ -38,7 +42,7 @@ export interface IAudioRecording extends HasAllUsers {
   status?: AudioRecordingStatus;
   notes?: Blob | any;
   originalFileName?: string;
-  recordedUtcOffset?: string;
```
atruskie, line 45: technically this field does exist, it's just not used by our harvester

src/app/models/AudioRecording.ts
```diff
@@ -100,15 +104,28 @@ export class AudioRecording
     return listenRecordingMenuItem.route.format({ audioRecordingId: this.id });
   }
 
+  public constructor(data: IAudioRecording, injector?: Injector) {
+    super(data, injector);
+
+    // Set timezone on recorded date
+    if (this.recordedDateTimezone) {
+      this.recordedDate = this.recordedDate.setZone(this.recordedDateTimezone);
```
atruskie, line 113: add a note to revert this.

If the recording were to be saved back to the database, there is a chance it could be saved in the wrong format since it is not longer a UTC date. How is it serialized? 

Even if the above point is not valid (it should not be), all dates are in UTC in the database, in the server. We should avoid changing the semantic meaning of this field.

atruskie (approved):
Ok, I'm going to accept these changes since we're under time pressure.  But like I said in https://github.com/QutEcoacoustics/workbench-client/issues/1815#issuecomment-1057516627 this should really be dealt with by our time component (#1736). So when that component is complete, we should expect the changes to the model here to be reverted.

Probably would have been good to signal intention to remove admin audio recording panels before doing it. But I do see how they're redundant. 

~~There are also a lot of menu item changes that are orthogonal to this PR. They're not described in your notes. High level update please? I assume you're just making things consistent?~~

The diff i'm seeing is showing multiple changesets from multiple PRs

src/app/models/AudioRecording.ts
Allcharles: > all dates are in UTC in the database, in the server

This sounds a lot more like a server issue than a front end issue. If dates should be stored as UTC in the database, this should be handled server-side. Doubly so if the reason you say this is because some values on the server are stored with a timezone instead of having the timezone stored separately (I'm not sure if any such fields exist). Controlling this logic client side sounds like a problem waiting to happen, especially as we expose and expect users to potentially directly use the API.

atruskie: Time zones are not offsets.  Time zones cannot be stored in date stamps.

If you send us date with an offset we will convert the date back to UTC when it is stored.

Regardless, there is no point modifying this field if we have a plan for showing these dates consistently, which for whatever reason you've chosen not to employ here 

Mark this as revertible, log a note in #1736 and move on.

We can discuss further later

PR #1863, opened 2022-03-28
Visualize link bug

### LINKED ISSUE #1854: Visualize link broken in Ecosounds and A2O
**Describe the bug**

The visualize link is broken for Ecosounds and A2O. Any navigation to the page produces the error message:

```
The page is unable to be loaded. The error is: No query string options were specified.
```

And it does appear query string options are not showing.

**To Reproduce**

1. Go to Ecosounds or A2O
2. Click on any visualize link

**Expected behavior**

A working visualize page.

**Screenshots**

image

Allcharles:
# Broken Strong Route Directive

A few links were broken on the site because they did not have proper access to the data required to set query parameters. `StrongRouteDirective` was updated to pass route parameters to the routes so they can access them. In the case of the visualize link, it was unable to grab the project/region/site/_point_ ids from the route params, and insert them into its query params. A fix was also implemented as the directive was no longer properly passing resolved models to the routes.

## Changes

- Updated generic test functions `assertUrl`, `assertStrongRoute`, `assertStrongRouteActive`, `assertIcon`
- Created `MockDirectivesModule`
  - Updates to `StrongRouteDirective` causes it to require a few services be mocked properly. This interfered with a lot of tests which were already mocking the services
- Fixed `hideProjects` environment setting breaking new regions page and link
- In-lined `AuthenticatedImageModule` into `DirectivesModule` (did not need to be separate)

## Issues

- Closes #1854

## Final Checklist


src/app/services/shared-activated-route/shared-activated-route.service.ts
```diff
@@ -134,12 +134,21 @@ export class SharedActivatedRouteService extends withUnsubscribe() {
   private observeRouterEvents(router: Router, route: ActivatedRoute): void {
     /** Recursively find all children, and return in order with observable */
     const expandAllChildren = (): Observable<ActivatedRoute> => {
+      // Set a maximum depth to check. The spectator ActivatedRouteStub has
+      // infinite children, which causes errors during tests
+      const maxDepth = 5;
+      let depth = 0;
```
atruskie, line 140: this should probably be a weakref list of instances visited to avoid the recursion... depth limits in recursion are always code smells

atruskie (approved):
I really dislike how all the updated assert functions take in objects, there's a lot of noise around the invocations. I think the functions should all have a target and an expected value arguments, and if you still want flexible assertions with the catch-all object argument, you can pass that as a third argument.

I would also suggest that designing actual assertion extensions would give us the proper power to make extensible assertions without the object catch all arg.

PR #1896, opened 2022-04-22
Time and date pickers

### LINKED ISSUE #1853: 24-hour time for time of day time picker
User feedback for our time of day inputs on the batch download page suggests that AM/PM is not the time format they were expecting.

This leads into our existing idea that the formats were insufficient.

Change the date and time inputs to use the ISO8601 locales as is consistent with the rest of our UI for the batch download page.

Allcharles:
# Time and date pickers

Implemented time and date pickers for the download audio recordings page

## Changes

- 24 hour time picker
- yyyy-mm-dd date picker

## Problems

- Some missing tests
- Would like to show the audio recording dates in the calendar

## Issues

- Closes #1853 

## Visual Changes

image

image

## Final Checklist


src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -165,17 +167,27 @@ class DownloadAudioRecordingsComponent
...
 
+  /** Convert date to string in format of yyyy-mm-dd */
+  private formatDate(date: Date): string {
+    const zeroPad = (num: number): string => String(num).padStart(2, "0");
+    const year = date.getFullYear();
+    const month = zeroPad(date.getMonth() + 1);
+    const day = zeroPad(date.getDate());
+    return `${year}-${month}-${day}`;
+  }
+
```
atruskie: ok multiple things:

- we have a date library that can do this formatting, why not use it?
- why are you just not converting the date using `toISOString`?

src/app/components/shared/input/date/date.component.ts
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  public ngOnChanges(): void {
+    const convertDateToStruct = (date: Date) => ({
+      year: date.getFullYear(),
+      month: date.getMonth() + 1,
+      day: date.getDate(),
+    });
+    const convertStructToDate = (struct: NgbDateStruct) =>
+      new Date(struct.year, struct.month - 1, struct.day);
```
atruskie: neither of these methods have anything to with structs. fix the naming.

src/app/components/shared/input/date/date.component.ts
```diff
@@ -0,0 +1,310 @@
...
+    } else {
+      this.minDate = convertStructToDate(this.minDateStruct);
+    }
+
+    if (this.maxDate) {
+      this.maxDateStruct = convertDateToStruct(this.maxDate);
+    } else {
+      this.maxDate = convertStructToDate(this.maxDateStruct);
+    }
+  }
```
atruskie: why are you avoiding a dependency on luxon?

src/app/components/shared/input/date/date.component.ts
```diff
@@ -0,0 +1,310 @@
...
+      return { minLength: true };
+    } else if (value.split("-").length !== 3) {
+      // Validate delimiters exist
+      return { delimiter: true };
+    }
+
+    const segments = value.split("-");
+    const year = parseInt(segments[0], 10);
+    const month = parseInt(segments[1], 10);
+    const day = parseInt(segments[2], 10);
```
atruskie: ...just parse the date? Why are we breaking out these components?

src/app/components/shared/input/date/date.component.ts
```diff
@@ -0,0 +1,310 @@
...
+    if (year < this.minDateStruct.year || year > this.maxDateStruct.year) {
+      // Validate year
+      return { year: true };
+    } else if (month < 1 || month > 12) {
+      // Validate month
+      return { month: true };
+    } else if (day < 1 || day > 31) {
+      // Validate day
+      return { day: true };
+    }
```
atruskie: Why are checking only the year component is within range?

src/app/components/shared/input/date/date.component.ts
```diff
@@ -0,0 +1,310 @@
...
+    const day = parseInt(segments[2], 10);
+
+    if (year < this.minDateStruct.year || year > this.maxDateStruct.year) {
+      // Validate year
+      return { year: true };
+    } else if (month < 1 || month > 12) {
+      // Validate month
+      return { month: true };
+    } else if (day < 1 || day > 31) {
+      // Validate day
```
atruskie: if I understand `ValidateErrors` correctly then these comments are wrong? Your returning an error? _Validate_ is a verb, you've but if you're in the branch you've already decided.

src/app/components/shared/input/date/date.component.ts
```diff
@@ -0,0 +1,310 @@
...
+      : null;
+  }
+
+  public onInput(event: Event): void {
+    this.dirty = true;
+    this.touched = true;
+    this.onTouched();
+
+    const value = (event.target as HTMLInputElement).value;
+    this._errors = this.validateInput(value);
```
atruskie: if you had bound to the `dateSelect` event you would have gotten back a `NgbDate` and would have avoided having to parse the string?

src/app/components/shared/input/time/time.component.ts
```diff
@@ -0,0 +1,209 @@
...
+    } else if (value.length < this.inputLength) {
+      // Validate length of input
+      return { minLength: true };
+    } else if (!value.includes(":")) {
+      // Validate value includes a colon
+      return { delimiter: true };
+    } else if (!/^(([0-1]\d)|2[0-4]):/.test(value)) {
+      // Validate hours are between 00 - 24
+      return { hour: true };
+    } else if (!/(:[0-5]\d)$/.test(value)) {
```
atruskie: name these regexes

src/app/components/shared/input/time/time.component.ts
```diff
@@ -0,0 +1,209 @@
...
+      // Validate minutes are between 00 - 59
+      return { minute: true };
+    } else if (value.startsWith("24") && !value.endsWith("00")) {
+      // Validate that only 24:00 is valid
+      return { afterMidnight: true };
+    } else {
+      // Validate time is valid
```
atruskie: comments are a bit weird as previous comment

atruskie (changes requested):
I think the `date` component is overkill. It just seems to be a very thin wrapper around `ngbDatePicker` - which itself implements `ControlValueAccessor`!

This seems like it could have been a lot simpler. Did you try passing in a form/model reference to the child component?

I think this code should be radically simpler. The time component actually has a mild excuse to be more complex, but the DateComponent is simply wrapping another already complex element.

src/app/components/audio-recordings/pages/download/download.component.ts
Allcharles: `toISOString` converts back to UTC, date picker outputs with a timezone, using `toISOString` causes the formatted output to be wrong. I could convert to luxon `DateTime`, just thought the conversion was unnecessary given the complexity

src/app/components/shared/input/date/date.component.ts
Allcharles: The name of the type converting to and from is `NgbDateStruct`, i could change to `convertDateToNgbDateStruct` if you would prefer?

src/app/components/shared/input/date/date.component.ts
Allcharles: Because this is converting from javascript dates to ngb dates, nothing on this conversion process would benefit or require luxon? The form output is a vanilla javascript date, something controlled by the ngb date picker

src/app/components/shared/input/date/date.component.ts
Allcharles: Because i dont know if the date is valid, and just parsing the date doesnt tell me what part is invalid. If i just parse the date, the error message would have to be a lot more generic

src/app/components/shared/input/date/date.component.ts
Allcharles: It is far more complex to validate the month and day ranges are within range, as it only matters if the year matches the min or max dates. Much simpler to handle this process after conversion to dates is done

src/app/components/shared/input/date/date.component.ts
Allcharles: I have bound to the `dateSelect` event, this event is for when users manually input a date?

src/app/components/audio-recordings/pages/download/download.component.ts
atruskie: The server treats the date as a UTC datetime anyway?

https://ng-bootstrap.github.io/#/components/datepicker/api#NgbDateParserFormatter handles the formatting of the date for display to a user.

Either way still seems redundant?

And again: luxon

src/app/components/shared/input/date/date.component.ts
atruskie: I'd avoid the term struct altogether. 

While there is [a proposal](https://github.com/tc39/proposal-structs) for JS to get structs, it currently does not have them. Just because a library uses a bad name doesn't mean we're going to get sucked in as well.

Also they have a converter: https://ng-bootstrap.github.io/#/components/datepicker/api#NgbDateAdapter

src/app/components/shared/input/date/date.component.ts
atruskie: https://ng-bootstrap.github.io/#/components/datepicker/api#NgbDateAdapter

src/app/components/shared/input/date/date.component.ts
atruskie: Fair, kind of. 

The quickest path here is just to `Date.parse` and in the case on invalid then work out what the error is in the case of failure.

I appreciate good feedback to the users.

src/app/components/shared/input/date/date.component.ts
atruskie: Huh?

```
x = Date.parse(...)
if (x < this.minDate || x > this.maxDate) {
  // fail validation?
}
```

Why is it harder?

Your component provides this definition:

```
  /** Minimum date allowed in selection */
  @Input() public minDate: Date;
  /** Maximum date allowed in selection */
  @Input() public maxDate: Date;
```

It in no way suggests that only the year component from an entire date object is considered in validating it is in range.

atruskie: Oh I see. You validate for if the date is in range below. In which case this year validation is simply redundant and not needed.

src/app/components/shared/input/date/date.component.ts
atruskie: I'm not sure on this one, but the component seems to handle this itself with very little code in https://ng-bootstrap.github.io/#/components/datepicker/examples#popup

The validators typically watch the model and validate changes, they don't need to bind to events themselves.

atruskie (approved):
Your gut instinct is probably telling you that this expanded definition of date control on the form is bad.

And I kinda agree. If you could work out how to pass the form/model reference into a component you could create a wrapper component that simply collects related layout components together. More syntactic sugar templating than an actual component. And I think that's what you probably wanted at the start of this, just a neat way to bundle up some common html bits.

Unfortunately I've only seen full component definitions. Because they effectively isolate all scope you are then pushed into creating a fully custom component with a control value accessor and all the unit tests associated with a custom control.

> Did you try passing in a form/model reference to the child component?

Since I keep asking and since you keep not answering I'm going to assume you don't understand what I mean. Hopefully the above descriptions shed some light.

What I do _not_ what is the responsibility to maintain a component like a date picker in our code base. This is why I pushed so heavily to find an existing component that could do what we wanted. So the current solution is the most preferable unless you work out what I am talking about.

PR #1932, opened 2022-06-01
Api Request Caching

### LINKED ISSUE #16: Caching Resources
Application could receive a significant performance boost with the additional caching of internal and external resources. This will also improve the Google Audit score improving website visibility.

Related links:
https://angularfirebase.com/lessons/installable-angular-progressive-web-app/
https://medium.com/codingthesmartway-com-blog/angular-5-service-worker-b722e571e306
https://en.wikipedia.org/wiki/HTTP/2_Server_Push
https://developers.google.com/web/fundamentals/web-app-manifest
https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content

Allcharles:
# Api Request Caching

Implemented caching for `show` API requests with a time to live of 50ms. This significantly improves the loading speed of pages which make multiple requests for the same resource.

## Issues

Closes #16

## Final Checklist

### DIFF
--- a/src/app/services/baw-api/baw-api.service.ts
+++ b/src/app/services/baw-api/baw-api.service.ts
@@ -138,7 +138,9 @@ export class BawApiService<
    */
   public list(classBuilder: ClassBuilder, path: string): Observable<Model[]> {
     return this.session.authTrigger.pipe(
-      switchMap(() => this.httpGet(path)),
+      switchMap(() =>
+        this.httpGet(path, defaultApiHeaders, { cache: cacheSettings.enabled })
+      ),
       map(this.handleCollectionResponse(classBuilder))
     );
   }
@@ -167,15 +169,10 @@ export class BawApiService<
    * @param classBuilder Model to create
    * @param path API path
    */
-  public show(
-    classBuilder: ClassBuilder,
-    path: string,
-    cacheResponse = true
-  ): Observable<Model> {
-    cacheResponse = cacheResponse && cacheSettings.enabled;
+  public show(classBuilder: ClassBuilder, path: string): Observable<Model> {
     return this.session.authTrigger.pipe(
       switchMap(() =>
-        this.httpGet(path, defaultApiHeaders, { cache: cacheResponse })
+        this.httpGet(path, defaultApiHeaders, { cache: cacheSettings.enabled })
       ),
       map(this.handleSingleResponse(classBuilder))
     );
@@ -270,9 +267,9 @@ export class BawApiService<
       responseType: "json",
       headers: options,
       context: withCache({
-        ...cacheOptions,
-        ttl: 50,
+        ttl: cacheSettings.httpGetTtlMs,
         context: withCacheLogging(),
+        ...cacheOptions,
       }),
     });
   }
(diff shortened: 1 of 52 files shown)


atruskie (commented):
Cashew looks great. Good pick for a library.

Some notes:

- I'd like some logging included for cached requests. Even if we can turn it off it would be good to have that data if needed
- Our primary use case for caching objects is when loading associations. Have you thought about narrowing the scope of the caching so it is only done in the association code?
- We know when a cache should be invalidated! When update/delete is called we should use cashew's CacheManager to destroy the cached request
- If the above two points are actionable, we could increase the TTL up significantly
- I'd love to get a sense of how the performance has changed with caching. Can you report some numbers?
- It'd be good if we could disable caching for debugging purposes, in prod if we had to.

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -223,11 +235,11 @@ export class BawApiService<
     if (body?.hasFormDataOnlyAttributes({ update: true })) {
       const formData = body.getFormDataOnlyAttributes({ update: true });
       return request.pipe(
-        mergeMap(() => this.httpPut(path, formData, multiPartHeaders)),
+        mergeMap(() => this.httpPut(path, formData, multiPartApiHeaders)),
         map(this.handleSingleResponse(classBuilder))
       );
     }
-    return request;
+    return request.pipe(tap(() => this.manager.delete(path)));
```
atruskie: should probably use their KeySerializer  https://github.com/ngneat/cashew#hack-the-library

image


I know we aren't using any QSP for our get requests but if we do this would break

atruskie (approved):
Looks good.

Using dedicated interceptors for logging seems a bit overkill but I don't have a better solution.

I'd like to talk to you in chat about the association only idea. I'm still confused as to why it wouldn't work

Allcharles:
# Cached vs Normal Page Load:

To try to mimic real world values, I have performed all of the tests running the compiled code from a docker container, similar to how our SSR tests are performed. I have used http://development.ecosounds.org:4000/projects/1029/audio_recordings because it is a worst case scenario with 25 duplicate requests for both users and sites.

## Full Page Load

From refresh, to final load

### Cached

Test 1

- SSR Page Load: 1.13s
- Table First Load: 2.02s (+0.89s)
- Table Finish Loading: 2.64s (+0.62s)

Test 2

- SSR Page Load: 0.97s
- Table First Load: 1.85s (+0.88s)
- Table Finish Loading: 2.43s (+0.58s)

Test 3

- SSR Page Load: 1.26s
- Table First Load: 2.18s (+0.92s)
- Table Finish Loading: 2.81s (+0.63s)

### Normal

Test 1

- SSR Page Load: 1.39s
- Table First Load: 2.37s (+0.98s)
- Table Finish Loading: 3.6s (+1.23s)

Test 2

- SSR Page Load: 1.3s
- Table First Load: 2.22s (+0.92s)
- Table Finish Loading: 3.61s (+1.39s)

Test 3

- SSR Page Load: 1.33s
- Table First Load: 2.29s (+0.96s)
- Table Finish Loading: 3.62s (+1.33s)

Allcharles:
> Using dedicated interceptors for logging seems a bit overkill but I don't have a better solution.

I could change the api interceptor to add the header data to the request, this way i should be able to just use the `responseSerializer` function from the `HttpCacheInterceptorModule`

PR #1943, opened 2022-06-29
Fix forms and notifications

### LINKED ISSUE #1880: First load failing authed resolved models
First load of website is resolving models using the guest details. This is because the authToken for api requests is updated after resolved model requests are made. A system to fix this would be to force angular to wait for the auth token api request to resolve, before finalizing the bootstrap process. This will likely use the `APP_INITIALIZER`

### LINKED ISSUE #1942: Various broken forms
A number of forms have been broken over time. Check and validate issues with the following forms:

- [ ] Project new/edit form not showing error notifications
- [ ] Login form not redirecting on success, or showing notifications
- [ ] Register form not redirecting on success, or showing notifications

Allcharles:
# Fix forms and notifications

There have been a few various bugs affecting notifications and forms in general. This attempts to patch them, and standardize the approach to handling notifications.

## Changes

- Prevent duplicate notifications from popping up
- Added `disableNotification` and `disableCache` options to baw api service
- Fix issues with auth forms
- Fixes bug in `createHarvestPredicate`

## Problems

- Need tests

## Issues

- Closes #1942 
- Closes #1880 

## Final Checklist

### DIFF
--- a/src/app/services/baw-api/security/security.service.ts
+++ b/src/app/services/baw-api/security/security.service.ts
@@ -1,4 +1,5 @@
 import { Injectable } from "@angular/core";
+import { Router } from "@angular/router";
 import { emptyParam, param } from "@baw-api/api-common";
 import { BawApiService } from "@baw-api/baw-api.service";
 import {
@@ -7,6 +8,7 @@ import {
 } from "@baw-api/baw-form-api.service";
 import { BawSessionService } from "@baw-api/baw-session.service";
 import { reportProblemMenuItem } from "@components/report-problem/report-problem.menus";
+import { BawApiError } from "@helpers/custom-errors/baw-api-error";
 import { stringTemplate } from "@helpers/stringTemplate/stringTemplate";
 import { AuthToken } from "@interfaces/apiInterfaces";
 import { LoginDetails } from "@models/data/LoginDetails";
@@ -14,7 +16,7 @@ import { RegisterDetails } from "@models/data/RegisterDetails";
 import { Session, User } from "@models/User";
 import { UNAUTHORIZED } from "http-status";
 import { CookieService } from "ngx-cookie-service";
-import { Observable, throwError } from "rxjs";
+import { Observable } from "rxjs";
 import { catchError, first, map, mergeMap, tap } from "rxjs/operators";
 import { UserService } from "../user/user.service";
 
@@ -36,7 +38,8 @@ export class SecurityService {
     private formApi: BawFormApiService<Session>,
     private userService: UserService,
     private cookies: CookieService,
-    private session: BawSessionService
+    private session: BawSessionService,
+    private router: Router
   ) {
     this.updateAuthToken();
   }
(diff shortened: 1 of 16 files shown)


src/app/services/baw-api/security/security.service.ts
```diff
@@ -137,20 +145,22 @@ export class SecurityService {
...
-      tap(() => this.clearData()),
-      catchError((err) => {
-        this.clearData();
-        // Don't use handleError function from api service, as it will throw
-        // out a notification
-        return throwError(() => err);
-      })
-    ) as Observable<void>;
+    return this.api
+      .destroy(signOutEndpoint(), { disableNotification: true })
```
atruskie: comment why disabling notifications is necessary here

src/app/services/baw-api/security/security.service.ts
```diff
@@ -209,7 +221,7 @@ export class SecurityService {
     this.sessionDetails()
       .pipe(
         tap((user) => (authToken = user.authToken)),
-        mergeMap(() => this.userService.show()),
+        mergeMap(() => this.userService.showWithoutNotification()),
```
atruskie, line 230: seems unbalanced, you just pass the option above in `destroy` why a special helper method here?

src/app/services/baw-api/security/security.service.ts
```diff
@@ -137,20 +145,22 @@ export class SecurityService {
...
+          return this.api.handleError(err, true);
+        })
+      );
   }
 
   /** Get details of currently logged in user */
   public sessionDetails(): Observable<Session> {
-    return this.api.show(Session, sessionUserEndpoint(Date.now().toString()));
+    return this.api.show(Session, sessionUserEndpoint(Date.now().toString()), {
+      disableNotification: true,
```
atruskie, line 168: add comment explaining why we're disabling notifications

atruskie (changes requested):
OK, all this `runGuardsAndResolvers` and the navigate on logout is poorly explained and not addressing issues previously discussed.

Remove changes to `runGuardsAndResolvers`, and `onSameUrlNavigation`, and the router navigation on logout, and let's spec out desired behaviour before we attempt another solution.

src/app/services/baw-api/security/security.service.ts
Allcharles: I cannot update the inputs of the `UserService`'s `show` method without causing significant churn. So instead i just created a `showWithoutNotification` method to be used by the `SecurityService`. In the case of the `signOut` function, there is no scenario where I would want to show the failure notification, because even if failure occurs, I will just clear the cookie and the auth token which achieves similar results.

Allcharles: One could argue that i shouldn't attempt to forcefully signout the user on the clientside only if the api fails to respond, and as such should create a `signOutWithoutNofication` method

atruskie (approved):

PR #1948, opened 2022-07-01
Harvest Instruction Changes

Allcharles:
# Minor changes to harvest instructions

- Added instructions on creating new sites if required to meta review screen
- Added instruction for when WinSCP gives with untrusted server prompt
- Added instruction for using Ethernet instead of WiFi
- Added instruction telling users to close a connection to the SFTP server before transitioning state
- Fixed bug in harvest service `transitionStatus` where status was not being case converted
- Removed old harvest instructions page detailing how to create harvest.yml files
- Added instruction to harvest list page explaining how to get permission to upload audio

## Final Checklist


src/app/components/harvest/screens/metadata-review/metadata-review.component.html
```diff
@@ -1,6 +1,11 @@
 <h3>Review</h3>
 
-<p>This is a review of the audio data</p>
+<p>
+  This is a review of the audio data. If you are missing a
+  {{ siteColumnLabel }} for one of the folders you have uploaded,
+  <a [strongRoute]="newSiteMenuItem.route">create the {{ siteColumnLabel }}</a>
+  and refresh this page
```
atruskie:
```suggestion
  and refresh this page.
```

src/app/components/harvest/screens/uploading/batch-uploading.component.html
```diff
@@ -47,6 +52,10 @@ <h3>Uploading Files</h3>
 <div [ngbNavOutlet]="nav" class="mt-2"></div>
 
 <ul>
+  <li>
+    We recommend utilising an Ethernet connection instead of WiFi. The file
+    transfer will be more stable, and slightly faster
+  </li>
```
atruskie:
```suggestion
    We recommend using an Ethernet connection instead of WiFi. The file
    transfer will be more stable, and faster
  </li>
```

src/app/components/harvest/screens/uploading/batch-uploading.component.html
```diff
@@ -55,6 +64,10 @@ <h3>Uploading Files</h3>
     Make sure files from each different sensor are in their own folder and are
     not mixed together
   </li>
+  <li>
+    When you are finished uploading files with WinSCP, close the application and
+    click on the "Finished uploading" button below
```
atruskie:
```suggestion
    When you are finished uploading files with WinSCP, close the WinSCP and
    click on the "Finished uploading" button below to proceed
```

src/app/services/baw-api/harvest/harvest.service.ts
```diff
@@ -153,7 +154,7 @@ export class ShallowHarvestsService implements StandardApi<Harvest> {
   ): Observable<Harvest> {
     return this.api
       .httpPatch(shallowEndpoint(model, emptyParam), {
-        [this.harvestKind]: { status },
+        [this.harvestKind]: { status: snakeCase(status) },
```
atruskie: i was thinking that we fixed this? 

but because we're not using toJson  this slipped through.

maybe this should be a call toJson and then we pick a just the key we want from the serialized object after that?

atruskie (approved):

src/app/services/baw-api/harvest/harvest.service.ts
atruskie: or make a helper method like toJson that allows us to pick just one field, and it runs through our converters

Allcharles: In this case, we are directly passing the status to this function instead of the model. As such we don't have access to the `toJson` function. Ie:

```typescript
this.stages.transition("complete", () => (this.loading = false));
```

```typescript
public transition(
    stage: HarvestStatus,
    onComplete: (resp: Harvest | BawApiError) => void
  ): void {
    // We want this api request to complete regardless of lifecycle destruction
    // eslint-disable-next-line rxjs-angular/prefer-takeuntil
    this.harvestApi.transitionStatus(this.harvest, stage).subscribe({
      next: (harvest): void => {
        onComplete(harvest);
        this.setHarvest(harvest);
      },
      error: (err: BawApiError) => {
        onComplete(err);
        this.notifications.error(err.message);
      },
    });
  }
```

atruskie: This feels wrong and like it has potential for future errors. I don't have a better solution though...

PR #1972, opened 2022-08-02
Google analytics

### LINKED ISSUE #19: Google Analytics Tag
Add google analytics to website. This should utilise the following module: https://github.com/revolunet/angular-google-analytics

### LINKED ISSUE #1969: Harvest: stuck on scanning screen
While demoing the uploader we got stuck on the _scanning_ screen two separate times.

A refresh of the page fixed the issue.

Allcharles:
# Implemented Google Analytics

## Changes

- Implemented google analytics
- Cleanup config system
- Overrode outdated peer dependencies in package.json
- Fixed bug in harvest stages service

## Problems

- TODO Tests

## Issues

- Fixes #1969 
- Fixes #19

## Final Checklist


src/app/helpers/embedScript/embedGoogleAnalytics.ts
```diff
@@ -0,0 +1,54 @@
...
+      } else if (count > 10) {
+        console.error("Failed to load google analytics.");
+        reject("Google Analytics API Bundle took too long to download.");
+      } else {
+        count++;
+        setTimeout(() => analyticsLoaded(), defaultDebounceTime);
+      }
+    }
+
+    analyticsLoaded();
```
atruskie: Ok: this does not seem very robust.

i think this is Google's example for how to set tags that a read when the script is laoded: https://developers.google.com/tag-platform/gtagjs/install

src/app/helpers/embedScript/embedGoogleAnalytics.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (typeof gtag !== "undefined") {
+        gtag("js", new Date());
+        gtag("config", key);
+        resolve();
+      } else if (count > 10) {
+        console.error("Failed to load google analytics.");
+        reject("Google Analytics API Bundle took too long to download.");
+      } else {
+        count++;
+        setTimeout(() => analyticsLoaded(), defaultDebounceTime);
```
atruskie: if this is actually a 500ms debounce then that is pretty huge

src/app/helpers/embedScript/embedGoogleAnalytics.ts
atruskie: https://developers.google.com/tag-platform/tag-manager/web/datalayer

src/app/services/config/config.module.ts
```diff
@@ -1,26 +1,29 @@
...
+import { API_CONFIG, API_ROOT } from "./config.tokens";
 
 @NgModule({
   imports: [ToastrModule],
   providers: [
     {
       provide: APP_INITIALIZER,
       useFactory: AppInitializer.initializerFactory,
+      deps: [
+        [new Optional(), API_CONFIG],
```
atruskie, line 16: what does this mean? I am unfamiliar with this syntax

src/app/services/config/config.service.ts
```diff
@@ -19,30 +22,52 @@ export const assetRoot = "/assets";
...
+      this.setConfig(await defaultConfig);
+      await embedGoogleServicesIfValid();
       return;
     }
 
-    this._config = new Proxy(environment, {});
-    this.theme.setTheme(this.settings.theme ?? {});
+    return firstValueFrom(
+      this.http.get("assets/environment.json").pipe(
+        retry({ count: 5, delay: 1000 }),
```
atruskie: why such a large delay?

src/index.html
```diff
@@ -16,6 +16,17 @@
...
+      Create dataLayer for google analytics:
+      https://developers.google.com/tag-platform/tag-manager/web/datalayer
+    -->
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag() {
+        dataLayer.push(arguments);
+      }
+    </script>
+
```
atruskie: I think this should moved back into your embed embedGoogleAnalytics function - there's really no reason to have it split from that file. Also when used with my suggestion about preembedding tags you should be able to avoid the retry loop altogether

atruskie (changes requested):
Overall great job simplifying the bootstrap process.

The google analytics retry cycle is a bit hacky and I think you can improve on it

src/app/services/config/config.module.ts
Allcharles: This is an optional DI input. In this case, only SSR declares the API_CONFIG injection token as it does not need to make an api request for the config.

src/app/services/config/config.service.ts
Allcharles: I've changed this down to a 0 second delay, but I wonder if we still need to retry requests. I'm not sure if this will help any users

atruskie: well some delay is useful... 250ms maybe?

did you add the retry for a reason?

src/app/helpers/embedScript/embedGoogleAnalytics.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // Create Create dataLayer for google analytics:
+  // developers.google.com/tag-platform/tag-manager/web/datalayer
+  window["dataLayer"] = window["dataLayer"] || [];
+  window["gtag"] = function (...args: any) {
+    window["dataLayer"].push(args);
+  };
+
+  gtag("js", new Date());
+  gtag("config", key);
```
atruskie: potential race condition here.

i think the point of the custom gtag/datalayer manipulation is to have those structures predefined for when the script loads.

might as well define them, and then load the script... that way the script loading quickly (i.e. from cache) can't define gtag/datalayer before we do

atruskie (approved):

PR #1989, opened 2022-08-25
fixed create region api route bug

### LINKED ISSUE #1985: Site/Region creation failing with a 404
When a user creates a new _site_  (`region`) the request fails with 404 not found response.

It looks the server URL is not embedding the project id for some reason.

Image

Note the POST url for object creation is incorrect: `https://api.ecosounds.org/projects//regions/`

peichins:
# Title of PR

Fixed create region api route bug

## Changes

Changed the 2nd parameter on the region service create function call to use the project object rather than the (missing) project id property on the form model.

Also added a test for the value of this 2nd parameter. 

## Problems

Any problems caused by the PR that will need to be address in another PR. For example, these issues may be caused by another branch which will change how this branch functions.

## Issues

fixes #1985 

## Visual Changes

none

## Final Checklist

### DIFF
--- a/src/app/components/regions/pages/new/new.component.spec.ts
+++ b/src/app/components/regions/pages/new/new.component.spec.ts
@@ -88,9 +88,8 @@ describe("RegionsNewComponent", () => {
     it("should call api", () => {
       setup();
       api.create.and.callFake(() => new Subject());
-
       spectator.component.submit({});
-      expect(api.create).toHaveBeenCalled();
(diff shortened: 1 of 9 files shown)


atruskie (commented):
Looks like a nice neat change, however, I have questions:

1. Why is the project resolver not required in either of the `linkToRoute` functions - as it is in https://github.com/QutEcoacoustics/workbench-client/blob/ad1c1aff51831d12f5c767e100fe2c4a9ba2c087/src/app/components/sites/pages/new/new.component.ts#L77-L80
2. Did you test the shallow route?

PR #2003, opened 2022-11-09
Harvest mapping fix

### LINKED ISSUE #1975: Harvest: at some point all default mappings are overwritten
While processing a large harvest, I noticed that all default mappings were removed from a  harvest object.

I had a stale active record instance loaded in a terminal. You can see before (the scanning stage) and after (processing stage):

Image

Since we don't have a delete mappings function, we must assume that some client operation overwrote the entire mapping set.

Other details: 

- we didn't use any of the default mapping for this harvest.
- 4 new mappings were created for one folder each, all within another sub-folder
  e.g. sub_folder/site_1, sub_folder/site_2, etc...

hudson-newey:
# Harvest mapping fix

When creating Harvest mappings, mappings were updated through the value of GUI components. This caused harvest mappings to be removed if they were not rendered on the webpage.

## Changes

Harvest mappings are now defined using the Harvest model

## Problems

No problems

## Issues

Fixes #1975 

## Visual Changes

No visual changes

## Final Checklist

### DIFF
--- a/src/app/components/harvest/screens/metadata-review/metadata-review.component.spec.ts
+++ b/src/app/components/harvest/screens/metadata-review/metadata-review.component.spec.ts
@@ -1,9 +1,14 @@
 import { discardPeriodicTasks, fakeAsync, flush, tick } from "@angular/core/testing";
 import { MockBawApiModule } from "@baw-api/baw-apiMock.module";
+import { FolderRowComponent } from "@components/harvest/components/metadata-review/folder-row.component";
+import { WhitespaceComponent } from "@components/harvest/components/metadata-review/whitespace.component";
 import { ConfirmationComponent } from "@components/harvest/components/modal/confirmation.component";
+import { StatisticItemComponent } from "@components/harvest/components/shared/statistics/item.component";
 import { HarvestStagesService } from "@components/harvest/services/harvest-stages.service";
+import { StatisticsComponent } from "@components/harvest/components/shared/statistics/statistics.component";
 import {
   Harvest,
+  HarvestMapping,
   HarvestStatus
 } from "@models/Harvest";
 import { Project } from "@models/Project";
@@ -21,18 +26,38 @@ import {
 } from "@test/fakes/Project";
 import { MockProvider } from "ng-mocks";
 import { ToastrService } from "ngx-toastr";
+import { StatisticGroupComponent } from "@components/harvest/components/shared/statistics/group.component";
+import { HarvestItem } from "@models/HarvestItem";
+import { UTCOffsetSelectorComponent } from "@components/harvest/components/inputs/utc-offset-selector.component";
+import { SiteSelectorComponent } from "@components/harvest/components/inputs/site-selector.component";
+import { menuRoute } from "@interfaces/menusInterfaces";
+import { generateMenuRoute } from "@test/fakes/MenuItem";
+import { SHALLOW_HARVEST } from "@baw-api/ServiceTokens";
+import { ShallowHarvestsService } from "@baw-api/harvest/harvest.service";
+import { generateHarvestItem } from "@test/fakes/HarvestItem";
+import { Inject } from "@angular/core";
 import { MetadataReviewComponent } from "./metadata-review.component";
 
-describe("MetaDatReviewComponent", () => {
+describe("MetadataReviewComponent", () => {
   let spec: SpectatorRouting<MetadataReviewComponent>;
   let modalService: NgbModal;
   let modalConfigService: NgbModalConfig;
   let stages: SpyObject<HarvestStagesService>;
+  let harvestService: SpyObject<ShallowHarvestsService>;
   let defaultProject: Project;
   let defaultHarvest: Harvest;
 
   const createComponent = createRoutingFactory({
-    declarations: [ConfirmationComponent],
+    declarations: [
+      ConfirmationComponent,
+      StatisticsComponent,
+      StatisticGroupComponent,
+      StatisticItemComponent,
+      WhitespaceComponent,
+      SiteSelectorComponent,
+      UTCOffsetSelectorComponent,
+      FolderRowComponent
+    ],
     component: MetadataReviewComponent,
     providers: [
       MockProvider(HarvestStagesService, {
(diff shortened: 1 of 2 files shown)


src/app/components/harvest/components/metadata-review/folder-row.component.ts
```diff
@@ -165,16 +166,19 @@ export class FolderRowComponent {
...
   }
 
   public setOffset(mapping: HarvestMapping, offset: string): void {
+    if ( !isInstantiated(offset) ) { return; }
     mapping.utcOffset = offset;
     this.mappingsChange.emit();
   }
 
   public setIsRecursive(mapping: HarvestMapping, isRecursive: boolean): void {
+    if ( !isInstantiated(isRecursive) ) { return; }
```
atruskie: inconsistent formatting, and also why are we doing these bail outs? How does that fix the bug?

src/app/components/harvest/screens/metadata-review/metadata-review.component.ts
```diff
@@ -248,22 +248,29 @@ export class MetadataReviewComponent
...
+    * this, however it's likely less performant
+    */
     this.rows.forEach((row): void => {
       if (row.rowType === RowType.folder && isInstantiated(row.mapping)) {
-        mappings.push(row.mapping);
+
+        if (this.harvest.mappings.find(mapping => mapping.path === row.mapping.path)) {
+          // since the path attribute is a unique identifier on a mapping, we can safely assume
+          // that there will ever only be one mapping returned from the bellow filter
+          Object.assign(this.harvest.mappings.filter(mapping => mapping.path === row.mapping.path)[0], row.mapping);
```
atruskie: two passes are redundant here, find the object first, and then update it by reference second. Don't need to do `.filter` again

src/app/components/harvest/screens/metadata-review/metadata-review.component.ts
```diff
@@ -248,22 +248,29 @@ export class MetadataReviewComponent
...
+          this.harvest.mappings.push(row.mapping);
+        }
+
       }
     });
 
     this.hasUnsavedChanges = true;
     this.harvestApi
-      .updateMappings(this.harvest, mappings)
+      .updateMappings(this.harvest, this.harvest.mappings)
```
atruskie: i'm usually all for storing information on a model, but in this case it's temporary. 

I think keep the constructed form separate, and then update the harvest mappings after you get the response back from the server (so we have authoritative mappings)

Actually I'm undecided on this. You can keep it on the model, but it should still be updated with the response from the server afterwards

atruskie (changes requested):
Tests! I need to see tests!

src/app/components/harvest/screens/metadata-review/metadata-review.component.ts
```diff
@@ -248,22 +248,29 @@ export class MetadataReviewComponent
...
-
     /*
-     * Iterate through all the rows, extract folders which have mappings and
-     * push that list to the mappings array. We could use a filter and map for
-     * this, however it's likely less performant
-     */
+    * Iterate through all the rows, extract folders which have mappings and
+    * push that list to the mappings array. We could use a filter and map for
+    * this, however it's likely less performant
+    */
```
atruskie: JSDoc indentation is incorrect - are you using your linter/formatter?

src/app/components/harvest/components/metadata-review/folder-row.component.ts
hudson-newey: I remember this fixing another issue when editing a harvest mapping, but this seems to been resolved when I changed the implementation of the harvestAPI

hudson-newey:
I'm going to be polishing the code quality and subsequently squashing before considering review/merging this. At the moment, it is a working MVP that is sure to have an unidentified bug.

I am going to be:
1. Fetching elements from the DOM through there inner text / class names (not index) ✅
2. Refactoring the code in the image attached below ✅
3. Adding code comments ✅
4. Consolidating helper functions ✅
5. Removing artefacts from testing / development ✅

e.g.
image

src/app/components/harvest/screens/metadata-review/metadata-review.component.ts
```diff
@@ -248,27 +248,41 @@ export class MetadataReviewComponent
...
+        // since the path attribute is a unique identifier on a mapping, we can safely assume
+        // that there will ever only be one mapping returned from the bellow filter
+        const newMapping = newMappings.filter(mapping => mapping.path === row.mapping.path)[0];
+        if (isInstantiated(newMapping)) {
+          Object.assign(newMapping, row.mapping);
+        } else {
+          // the mapping doesn't exist on the current harvest model, so create a new one
+          newMappings.push(row.mapping);
+        }
+
```
atruskie: i just realised this would be a *lot* simpler if we used a `Map<string, HarvestMapping>` as the temporary collection for `newMappings`

atruskie (approved):
OK, great, well tested, neat.

However, I'd like to see another test added that tests the other scenario. When the harvest is created, we create default mappings for each site. Any of those mapping that weren't used (i.e. not shown in the UI) were not preserved. Let's add a test case for this. 

- Add 3 default mappings (/a, /b, /c)
- Add one harvest items results set (/b)
- modify this mapping (set change utc offset)
- Test that updateMappings emits mappings for /a, /b (modified, and /c

src/app/components/harvest/screens/metadata-review/metadata-review.component.ts
hudson-newey: This seems to be an old commit
I've since fixed the problem in the most recent commit

src/app/components/harvest/screens/metadata-review/metadata-review.component.spec.ts
```diff
@@ -120,4 +182,172 @@ describe("MetaDatReviewComponent", () => {
...
+        path: "C/ca",
+        recursive: false,
+        siteId: null,
+        utcOffset: "-08:00"
+      })
+    ];
+
+    // when harvest item rows are organically generated, they will always include an injector services
+    // as part of the association decorator, it requires this injector service, therefore, we have to add one manually
+    Object.defineProperty(mapping[1], "injector", { value: spec.component });
```
hudson-newey: This is obviously not ideal, I'm working on a more correct solution to the injector requirement for
https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/models/AssociationDecorators.ts#L179
>
https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/harvest/components/metadata-review/folder-row.component.ts#L56

atruskie (approved):
Looks great. You left a `f` on an `it` but otherwise, let's ship it!

Issue #2001, opened 2022-11-07
Site/point location setting by point&click does not appear to be working


atruskie:
**Describe the bug**

When I try and set a site or point location by clicking on the embedded map, it does not move the marker or update the coordinates.

**To Reproduce**

1. Go to the edit/create page of any site/point
2. Click on the map to set the location
4. Note the form/marker are not updated

**Expected behavior**

For location selection by point and click of the map to work.

**Screenshots**

N/A

**Desktop (please complete the following information):**

- OS: Windows
- Browser: Edge/Chrome
- Version: latest

**Smartphone (please complete the following information):**

N/A

**Additional context**

Also noticed by users.

PR #2014, opened 2022-12-06
Added ability to edit site position through map

hudson-newey:
# Added ability to edit site position through map

The ability to change the position of a site through the embedded Google map was not functional. This PR adds the functionality required to edit the location of a site using the map provided in the `location-input.component.ts`

## Changes

* Made marker attributes an optional `@Input()` / attribute of the `baw-map`
* Modified `updateModel()` method in `location-input.component.ts` to optionally take a new longitude and latitude
* Made the `baw-map` markers emit an event every time the marker is moved
* Created unit tests for changes made to the `location-input.component.ts`

## Problems
N/A

## Issues

#2001 

## Visual Changes
N/A

## Final Checklist

### DIFF
--- a/src/app/components/shared/formly/location-input.component.spec.ts
+++ b/src/app/components/shared/formly/location-input.component.spec.ts
@@ -1,1 +1,176 @@
-// TODO
+import { fakeAsync, tick } from "@angular/core/testing";
+import {
+  FormControl,
+  FormGroup,
+  FormsModule,
+  ReactiveFormsModule,
+} from "@angular/forms";
+import { GoogleMapsModule } from "@angular/google-maps";
+import { embedGoogleMaps } from "@helpers/embedScript/embedGoogleMaps";
+import { createHostFactory, SpectatorHost } from "@ngneat/spectator";
+import { FormlyBootstrapModule } from "@ngx-formly/bootstrap";
+import { FormlyFieldProps, FormlyModule } from "@ngx-formly/core";
+import { MapComponent } from "@shared/map/map.component";
+import { modelData } from "@test/helpers/faker";
+import { MockComponent, MockedComponent } from "ng-mocks";
+import { formlyConfig } from "./custom-inputs.module";
+import { LocationInputComponent } from "./location-input.component";
+
+const mockMapComponent = MockComponent(MapComponent);
+
+describe("FormlyLocationInput", () => {
+  let model: any;
+  let formGroup: FormGroup;
+  let spectator: SpectatorHost<LocationInputComponent>;
+
+  const createHost = createHostFactory({
+    declarations: [mockMapComponent],
+    component: LocationInputComponent,
+    imports: [
+      FormsModule,
+      ReactiveFormsModule,
+      FormlyModule.forRoot(formlyConfig),
+      FormlyBootstrapModule,
+      GoogleMapsModule,
+    ],
+  });
+
+  function setup(key: string = "input", options: FormlyFieldProps = {}) {
+    formGroup = new FormGroup({ input: new FormControl("") });
+    model = {};
+
+    spectator = createHost(
+      `
+      <form [formGroup]="formGroup">
+        <baw-location-input></baw-location-input>
+      </form>
+      `,
+      {
+        hostProps: { formGroup },
+        props: {
+          field: {
+            model,
+            key,
+            formControl: formGroup.get("input"),
+            props: options,
+          },
+        },
+      }
+    );
+    spectator.detectChanges();
+
+    return spectator;
+  }
+
+  beforeAll(async () => await embedGoogleMaps());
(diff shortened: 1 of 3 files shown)


src/app/components/shared/formly/location-input.component.spec.ts
```diff
@@ -1 +1,170 @@
...
+    latitudeInput.dispatchEvent(new Event("input"));
+
+    spectator.detectChanges();
+  }
+
+  /**
+   * Changes the markers position by directly editing the model.
+   * This is used to initialize the marker to the correct position, when the input field might not work
+   */
+  function explicitlySetMarket(longitude: number, latitude: number) {
```
atruskie: spelling mistake `Market` should be `Marker`

src/app/components/shared/formly/location-input.component.spec.ts
```diff
@@ -1 +1,170 @@
...
+      domEvent: new Event("mapDragend"),
+      latLng: {
+        lng: () => updatedLongitude,
+        lat: () => updatedLatitude,
+        equals: () => null,
+        toJSON: () => null,
+        toUrlValue: () => null
+      },
+      stop: () => {}
+    });
```
atruskie: I wonder if we can't create a slightly more true to life test. 

Assuming the marker is centered on the map, we should be able to click on the map (say at 0.1 .0.1) and see updated coordinates flow through.

atruskie (changes requested):
Great work.

Question: I see you bound to draggable, but what about just point and click the move the marker? It maybe that dragend caters for point and click as well a drag, but I'm not sure.

src/app/components/shared/formly/location-input.component.spec.ts
hudson-newey: Would we do this by sending out a click event?

I don't believe there's a way to click on an element at a certain position using `HTMLElement.click()`
[https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click)

hudson-newey:
At the moment, the user has to drag the marker for it to move

It shouldn't be to hard to add a click event to the map, that triggers the marker to move

I'll have a look at this now

src/app/components/shared/formly/location-input.component.spec.ts
hudson-newey: To query the selector for an interactive surface, and sub-subsequently click on it, I'll first need to render the map component (`baw-map`). However, this is impossible without removing existing safeguards that are present to prevent the embedded Google map from rendering in a testing environment

[source](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/helpers/embedScript/embedGoogleMaps.ts#L14-L18)
image

I'm assuming that removing this bail-out safeguard is not the correct way forward?

atruskie: hmm let's try and find out why rendering googlemaps is bad in tests?

atruskie: Otherwise yeah, let's go with the method you had, where you simulate an event fired by a maps component

hudson-newey: I can't seem to find a reference to why it was implemented in PR #1972 [commit](https://github.com/QutEcoacoustics/workbench-client/commit/07e18bc639888b4a347fd84dc951463a2d525fea#diff-f558660b3435ca693715d2962016a8ffb2f0f267167bc9c04c5974b5426e9637R14-R18)

My assumption is that we don't want to make an API call (with our key) every time we run tests as we:
- Don't want to bombard the Google Maps API with 10+ requests in fast succession when running unit tests
- Don't want to have unit tests potentially failing because of a. a bad internet connection, or b. because the maps API is unreachable. (Our unit tests don't want to be testing if Google is alive)
- We pay per request

[Similar (but not very friendly) response to a similar question on SO](https://stackoverflow.com/questions/13657965/angularjs-unit-testing-application-based-on-google-maps-api#comment33737860_13657965)

atruskie: ok, in dev, record the event, and in tests simulate the firing, that should be enough

it still doesn't really test that we've bound to google's click handler but it's close enough

hudson-newey: I've left a JsDoc comment stating the need for manual testing

https://github.com/QutEcoacoustics/workbench-client/pull/2014/files#diff-ec8b1ef587154c95ecb3ba88fb4d1296a839d2fe11521acdb88174dffb0a7d97R11-R15

I've reverted the tests to be in line with the event firing

Pending approval

atruskie (approved):

PR #2016, opened 2022-12-09
Analyses results view

### LINKED ISSUE #2011: Start to sketch out a UI for downloading the results

hudson-newey:
# Default Analyses results view

Downloading the default analyses from audio recordings

## Changes

* Created analysis download row components
* Creates new routes for analyses downloads
* Creates analysis results component

## Problems

Download functionality for directories in the form of ZIP files needs to be completed as part of another ticket

## Issues

Fixes: #2011 

## Visual Changes

image

## Final Checklist


hudson-newey:
Todo:
- [x] Unit tests for opening and closing a folder in Analysis Results view
- [x] Loading rows through be done asynchronously using an observable [code](https://github.com/QutEcoacoustics/workbench-client/pull/2016/files#diff-10245471f27ac2508c142cc03366bf0ce489c9bfb1483ccf933ac7aa2600d145R34-R47) (at the moment it's rendering opening a folder by [code](https://github.com/QutEcoacoustics/workbench-client/pull/2016/files#diff-6470e1192550c02635caff1dd6dd0daee2914e46b4d522bd473daf6472127614R49-R55), this is obviously not ideal)
- [x] Method for converting large bytes to readable format (KB, MB, GB, etc...) needs to either a. find a library or b. be refactored [code](https://github.com/QutEcoacoustics/workbench-client/pull/2016/files#diff-12c5e746c47cc65997e14d81ccd64b708501f3e1b3e0764eb01db435125db5abR62-R73)
- [x] Link view to real data, (instead of mock / predetermined data) [code](https://github.com/QutEcoacoustics/workbench-client/pull/2016/files#diff-890f7b803e09ccb11110421b6dde467ee1dc013de1075d2f1c0c57f0544ae7a1R28-R86)
- [x] Bind `analyses-results` component to `AnalysisResults` model
- [x] Refactor & clean up code quality
- [x] Investigate why `faker.js` is causing bundle to be so large (causing CI to fail)
- [x] Change `getItems()` method in `analyses-results.component.ts` to an async/await method. This should fix this race conditions & unit tests

hudson-newey:
I'm currently observing a race condition.
It is originating from [this](https://github.com/QutEcoacoustics/workbench-client/pull/2016/files#diff-10245471f27ac2508c142cc03366bf0ce489c9bfb1483ccf933ac7aa2600d145R73-R85) `getItems()` method in `analyses-results.component.ts`. To fix this, I am going to be changing `getItems()` to an async/await method, and am going to change the `api.list()` to `api.show()`. This is the last remaining task, everything else should be all done and ready for review

src/app/components/audio-recordings/audio-recording.menus.ts
```diff
@@ -64,6 +65,18 @@ function makeBatchMenuItem(subRoute: RecordingRoute): MenuRoute {
   });
 }
 
+export function makeAnalysesMenuItem(subRoute: RecordingRoute): MenuRoute {
+  return menuRoute({
+    icon: ["fas", "file-arrow-down"],
```
atruskie: probably the wrong icon - the route goes to a view where a user could then download a files

probably want an icon representing a set of files, or a folder of files

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /**
+   * Fetches the `AnalysisJobItemResult` model from the baw-api
+   *
+   * @param item The id of the model that is requested from the baw-api
+   * @returns The complete `AnalysisJobItemResult` model of the requested item. If the item is not defined, the root path will be returned
+   */
+  public getItems(item?: AnalysisJobItemResult): AnalysisJobItemResult[] {
+    const analysisJobId = this.analysisJob;
```
hudson-newey: This should be `analysisJob`

src/app/components/audio-recordings/audio-recording.menus.ts
```diff
@@ -64,6 +65,18 @@ function makeBatchMenuItem(subRoute: RecordingRoute): MenuRoute {
...
+    label: "Download Analyses",
+    tooltip: () => "Download audio recording analyses",
```
atruskie: Use the phrasing "download analysis results"

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,62 @@
...
+    [ngStyle]="{ backgroundColor: even ? '#f2f2f2' : 'white' }"
+  >
+    <span class="grid-table-item" (click)="toggleOpen()" style="cursor: pointer">
+      <!-- indentation of files in tree hierarchy -->
+      <span class="whitespace-indentation" *ngFor="let i of indentation()">
+        <baw-directory-whitespace
+          [isFolder]="isFolder"
+          [open]="open"
+        ></baw-directory-whitespace>
+      </span>
```
atruskie: how about our custom directive accepts an indentation and just uses a single element with variable width?

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+          item: defaultAnalysisJobItemResult
+        }
+      }
+    );
+
+    spectator.component.item = defaultAnalysisJobItemResult;
+    spectator.detectChanges();
+  }
+
+  beforeEach(() => setup());
```
atruskie: incomplete tests?

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  public get downloadUrl(): string {
+    // the root folder does not have a parent item, and therefore the parent item's path is optional
+    return `${this.apiRoot}${this.item.parentItem?.path ?? ""}/${this.item.name}`;
+  }
+
+  /**
+   * Returns the path of the analysis result item, relative to the root path
+   */
+   private get relativePath() {
+    return `${this.item.parentItem.path}/${this.item.name}`;
```
atruskie: delegate these routes to the model or service. Should be using a strong route too?

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Calculates how much indentation a certain folder needs
+   *
+   * @returns an empty array of length `n`, representing how many nested sub folders the item is under
+   */
+  protected indentation(): Array<void> {
+    // some files don't have a path explicitly defined, so we need to use the name and parent item's path in these cases
+    const subPaths = this.subDirectoriesCount(
+      isInstantiated(this.item.path) ? this.item.path : this.relativePath
```
atruskie: this seems like a smell; explain this to me?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.html
```diff
@@ -0,0 +1,32 @@
...
+  </h1>
+
+  <p>
+    Results are available as they are generated. Download analysis reports of
+    individual files by clicking the <i>Download</i> button, or
+    <i>Download ZIP</i> to download an entire folders results in the form of a
+    ZIP file.
+  </p>
+
+  <!-- my code goes here -->
```
atruskie: redundant comment?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.scss
```diff
@@ -0,0 +1,30 @@
...
+  background-color: white;
+
+  .grid-table-header {
+    display: contents;
+  }
+}
+
+.grid-table-wider {
+  border-bottom: $container-borders;
+}
```
atruskie: there's a lot of overloaded definitions for the terms grid and table. Let's try and use some clearer names that better explain the purposes on these classes. As far as I can see you're not even using a table, so I don't see why the phrase table makes any sense.

e.g. `.directory-listing-container`

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
```diff
@@ -0,0 +1,38 @@
...
+    1fr
+    14rem
+    9rem
+    min-content
+    min-content;
+
+  .grid-table-row {
+    display: contents;
+  }
+}
```
atruskie: i commented further down on naming, applies here as well

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.btn-border {
+  border-style: solid;
+  border-color: $border-color;
+}
```
atruskie: seems sketch we're defining custom styles for bootstrap things... is there not a class that does this?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+          children: [
+            new AnalysisJobItemResult(
+              generateAnalysisJobResults({
+                path: "result_item.csv",
+              })
+            ),
+          ],
+        })
+      ),
+    ];
```
atruskie: a lot of boiler plate here; seems like defining a helper function might be useful

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    mockDirectoryStructure(singleFolderDirectory);
+
+    // get /FolderA and asset that it has the greyed out download zip button
```
atruskie:
```suggestion
    // get /FolderA and assert that it has the greyed out download zip button
```

Further usages below to correct

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    spectator.detectChanges();
+  }
+
+  beforeEach(() => setup());
+
+  const getDirectoryRow = (index: number): HTMLElement =>
+    spectator.queryAll<HTMLElement>("baw-directory-row")[index];
+
+  const getDownloadButton = (index: number): HTMLButtonElement =>
+    getDirectoryRow(index).getElementsByTagName("button")[0];
```
atruskie: indexing is super fragile; perhaps there's a better way to achieve this

atruskie: i'd also say your surrounding functions are fragile

there must be a better way to do this. Maybe: https://stackoverflow.com/a/63049011/224512

this looks like overkill but might be useful: https://testing-library.com/docs/angular-testing-library/intro

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  public updateRows = (): Observable<AnalysisJobItemResult[]> =>
+    this.rows$ = this.getRows();
+
+  public getRows() {
+    const data = new Observable<AnalysisJobItemResult[]>((observer) => {
+      observer.next(this.rows);
+    });
+    return data;
```
atruskie: why are you creating a new observable each time?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+        // this prevents false positives e.g. if the folder name is contained within another folders name
+        this.rows = this.rows.filter(item =>
+          // only the root item will not have a parent item, in this case, we can quickly determine that all folders need to be closed
+          !isInstantiated(item.parentItem) ||
+          !item.parentItem.path.includes(`/${model.name}/`) ||
+          item.name === model.name
+        );
+      } else {
+        this.rows.splice(this.rows.indexOf(model) + 1, 0, ...childItems);
+      }
```
atruskie: this is pretty messy. I wonder if you could do something simpler like track the parent model for each row. Then when closing the folder it is simply: remove any item that has parent == model (the row clicked).

Also, there's an asymmetry here: one branch modifies the collection in place, the other reassigns a new collection. Be consistent.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+      this.addParentInformation(childItems, model);
+
+      if (model.open) {
+        // the relative path for all sub folders should include the folders name & two back slashes before and after the name
+        // this prevents false positives e.g. if the folder name is contained within another folders name
+        this.rows = this.rows.filter(item =>
+          // only the root item will not have a parent item, in this case, we can quickly determine that all folders need to be closed
+          !isInstantiated(item.parentItem) ||
+          !item.parentItem.path.includes(`/${model.name}/`) ||
+          item.name === model.name
```
atruskie: i actually don't understand what this branch is doing, can we only have one directory open at a time?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+        );
+      } else {
+        this.rows.splice(this.rows.indexOf(model) + 1, 0, ...childItems);
+      }
+
+      model.open = !model.open;
+      this.updateRows();
+    });
+  }
```
atruskie: take another simplicity pass at this method. I read through it nearly 3 times and my understanding changed on each read. It may be correct, but it isn't doing the reader any favours

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  // not all analysisJobItemResult models from the baw-api contain a path attribute therefore, we can not know the location of the item
+  // therefore, by adding information about the parent object, we can easily derive the relative path and hierarchy of a sub item
+  private addParentInformation(models: AnalysisJobItemResult[], parent: AnalysisJobItemResult) {
+    models.forEach(analysisJobItemResultItem => {
+      analysisJobItemResultItem.parentItem = parent;
+    });
+  }
```
atruskie: I need you to double check with me about this path thing

atruskie: also parent linking should be done in the service or model, not a random component

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * Sub directories of `AnalysisJobItemResults` do not contain all the information needed e.g. How many child items the sub directory has.
+   * Therefore, we need to evaluate the child sub directory items explicitly and use their `AnalysisJobItemResult` model.
```
atruskie: ok major assumption here

- why do you need counts at all?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+    parent: AnalysisJobItemResult
+  ): Promise<AnalysisJobItemResult[]> {
+    const evaluatedSubItems = Array<AnalysisJobItemResult>();
+
+    // file elements do not have to be evaluated, so we can return the child items as specified in the children attribute
+    children.forEach(childItem => {
+      if (childItem.type === "file") {
+        evaluatedSubItems.push(childItem);
+      }
+    });
```
atruskie: seems like you could just roll this skip into your loop

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+   */
+  private async getModelChildren(
+    children: AnalysisJobItemResult[],
+    parent: AnalysisJobItemResult
+  ): Promise<AnalysisJobItemResult[]> {
+    const evaluatedSubItems = Array<AnalysisJobItemResult>();
+
+    // file elements do not have to be evaluated, so we can return the child items as specified in the children attribute
+    children.forEach(childItem => {
+      if (childItem.type === "file") {
```
atruskie: i'd define helpers on the models like `isFile` and `isDirectory`

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
atruskie: it looks like this method is doing n+1 requests for every directory. This shouldn't be happening

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,185 @@
...
+    resolvers: {
+      [audioRecordingKey]: audioRecordingResolvers.show,
+      [projectKey]: projectResolvers.showOptional,
+      [regionKey]: regionResolvers.showOptional,
+      [siteKey]: siteResolvers.showOptional,
+    },
+  };
+}
+
+AnalysesResultsComponent.linkToRoute(getPageInfo("base"))
```
atruskie:
```suggestion
AnalysesResultsComponent
  .linkToRoute(getPageInfo("base"))
```

src/app/models/AnalysisJobItemResult.ts
```diff
@@ -19,6 +19,8 @@ export interface IAnalysisJobItemResult {
   hasZip?: boolean;
   type?: ResultsItemType;
   children?: AnalysisJobItemResult[];
+  open?: boolean;
+  parentItem?: AnalysisJobItemResult;
```
atruskie: these don't belong on the interface. We use the interface to represent loosely typed objects returned from the services (after deserialization but before model construction).

Calculated properties (client side only) don't have a reason to be on the interface.

src/app/models/AnalysisJobItemResult.ts
```diff
@@ -44,6 +46,8 @@ export class AnalysisJobItemResult
   public readonly hasZip?: boolean;
   public readonly type?: ResultsItemType = "directory";
   public readonly children?: AnalysisJobItemResult[];
+  public parentItem?: AnalysisJobItemResult;
+  public open?: boolean;
```
atruskie: These properties represent view model properties. They have nothing to do with the model but rather how the model is presented. They should be in a different class.

src/app/models/AudioRecording.ts
```diff
@@ -141,6 +142,15 @@ export class AudioRecording
...
 
+  public getAnalysisResultsUrl(
+    project?: IdOr<Project>,
+    region?: IdOr<Region>,
+    site?: IdOr<Site>
+  ): string {
+    const routes = audioRecordingResultsRoutes;
+    return this.selectRoute(routes, project, region, site);
+  }
+
```
atruskie, line 156: missing tests for this function

atruskie (changes requested):
I think it's important you discuss with me API deficiencies before you code around them. Maybe it is something that can be fixed, maybe your understanding is incorrect. 

Why are you using promises when observables are supported by the async pipe as well?

Okay looks promising but I'd say another solid pass is needed.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: The counts are needed to display the number of children in a directory (the little number 3 in the example below).

image

If we were to remove this requirement, it would drastically reduce the complexity of the code, and resolve about a third of the comments on this review.

hudson-newey:
I know this is a big PR, so thanks for the review. I'm going through the requested changes now and will get back to you asap

src/app/components/audio-recordings/audio-recording.menus.ts
hudson-newey: Idealy, [this](https://fontawesome.com/icons/files?s=solid&f=classic) icon would be the perfect match. However, without the Font Awesome "pro plan" we cannot use this icon.

[This](https://fontawesome.com/icons/folder-tree?s=solid&f=classic) icon seems to be a perfect middle ground between the two

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
hudson-newey: For the row component, I'll be able to add tests to assert that
- The file name is displayed correctly
- A download button is present / disabled for folders

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
atruskie: was this ever a stated requirement? Consider it removed

But the solution here is to ask for an API patch, not write hugely inefficient code

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
hudson-newey: The path attribute is only returned by the API for folders [see here](https://api.staging.ecosounds.org/analysis_jobs/system/results/1).

However, `this.relativePath` will always resolve to the same value as `this.item.path`, so it's more correct just to use `this.relativePath`. I will be changing this in the upcoming commit.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.scss
hudson-newey: I've fixed this in an upcoming commit using the names
- `directory-listing-container`
- `directory-column-header`
- `directory-listing-item`

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
atruskie: Again, correct courses of action:

- file an issue against the API
- mock the property in the model/view model

But the cases where you need these paths should be different. I'm still not convinced you need to backfill this property.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
hudson-newey: Good catch, [The Bootstrap borders](https://getbootstrap.com/docs/4.3/utilities/borders/) seem to fit the requirements

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: I originally had something similar, but collapsing sub folders under sub folders does not work with the remove if `parent === model` condition.

e.g.
A/
 aa/
  file.csv

Collapse folder `A/`, and folder `aa/` will be removed, but not `file.csv`. To fix this I originally had an array of parent items. This won't be to hard to reverse, but I originally modified it in order to conserve memory & improve performance.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: You are able to have multiple folders open at the same time, but only the root folder will not have a `parentItem`. So therefore if the `parentItem` attribute does not exist on the clicked item, all filter conditions will fail, causing all folders to collapse.

The second condition removes all items that are sub items of the directory that was clicked on. This is possible without a DFS (or equivalent search) due to how POSIX complaint file structures work. Because files & directories can't have backslashes, we can concur that only sub directories will include the path pattern `/${model.name}/`

The third condition is needed so that the original file is preserved (since the `includes`) condition will also match against the directory that the user clicked on.

hudson-newey: I'll be changing the functionality in a future commit, because I do agree that these chained conditions can be hard to keep track of

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: As the parent information was only used to derive the path of the `AnalysisJobItemResult` items, this method is no longer needed as I am now appending the `path` information to `AnalysisJobItemResult` item models.

This will greatly simplify the solution and cut ~30 LLOC

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: Absolutely, this was in my original refactor todo list, must have overlooked this

hudson-newey: This overarching method will be removed as part of [this request](https://github.com/QutEcoacoustics/workbench-client/pull/2016#discussion_r1050244083)

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
atruskie: so the trick there is to link your view models together so you can recurse up the chain of parents

hudson-newey: Easy enough, I'm 90% sure I had a historical commit that worked this way.

I've also come to realize that this implementation through a filter is also buggy at best.

E.g. a file structure such as

/
  A/
    A/aa/
  B/
    B/A

Would break this filter

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
hudson-newey: It would definitely be best practice to configure a strong route for the API. However, at the moment, I don't think this is a thing. e.g.
[code example 1](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/models/AudioRecording.ts#L130-L132)
[code example 2](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/models/AttributeDecorators.ts#L105-L111)

This shouldn't be to hard to implement, as the strong route code looks nice, and has an existing propertys named `routePath` and `routeRoot`

atruskie: yeah but it should be

- the service is the right place for knowledge about API routes
- the .routes files are good places for knowledge about client routes

hudson-newey: That's a good point that I didn't think of. I'll delegate the `downloadUrl()` method to the services

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: This has been resolved via [this code](https://github.com/QutEcoacoustics/workbench-client/pull/2016/files#diff-10245471f27ac2508c142cc03366bf0ce489c9bfb1483ccf933ac7aa2600d145R80-R90)

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,50 @@
+<div class="directory-listing-container">
+  <div
+    class="display-contents"
+    [ngStyle]="{ backgroundColor: even ? '#f2f2f2' : 'white' }"
```
atruskie: these should be classes applied to the row so we can adjust the color in the css

the colour values should be css variables and part of our theme

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,50 @@
...
+  >
+    <span class="directory-listing-item" (click)="toggleOpen()" style="cursor: pointer">
```
atruskie: again, leave styling to css, move that cursor definition

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,50 @@
...
+      <!-- indentation of files in tree hierarchy -->
+      <span class="whitespace-indentation" *ngFor="let i of indentation()">
+        <span class="whitespace-vertical-line"></span>
+        <div class="whitespace-block"></div>
```
atruskie: these can be collapsed, just add the width of the `whitespace-block` as `right-margin` to the `whitespace-vertical-line`

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,50 @@
...
+        <span *ngIf="isFolder && open" class="whitespace-vertical-half-line"></span>
+      </span>
+
+      <fa-layers class="fa-custom-icon m-2" [fixedWidth]="true">
+        <fa-icon
+          class="fa-custom-icon"
+          [icon]="[
+            'fas',
+            !isFolder ? 'file' : open ? 'folder-open' : 'folder-closed'
```
atruskie: either parenthesize the nested conditional or make this a function on the component (probably the better choice) 

`[icon]="chooseIcon('file', 'folder-open', 'folder-closed')"`

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,50 @@
...
+        ></fa-icon>
+      </fa-layers>
+
+      {{ itemName }}
+    </span>
+
+    <span class="directory-listing-item">
+      <ng-container *ngIf="!isFolder">
+        {{ item.sizeBytes }} Bytes
+      </ng-container>
```
atruskie: this is show directly to the user? Not in a tooltip? It should probably be humanized then

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
```diff
@@ -0,0 +1,50 @@
...
+        <a
+          class="btn btn-light btn-sm mb-2 mt-1 border"
+          [class.disabled]="isFolder"
+          [href]="downloadUrl"
+        >
+          <fa-icon
+            [icon]="['fas', isFolder ? 'file-archive' : 'file-arrow-down']"
+          ></fa-icon>
+        </a>
+      </span>
```
atruskie: this might look easier to read as two separate templates, one for the folder, one for the files

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
```diff
@@ -0,0 +1,49 @@
+$vertical-line-color: grey;
```
atruskie: prefer css variables for theming - pick them or define them from out central theme file

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
```diff
@@ -0,0 +1,49 @@
...
+$margin-offset: 13px;
```
atruskie: avoid using pixel units; use em units so spacing scales with font-size

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.directory-listing-container {
+  display: grid;
+  grid-template-columns:
+    1fr
+    14rem
+    9rem
```
atruskie: i reckon we can do better than hard coding these widths. Does the css grid spec allow for something like that? `auto-fit`  or `min-content` maybe?

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      generateAnalysisJobResults({
+        name: expectedFileName
+      })
+    );
+
+    setup();
+
+    const realizedFileName = getFileName();
+
+    expect(realizedFileName).toEqual(` ${expectedFileName} `);
```
atruskie: why is there white space around the filename?

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    new AnalysisJobItemResult(generateAnalysisJobResults());
+
+  const getDirectoryRow = (): HTMLSpanElement =>
+    spectator.query<HTMLSpanElement>(".directory-listing-item");
+
+  const getDirectoryRowDownloadButton = (): HTMLAnchorElement =>
+    spectator.query<HTMLAnchorElement>("a");
+
+  const getFileName = (): string =>
+    getDirectoryRow().textContent;
```
atruskie: perhaps innerText is better: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    const realizedFileName = getFileName();
+
+    expect(realizedFileName).toEqual(` ${expectedFileName} `);
+  });
+
+  it("should display a download button next to file", () => {
+    setup();
+    const downloadButtonElement = getDirectoryRowDownloadButton();
+    expect(downloadButtonElement).toBeTruthy();
+  });
```
atruskie: missing a test for download button being disabled for folders

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  ) { }
+
+  public open: boolean;
+  @Input() public item: AnalysisJobItemResult;
+  @Input() public analysisJob: AnalysisJob;
+  @Input() public audioRecording: AudioRecording;
+  @Input() public even: boolean;
+  @Output() public loadChildren = new EventEmitter<AnalysisJobItemResult>();
+
+  private subDirectoriesCount = (path: string) => path.split("/").length;
```
atruskie: why not make this a normal function? why is it a function value?

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public get isFolder(): boolean {
+    return this.item.type === "directory";
+  }
+
+  public get isFile(): boolean {
+    return this.item.type === "file";
+  }
+
```
atruskie: i'd personally add these functions to the model... no need to move for now, unless they're used elsewhere.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.html
```diff
@@ -0,0 +1,31 @@
...
+  </p>
+
+  <div class="directory-listing-container">
+      <div class="directory-column-header">
+        <span>Path</span>
+        <span>Size</span>
+        <span>Download Results</span>
+      </div>
+    <div
+      *ngFor="let row of rows$ | async; let even = even; trackBy: trackByRow"
```
atruskie: do you actually need to rebind even ?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.scss
```diff
@@ -0,0 +1,27 @@
...
+  grid-template-columns:
+    1fr
+    14rem
+    9rem
+    min-content
+    min-content;
+  grid-template-rows: 1fr;
+  border-bottom: $container-borders;
+  background-color: white;
+}
```
atruskie: this grid size definition is essentially repeated from src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss - remove the deduplication

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+  public audioRecording: AudioRecording = this.routeData[audioRecordingKey]?.model;
+  private rows = Array<ResultNode>();
+  private rootItemPath: string;
+
+  public get analysisJob(): AnalysisJob {
+    // since this component is currently only used for default analysis jobs, we can hard code the analysis job
+    return new AnalysisJob({
+      id: "system",
+      name: "system",
+    });
```
atruskie: Make this a static property on the `AnalysisJob` class

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
```
atruskie: this should be an input for the component; specifically, since this is a page, it should be resolved via route data.

This view can't be mounted for any other analysis job currently.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  public ngOnInit() {
+    this.rootItemPath = `${rootPath}${this.audioRecording.id}/`;
```
atruskie: this component should not know anything about paths/routes.

Create a function in the analysis jobs service that creates api routes (using a strong route)

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    const rootItem = new AnalysisJobItemResult({
+      name: "",
+      path: this.rootItemPath,
+      type: "directory"
+    });
```
atruskie: seems like a code smell, why is there an empty root item here?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    this.rows = [rootNode];
+
+    this.rows$ = new Observable(subscribers => {
+      subscribers.next(this.rows);
+    });
+  }
+
+  /**
+   * Fetches the `AnalysisJobItemResult` model from the baw-api and converts it to to be compliant with the `ResultNode` interface
```
atruskie: this comment is inaccurate, the return type is `Observable<AnalysisJobItemResult>`

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+    const analysisJobId = this.analysisJob;
+    return this.api.show(model.result, analysisJobId, this.audioRecording.id);
+  }
+
+  private closeRow(model: ResultNode): void {
+    model.children?.forEach(child => {
+      // remove the child element from the rows
+      this.rows.splice(this.rows.indexOf(child), 1);
+
+      // close all the child elements of the row
```
atruskie:
```suggestion
      // Recursive: close all the child elements of the row
```

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+      this.closeRow(child)
+    });
+
+    model.open = false;
+  }
+
+  private openRow(model: ResultNode): void {
```
atruskie: `model` isn't a great name; maybe `node`

atruskie: applies to all usages below

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+    return new Observable<ResultNode[]>(
+      subscriber => {
+        this.getItem(model)
+          // add the path & parent information to all child items
+          .pipe(map(returnedValue => this.childItemsWithParentInformation(returnedValue)))
+          .pipe(takeUntil(this.unsubscribe))
+          .subscribe({
+            next(modelChildren) {
+              subscriber.next(modelChildren);
+            }
```
atruskie: is this part really necessary?

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * Takes a view model and returns the child items, with the `path`, `analysisJobId`, and `audioRecordingId` attributes
+   *
+   * @param model A view model to fetch the children of
+   * @returns An array of `AnalysisJobItemResult` representing the child items inside the model
+   */
+  private childItemsWithParentInformation(model: AnalysisJobItemResult): ResultNode[] {
+    const evaluatedSubItems = Array<ResultNode>();
+
+    model.children.forEach(item =>
```
atruskie: could just use a map function

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+      evaluatedSubItems.push({
+        parentItem: model,
+        result: new AnalysisJobItemResult({
+          type: "file",
+          path: model.path + item.name,
+          ...item
+        }),
```
atruskie: why are you remaking the item?

Just add a getter to the AnalysisJobItemResult model to backfill the path

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    model.open = false;
+  }
+
+  private openRow(model: ResultNode): void {
+    this.getModelChildren(model)
+      // since we have evaluated the children of the model, append this information to the model
+      // so that we don't have to re-query this information in the future from the API
+      .pipe(map(children => model.children = children))
+      .pipe(map(returnedValues => this.rows.splice(this.rows.indexOf(model) + 1, 0, ...returnedValues)))
```
atruskie: I don't like all the splicing here. It's super messy. 

You've got the right structure but try immutable and flat approaches over mutative recursion.

`closeRow` should be something like `this.rows = this.rows.filter(row => row.isChildOf(node);`. No recursion, no splicing.

`openRow` should be something like: `this.rows = this.rows.append(newNodes).sort(compareByPath)`;

src/app/models/AudioRecording.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    expect(realizedUrl).toEqual(expectedUrl);
+  });
+
+  it("should construct correct analyses download URL", () => {
+    const expectedUrl = `/audio_recordings/${audioRecordingModel.id}/results`;
+    const realizedUrl = audioRecordingModel.getAnalysisResultsUrl();
+
+    expect(realizedUrl).toEqual(expectedUrl);
+  });
```
atruskie: It's mounted at 4 different routes; why does it only return the route for one of them?

atruskie: ditto the above and below unit tests

src/app/services/baw-api/analysis/analysis-job-item-result.service.ts
```diff
@@ -21,7 +21,8 @@ const analysisJobId: IdParamOptional<AnalysisJob> = id;
 const audioRecordingId: IdParamOptional<AudioRecording> = id;
 const analysisJobItemResultsPath = param;
 
-const endpoint = stringTemplate`/analysis_jobs/${analysisJobId}/results/${audioRecordingId}/${analysisJobItemResultsPath}${option}`;
+export const analysisJobItemResultsEndpoint =
```
atruskie: by convention we name this constant endpoint in every file... i can't find where you used it, but the fact that it is exported hints to me that something isn't right.

atruskie (changes requested):

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
hudson-newey: 100%, makes sense, I'll be changing this to a function on the component

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
hudson-newey: It seems to be a weird result of how the code is formatted in the editor. I was able to remove this by removing the new line characters before and after the item name mustache, I ended up reverting this "fix" in favor of code legibility.

I agree that this is not ideal, and I'll have another look at it to see if there is a more correct solution. Seems to be similar to [this](https://stackoverflow.com/questions/35937687/white-space-added-by-angularjs-after-variable-write-in-a-for-attribute) SO question.

Current code formatting:
image

Fix I found to remove space character before and after file name:
image

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
hudson-newey: Good catch! I'll be adjusting the code shortly

src/app/services/baw-api/analysis/analysis-job-item-result.service.ts
hudson-newey: This seems to be an artifact of some old testing I was doing, I have reverted the change

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
hudson-newey: In all honestly, I did this for code clarity. This can be easily changed to a regular function, and I won't use arrow functions in this way in the future.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
hudson-newey: 100%

I remember there being a test for this in the past, not sure where it went though (might have been removed in one of my merges). I'll be able to restore it easy enough.

**Edit**: I've completed the test bellow to fulfill these requirements
image

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: When calling the API for the first time, you are fetching the model for a common ancestor e.g.

image

Because I thought there was no way of knowing what the root folder would be, I created a mock root item client side, that all other items would be relative to.

**Edit**: Now that I'm looking back on this though, there is a way to know the path of the root item. I probably should have used route data to determine the root folder / item and treat it like a normal result node.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.html
hudson-newey: Surprisingly yes. I've tried it without, and it returns the following error

image

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
hudson-newey: This makes sense to me as the attributes relate to the model instead of the component

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
hudson-newey: This seems to have been related to the comment bellow
> perhaps innerText is better: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext

The issue has now been resolved by using `innerText` apposed to `textContent`

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
hudson-newey: Sounds good. I'll be moving the `cursor: pointer;` definition into the `analysis-download-row.component.scss` file

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
hudson-newey: From my observation, we don't currently have a helper function for this.

There are two solutions:
a. Use answers [this](https://stackoverflow.com/a/20732091) answer from StackOverflow thread (not ideal, and is most of the answers are messy code)
b. Write our own helper method to humanize values
c. Use the [pretty-bytes](https://www.npmjs.com/package/pretty-bytes) npm package

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
hudson-newey: This can easily be done.

I originally decided not to because the column widths would change when opening and closing nodes / items. This was because the columns were scaled to fit the file names, size, etc.. so if a file with a larger name / size came along, the entire table would re-scale to fit the content best.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
hudson-newey: I've fixed this in a upcoming commit I will push at the end of the day as a WIP

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: The AnalysisJobItemResult model doesn't have the information on the parent node / model, thus why I placed this on the `ResultNode` interface. The path / parent information also needs to be communicated to the `analysis-download-row.component.ts`, which is why I added it to the result model.

The path information can be derived using the parent node + item name, but this information will need to be communicated to the `analysis-download-row.component.ts`, which will require the `ResultNode` interface to be replicated on the `analysis-download-row.component.ts`.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: This will be resolved by the following comment
> this should be an input for the component; specifically, since this is a page, it should be resolved via route data.
>  
> This view can't be mounted for any other analysis job currently.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
hudson-newey: I was able to get this working without any ng properties by using css's nth-child(2n) psudo attribute.
image

This will also resolve the comments bellow relating to `even` attribute, and will greatly simplify the code

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.html
hudson-newey: The need to track if a row is even has been removed due to css having an inbuilt psudo class to handle situations like this (`nth-child`)

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
hudson-newey: I am going to be using the `--baw-secondary` variable for this colour

src/app/models/AudioRecording.spec.ts
hudson-newey: I've completed this by explicitly creating tests for each route.

As this is not ideal, I'll have a look into if I can use some sort of `for in` loop to prevent a lot of duplicate code.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.scss
atruskie: hmm you make a good point.

If you could max-width the name column and auto-min-width the two end columns (without reflow on folder open/close) then that would be good. Otherwise, yeah avoiding reflow take priority.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.spec.ts
atruskie: ok, if the spacing is an artefact of the layout you probably wanted to strip leading/trailing whitespace of the value you retrieved before testing it. Striping implies the whitespace is not important, whereas actually including the white space in the test implies it is important to the test.

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.ts
atruskie: if you answer what "I want to use it as a lambda for functional programming" then that would have been fine.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.html
atruskie: that's whack

hudson-newey: Good whack hopefully, I thought it was pretty neat when I found it

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
atruskie: so; 

i'm not sure what information is missing; show me in a call

but you didn't answer why you're recreating the model

another design: a function on the model `getFullPath(parent)`

src/app/components/audio-recordings/components/analyses-download/analysis-download-row.component.html
hudson-newey: I have got this working using the `fileSize` npm package that is already part of our project.

image

~However, I think this is a bit of a code smell and I should be using the `@bawByte` decorator instead. I've had a try and had some troubles, but I'll have a look at how to replicate this same functionality using `@bawByte` on the model~

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: No; good catch, I was able to remove this weird double observable

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: The need for the path information to be transferred between `analyses-results.component.ts` and `analysis-download-row.component.ts` has been removed since indentation can be included in the `ResultNode` interface. Removing the need poor code quality that lead to this situation.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: I have been able to get this method working with a nifty map function replacement. I might even be able to remove this method entirely, and add it to a pipe as it's only used in one other place in the program
image

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.scss
hudson-newey: I've successfully got this to work using the `::ng-deep` pseudo element

image

Code is subject to change in the future in a refactor of code quality / cleanup

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: The sort algorithm might be possible by doing something along these lines:
1. Convert the paths to two arrays of ASCII/UTF values
2. Compare the two arrays index by index
  a. If the value of `pathA[i] < pathB[i]` return -1, if the value of `pathA[i] > pathB[i]`, return 1, otherwise
  b. If the character value of the nth character is equal, move to the next character and repeat
  c. To prevent reading out of index, if the length of `length pathA < i` return -1, if the length of `pathB < i` return 1
3. Default to returning 0, this means that both paths are identical (which will never occur, but is good practice and good for redundancy so we fail gracefully)

We might even be able to do some weird bitwise trickery?? This is just a thought, but I don't know the specifics on how this would work

hudson-newey: Binning paths by path components and sorting by size was effective in creating a valid sort algorithm.

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: Since analysis job information is not currently in the route data, I am going to be structuring the new route as follows

`/audio_recordings/:audioRecordingId/analysis_jobs/:analysisJobId/results:path`

hudson-newey: By using a predefined (hard coded) system analysis job defined in the `analysis-jobs.service.ts`, I was able to get this working with both system jobs, and theoretically user defined jobs.

image

src/app/components/audio-recordings/pages/analysis-results/analyses-results.component.ts
hudson-newey: The need for a root item path has been removed since I am now deriving the root item from the route information.

**Edit**: I have since removed this `rootItemPath` variable as it's no longer needed

hudson-newey:
The code currently pushed is not ideal & is of bad quality, but has the requirements listed above implemented (sub 1).

I am going to be cleaning up the code quality before submitting it for review, but I wanted to push it to avoid losing my work due to a hard drive failure, etc... and create a point of reference for a very rough MVP.

**Personal Review TODO list**
- [x] Fix indentation method
- [x] Update `analyses-results.component.spec.ts` test
- [x] Update `analysis-download-row.component.spec.ts` tests
- [x] Update `AudioRecording.spec.ts` tests
- [x] Add unit tests for file `Array.prototype.sort()` method
- [x] Using filter & map methods means the `this.rows` array is a new instance every change. Therefore the observable isn't being updated. I need to find a fix for this
- [x] Remove root path from `analysis-results.component.ts`
- [x] Fix system routes in menu items
- [x] Change typescript access modifiers to better reflect usage
- [x] Add code comments
- [x] Add `@BawBytes` decorator to `AnalysisJobItemResult` model
- [x] Write JsDoc comments
- [x] Consolidate duplicate code in `analysis-download-row.component.ts`
- [x] Fix download button for folders of more than 2 levels of depth
- [x] Add error notification if there are no analysis job item results
- [x] Fix eslint errors
- [x] Move system analysis job to menu items instead of routes
- [x] Fix code comment in `analysis-download-row.component.spec.ts`
- [x] Assert that buttons next to files are not disabled
- [x] Use `analysis-download-row.component.ts` view model in tests
- [x] Remove API root references from `analysis-download-row.component.ts` (it shouldn't have any knowledge of routes)
- [x] Remove reminence of using `px` in css and replace with em / rem
- [x] Change even checkered background colour to use css variables defined in the global style
- [x] Change `analysis-results.component.spec.ts` default audio recording not to have a specified ID attribute
- [x] See if `setup()` in `analysis-results.component.spec.ts` needs `{ detectChanges: false }` (it does)
- [x] In `getDownloadButton` in `analysis-results.component.spec.ts` I should use `By.css()` instead of by index (`By.css` only works on jasmine fixtures, not HTML elements, this is NA)
- [x] Remove re-instantiating `AnalysisJobItemResult` model on `ResultNode` (not needed because the models are incomplete)
- [x] Download button tests sometimes fail randomly
- [x] Fix view on mobile

atruskie:
I'm not going to merge this until associated changes with the API are finalized. https://github.com/QutEcoacoustics/baw-server/tree/pbs-analysis

atruskie:
We've got a new API and the codebase has evolved significantly since this was first opened. Recommend scrapping this PR and taking parts for a new attempt.

PR #2017, opened 2022-12-29
Fix remove image button

### LINKED ISSUE #1994: Remove image button should not shown when there is no image to remove.
**Describe the bug**

Remove image button should not shown when there is no image to remove.

**To Reproduce**

1. Go to any project/site/region edit page
2. Remove and image if necessary and update the resource
3. Go back to the edit page and observe the remove button is still available

image

**Expected behavior**

The remove button should be disabled, or hidden if the remove image action is not relevant.

hudson-newey:
# Fix remove image button

The remove image button should only show when a project has an image

## Changes

- Displays the currently set image file name in the image input box
- Adds a `default` flag to baw images (outlined in `AttributeDecorators.ts`)
- Adds a _TODO_ to delete the code block [here](https://github.com/QutEcoacoustics/workbench-client/pull/2017/files#diff-5cf4eef2ac11e42d9b8698e07af25283f1d339b38cd87d8188bc87ad5984ae3bR135) once the baw-api issue baw-api#640 is resolved
- Fixes bug where remove button caused the file input containers height to expand
- Changes image input component to use classes from the baw stylesheet instead of hard coded inline css values

## Issues
Fixes: #1994 

## Visual Changes
From:
image

To:
image

**Visual Changes List:**
- Image input component now uses boostrap classes for border colours (it had previously hard coded these values in a `style` attribute)
- The remove image button now fits inside it's parent item
- If the model does not have a default image, it now shows the file name in the input box

_Notes: First image was taken on Firefox, where the "Browse..." text is used, the second was taken from Edge where the "Choose File" text is used, hence the change in text_

## Final Checklist

### DIFF
--- a/src/app/components/shared/formly/image-input.component.ts
+++ b/src/app/components/shared/formly/image-input.component.ts
@@ -1,5 +1,6 @@
-import { Component, ViewChild, ElementRef } from "@angular/core";
+import { Component, ViewChild, ElementRef, AfterViewInit } from "@angular/core";
 import { FieldType } from "@ngx-formly/core";
+import { ImageUrl } from "@interfaces/apiInterfaces";
 import { asFormControl } from "./helper";
 
 /**
@@ -14,10 +15,7 @@ import { asFormControl } from "./helper";
         {{ props.label + (props.required ? " *" : "") }}
       </label>
 
-      <div
-        class="input-group"
-        style="border: 1px solid #ced4da; border-radius: 0.25rem;"
-      >
+      <div class="form-control input-group p-0">
         <!-- Ensure only one file can be selected in input -->
         <input
           #imageInput
@@ -30,18 +28,63 @@ import { asFormControl } from "./helper";
         />
 
         <button
+          *ngIf="!usesDefaultImage"
           type="button"
           (click)="removeImage()"
-          class="btn btn-outline-danger"
-        >Remove</button>
+          class="btn btn-outline-danger pb-1"
+        >
+          Remove
+        </button>
       </div>
     </div>
   `,
 })
-export class ImageInputComponent extends FieldType {
+export class ImageInputComponent extends FieldType implements AfterViewInit {
   @ViewChild("imageInput")
   public imageInput: ElementRef;
   public asFormControl = asFormControl;
+
+  public ngAfterViewInit() {
+    if (!this.usesDefaultImage) {
+      const imageUrls = this.model.imageUrls as ImageUrl[];
+      const currentImage = imageUrls && imageUrls[0];
+      const imageFileName = this.fileName(currentImage?.url);
+
+      // if the current model has a different image to the default image, display the name of the file in the image input
+      // for security reasons, modern web browsers don't allow changing a file inputs value directly
+      // because of this we need to use a data transfer which is performed in "protected mode"
+      // https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files
+      const dataTransfer = new DataTransfer();
+      dataTransfer.items.add(new File([""], imageFileName));
+      this.imageInput.nativeElement.files = dataTransfer.files;
+    }
+  }
+
+  /**
+   * A predicate that returns if the models image is a default image
+   */
+  public get usesDefaultImage(): boolean {
+    const imageUrls = this.model?.imageUrls as ImageUrl[];
+    const isUsingServerDefaultImage = imageUrls?.every((image: ImageUrl): boolean => image.default);
+
+    // returns true if the current image used is a default image and if the user hasn't redefined the image
+    // or returns true if the user explicitly sets the image as default, by setting the models image attribute to null
+    return (
+      (isUsingServerDefaultImage && this.model.image === undefined) ||
+      this.model.image === null
+    );
+  }
+
+  private fileName(filePath: string): string {
+    return (
+      filePath?.split("/")
+        .pop()
+        // remove URL parameters from the file name
+        .split("?")
+        .shift()
+    );
+  }
+
   public readFile(): void {
     // File input returns a list of files, grab the first file and set it as
     // the value of this field
--- a/src/app/models/AttributeDecorators.ts
+++ b/src/app/models/AttributeDecorators.ts
@@ -83,7 +83,11 @@ export function bawImage<Model>(
   opts?: BawDecoratorOptions<Model>
 ) {
   // Retrieve default image and prepend site url if required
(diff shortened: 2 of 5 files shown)


hudson-newey:
Minor improvement to code quality: The [code](https://github.com/QutEcoacoustics/workbench-client/pull/2017/commits/7d17743e7d18efb196b123dc9132eb5846844649#diff-70c288a840943b6e9bb1df6332c0663fc0aa4cb739622f7b66ab8796b95fbf9bR109) for how images are removed was changed to be consistent with the [code](https://github.com/QutEcoacoustics/workbench-client/pull/2017/commits/7d17743e7d18efb196b123dc9132eb5846844649#diff-70c288a840943b6e9bb1df6332c0663fc0aa4cb739622f7b66ab8796b95fbf9bR105) to upload images

src/app/components/shared/formly/image-input.component.ts
```diff
@@ -30,18 +37,55 @@ import { asFormControl } from "./helper";
...
   @ViewChild("imageInput")
   public imageInput: ElementRef;
   public asFormControl = asFormControl;
+
+  public ngOnInit() {
+    // since the fallback image will always have the default attribute
+    // we need to check if there is more than one image with the default attribute set
+    // if the the image is the default image, update the model to use null as the image.
+    // we use an additional view model attribute to prevent sending duplicate image information in the PATCH request
+    this.model.hasDefaultImage = this.model.imageUrls.filter((image) => image.default === true).length > 1;
```
atruskie: 🤔 this tricky condition makes me think we haven't set our flags right...

atruskie: also why is a component updating a model about with knowledge the model has about itself?

atruskie: okay the condition here should be: if the number of images == the number of default images, then there are only default images.

atruskie: and make this check a predicate that accepts the model to check. Don't mutate the model object

src/app/models/AttributeDecorators.ts
```diff
@@ -128,6 +128,12 @@ export function bawImage<Model>(
           defaultImage,
         ];
       } else if (imageUrls instanceof Array && imageUrls.length > 0) {
+
+        imageUrls = imageUrls.map((image) => ({
+            ...image,
+            ...defaultImage.url.includes(image.url) && {default: true},
+          }));
```
atruskie: I don't understand why this part is needed.

The default image, wherever it is used, has `default: true`, why then do you need to check the responses from the server. If the server sends any non-null image paths then they are by definition 'set' and not default

atruskie (changes requested):

src/app/models/AttributeDecorators.ts
hudson-newey: Sorry! I should have added a comment saying that this PR is a WIP.

This code above is a very hacky work around that I created to test if other components would react to the default attribute correctly. I'm working towards the proper approach described above, but I'm still having troubles locating why I'm receiving extra default images (located in the `image_urls` property of the model).

After sending my patch setting `project: { image: null }`, the image_urls attribute on the model is still populated with `/images/project/project_span4.png`, `/images/project/project_span3.png`, etc.... I've been able to replicate this in the client & postman so I'm actively trying to narrow it down further.
e.g. explicitly setting `image_urls: null` or `image_urls: []`, etc... (as this bug may be the result of the remove image portion of code not working correctly)

Delete image PATCH request in (Postman)
image

response from remove image PATCH request (Postman)
image

GET request response for the project model after removing the image
image

atruskie: Ok, I found the root cause: https://github.com/QutEcoacoustics/baw-server/blob/b5a9f807d67a414610c665a499f39d50aa37d423/app/models/project.rb#L68

This will happen for [sites](https://github.com/QutEcoacoustics/baw-server/blob/b5a9f807d67a414610c665a499f39d50aa37d423/app/models/site.rb#L62), [projects](https://github.com/QutEcoacoustics/baw-server/blob/b5a9f807d67a414610c665a499f39d50aa37d423/app/models/project.rb#L68), and [users](https://github.com/QutEcoacoustics/baw-server/blob/b5a9f807d67a414610c665a499f39d50aa37d423/app/models/user.rb#L90), and is considered deprecated behavior. The hack you have here we might need to keep for the time being, but I'd probably suggest:

- matching urls with a regex: `\/images\/(.*)\/(\1)_span.*.png`
- documenting this fully
- filing an issue with the server to remove the default_url attribute

hudson-newey: Sounds like a good plan.

I agree using regex is 100% a more robust approach to this hack.

I'll be updating the PR with the requested changes (including what I can complete from the comments above e.g. if # images == # default images), documenting, and filing an issue shortly.

atruskie: remember to document your hack including a link to the API issue you create

hudson-newey: I'll be accomplishing this through JsDoc comments, inline code comments, and `TODO: xyz` comments (along with linked API issues) as I don't think this warrants it's own wiki article.

src/app/components/shared/formly/image-input.component.ts
```diff
@@ -30,18 +27,65 @@ import { asFormControl } from "./helper";
...
 })
-export class ImageInputComponent extends FieldType {
+export class ImageInputComponent extends FieldType implements AfterViewInit {
   @ViewChild("imageInput")
   public imageInput: ElementRef;
   public asFormControl = asFormControl;
+
+  public ngAfterViewInit() {
+    if (!this.usesDefaultImage) {
+      const currentImage = this.model.imageUrls?.shift();
```
peichins: Just use the array index to get the first element without removing it from the array

src/app/components/shared/formly/image-input.component.ts
```diff
@@ -30,18 +27,65 @@ import { asFormControl } from "./helper";
...
+      this.imageInput.nativeElement.files = dataTransfer.files;
+    }
+  }
+
+  /**
+   * A predicate that returns if the models image is a default image
+   */
+  public get usesDefaultImage(): boolean {
+    const imageUrls = this.model?.imageUrls;
+    const defaultImageCounts = imageUrls?.filter(
```
peichins: maybe use array.every() to test for the presence of a default image in the array, for succinctness and efficiency

peichins (changes requested):

atruskie (approved):
Looks good. Nice work.

PR #2020, opened 2023-01-03
Do not show bulk audio download command if the user is not logged in

### LINKED ISSUE #1930: Bulk audio download: don't show command when a user is not logged in
From user testing: when users are not logged in they tend to just copy and paste the download command without realising that the `"INSERT_AUTH_TOKEN"` here token needs  to be replaced.

Instead we should hide the command and only show it when someone is logged in. If they're not logged in we should have a warning panel in place of the command with a message like _log in to see command_..

hudson-newey:
# Do not show bulk audio download command if the user is not logged in

It has been reported that users are regularly pasting the bulk audio download command directly into their shell without replacing the `"INSERT_AUTH_TOKEN_HERE"` placeholder. To fix this, we should require users to log in before this command is shown.

## Changes

- Requires the user to log in to see the bulk download audio command
- Removes redundant information above command stating that they need to "Login to see the command filled with your authentication token." This is now replaced by the command not showing, but alerting the user to log in through an embedded anchor that redirects back to the view once the user is authenticated.

## Problems

None

## Issues

Fixes: #1930 

## Visual Changes

image

## Final Checklist

### DIFF
--- a/src/app/components/audio-recordings/pages/download/download.component.html
+++ b/src/app/components/audio-recordings/pages/download/download.component.html
@@ -329,14 +329,15 @@
 
   <p id="run-script-description">
     Copy and paste the following command into your PowerShell session.
-    <span *ngIf="!session.authToken">
-      Login to see the command filled with your authentication token.
-    </span>
   </p>
 
-  <code id="guest-run-script" *ngIf="!session.authToken">{{
-    runScriptCommand
-  }}</code>
+  <div
+    id="guest-run-script"
+    class="alert alert-warning"
+    *ngIf="!session.authToken"
+  >
+    <i><a [strongRoute]="loginStrongRoute">Log in</a> to see this command.</i>
+  </div>
 
   <baw-hidden-copy
     *ngIf="session.authToken"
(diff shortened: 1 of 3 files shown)


src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -328,9 +328,9 @@ <h3>5. Run the script</h3>
     </span>
   </p>
 
-  <code id="guest-run-script" *ngIf="!session.authToken">{{
-    runScriptCommand
-  }}</code>
+  <div id="guest-run-script" class="alert alert-warning" *ngIf="!session.authToken">
+    <i>You will have to log in to see this command.</i>
```
atruskie:
```suggestion
    <i>Log in to see this command.</i>
```

src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -328,9 +328,9 @@ <h3>5. Run the script</h3>
...
+  </div>
```
atruskie: add an anchor to the sign in view, and make sure it redirects back to this view after a user clicks on it

atruskie (changes requested):

atruskie (approved):

PR #2030, opened 2023-01-17
Add recording filtering component and adjust existing filter form validations

### LINKED ISSUE #1724: Adding filters to the recording's list page
There are just too many recordings and it would be useful to have a filter for when you're looking for a specific date, for example.

A filter feature from the recording's list page.

False Colour Visualisation can make it faster to find specific date/time, but if it's not working or you don't want to use it, then filtering would be helpful.

**Additional context**
I'm referring to add a filter in this page:

image

### LINKED ISSUE #1927: Allow selecting one day of recordings for batch download
User tried to select one day of recordings to download.

With one day range: 0 results
With two days range:  two-days worth.

This is likely a bug our end exclusion constraint that required some trickery because arrays of conditions were not supported for filter queries (https://github.com/QutEcoacoustics/baw-server/issues/567)

### LINKED ISSUE #2018: Batch download filter payload is submitting a short-year date
**Describe the bug**

```
 {"filter"=>{"sites.id"=>{"eq"=>307}, "recorded_date"=>{"greater_than_or_equal"=>"2020-12-29"}, "recorded_end_date"=>{"less_than_or_equal"=>"20-12-31"}}, "paging"=>{"page"=>1}, "format"=>"json", "controller"=>"audio_recordings", "action"=>"filter", "audio_recording"=>{}}
```

This payload was submitted by the batch download page. It looks like somehow the user managed to submit a short form date string for the less_than_or_equal component.

**To Reproduce**

Not sure, try and type the short date in the form inputs. I'm pretty sure we validated for this.

**Expected behavior**

The server expects only long form years for dates.

---

Related to: https://github.com/QutEcoacoustics/baw-server/issues/639

hudson-newey:
# Add recording filtering component and adjust existing filter form validations

An audio recording filter component already exists on the batch download page, but it is not generalised for use on the audio recordings list page and in other places in the client.

In the old batch download filter component, we were relying on the Angular-Boostrap validations for date inputs. However, dates with less than 4 characters are not valid and cause an internal server error when the api is called.

Therefore, we want to pad dates with less than 4 characters using the ISO 8601 format.

## Changes

**Pages & Components**
- Creates an `audio-recording-filter` component that emits a filter update event
- Adds an `audio-recording-filter` component to the audio recordings list page

- Adds a midnight time filter, allowing users to filter times past day ranges
- Adds filter helper method to filter by model id's
- Adds audio recording filter helper methods
  - Adds filter helper method to filter audio recordings by a time range/interval
  - Adds filter helper method to filter audio recordings by date range/interval
- With the addition of the `and[]` filter array condition to the api 🎉, there is now an `addFilter` method which adds an inner filter condition to an additional filter using the `and` filter operator.
- Recording filter component now uses Angular form validation 🎉

---

**Time Component**

- Now emits a Luxon `Duration` type

---

**Case Converter**
- modifies `caseConverter` to simplify object values through their `toJSON` implementations
- Adds a `SerializableObject` interface type that can be used with inner filters

---

**Tests & Unit Tests**

- Creates tests for the batch download component to validate that the client displays the correct errors when an invalid date is added
- Creates tests for the new `audio-recording-filter.component.ts` component to ensure filter updates are only emitted when needed
- Creates tests for the new `audio-recording-filter.component.ts` component to ensure the correct errors are displayed and no filter update is emitted when the user inputs a "bad" date/time
  - Start date comes after the end date
  - The time is not in the correct format
  - The date is not in the correct format

- Creates tests to ensure the `addFilter` condition uses the `and` filter operation correctly
- Creates tests for the audio recording filter helper functions

---

**Minor Changes**

- Changes wording used for invalid dates to use the word "date" instead of "time" in errors

## Problems

None

## Issues

Fixes: #2018 
Fixes: #1724 
Fixes: #1927 

## Visual Changes

**Invalid dates:**
image

**Error if the start date > end date:**
image
  
---
  
**Filter component on audio recordings list page (collapsed):**
image

**Filter component on audio recordings list page (expanded):**
image

## Final Checklist

### DIFF
--- a/src/app/components/audio-recordings/pages/list/list.component.ts
+++ b/src/app/components/audio-recordings/pages/list/list.component.ts
@@ -1,7 +1,7 @@
 import { AfterViewInit, Component, Inject, OnInit } from "@angular/core";
 import { ActivatedRoute } from "@angular/router";
 import { AudioRecordingsService } from "@baw-api/audio-recording/audio-recordings.service";
-import { Filters, InnerFilter } from "@baw-api/baw-api.service";
+import { Filters } from "@baw-api/baw-api.service";
 import { projectResolvers } from "@baw-api/project/projects.service";
 import { regionResolvers } from "@baw-api/region/regions.service";
 import { siteResolvers } from "@baw-api/site/sites.service";
@@ -10,9 +10,10 @@ import {
   audioRecordingsCategory,
 } from "@components/audio-recordings/audio-recording.menus";
 import { visualizeMenuItem } from "@components/visualize/visualize.menus";
+import { filterModel } from "@helpers/filters/filters";
 import { IPageInfo } from "@helpers/page/pageInfo";
 import { PagedTableTemplate } from "@helpers/tableTemplate/pagedTableTemplate";
-import { Id, Ids, toRelative } from "@interfaces/apiInterfaces";
+import { toRelative } from "@interfaces/apiInterfaces";
 import { AudioRecording } from "@models/AudioRecording";
 import { Project } from "@models/Project";
 import { Region } from "@models/Region";
@@ -20,7 +21,7 @@ import { Site } from "@models/Site";
 import { ConfigService } from "@services/config/config.service";
 import { API_ROOT } from "@services/config/config.tokens";
 import { List } from "immutable";
-import { Observable } from "rxjs";
+import { BehaviorSubject, Observable, takeUntil } from "rxjs";
 
 const projectKey = "project";
 const regionKey = "region";
@@ -51,6 +52,8 @@ class AudioRecordingsListComponent
     duration: "durationSeconds",
     site: "siteId",
   };
+  public filters$: BehaviorSubject<Filters<AudioRecording>> =
+    new BehaviorSubject({});
   protected api: AudioRecordingsService;
 
   public constructor(
@@ -92,6 +95,9 @@ class AudioRecordingsListComponent
 
     // Update table to show default sort
     this.table.sorts = [{ prop: "recorded", dir: "asc" }];
+    this.filters$.pipe(takeUntil(this.unsubscribe)).subscribe((filters) => {
+      this.updateFilters(filters);
+    });
   }
 
   public get projectId(): number | undefined {
@@ -128,28 +134,26 @@ class AudioRecordingsListComponent
     return recording.recordedDate.toFormat("yyyy-LL-dd HH:mm");
   }
 
+  public updateFilters(incomingFilters: Filters<AudioRecording>): void {
+    // since sorting is handled by the pagination table, when the filter is updated, it would remove the sorting order
+    // to ensure that sorting is retained throughout filtering, retain the previous sorting information
+    this.filters = {
+      sorting: this.filters?.sorting,
+      ...incomingFilters,
+    };
+
+    this.getPageData();
+  }
+
   protected apiAction(
     filters: Filters<AudioRecording>
   ): Observable<AudioRecording[]> {
-    function updateFilterWithSite(sites: Ids | Id) {
-      const siteFilters: Filters<AudioRecording> = (filters ??= {});
-      const siteInnerFilter: InnerFilter<AudioRecording> =
-        (siteFilters.filter ??= {});
-      const siteFilter = (siteInnerFilter.siteId ??= {});
-      if (sites instanceof Set) {
-        siteFilter.in = Array.from(sites);
-      } else {
-        siteFilter.eq = sites;
-      }
-    }
-
-    // Order matters
     if (this.site) {
-      updateFilterWithSite(this.site.id);
+      filters.filter = filterModel<Site, AudioRecording>("sites", this.site, filters.filter);
     } else if (this.region) {
-      updateFilterWithSite(this.region.siteIds);
+      filters.filter = filterModel<Region, AudioRecording>("regions", this.region, filters.filter);
     } else if (this.project) {
-      updateFilterWithSite(this.project.siteIds);
+      filters.filter = filterModel<Project, AudioRecording>("projects", this.project, filters.filter);
     }
 
     return this.api.filter(filters);
--- a/src/app/components/shared/date-time-filter/audio-recordings-filter.component.spec.ts
+++ b/src/app/components/shared/date-time-filter/audio-recordings-filter.component.spec.ts
@@ -1,0 +1,791 @@
+import {
+  Spectator,
+  createRoutingFactory,
+  mockProvider,
+} from "@ngneat/spectator";
+import { NgbCollapse, NgbCollapseModule } from "@ng-bootstrap/ng-bootstrap";
+import { CacheModule } from "@services/cache/cache.module";
+import { MockConfigModule } from "@services/config/configMock.module";
+import { SharedModule } from "@shared/shared.module";
+import { ToastrService } from "ngx-toastr";
+import { fakeAsync } from "@angular/core/testing";
+import { Project } from "@models/Project";
+import { generateProject } from "@test/fakes/Project";
+import { TimeComponent } from "@shared/input/time/time.component";
+import { BehaviorSubject } from "rxjs";
+import { Filters } from "@baw-api/baw-api.service";
+import { AudioRecording } from "@models/AudioRecording";
+import { AudioRecordingsFilterComponent } from "./audio-recordings-filter.component";
+
+describe("AudioRecordingsFilter", () => {
+  let spectator: Spectator<AudioRecordingsFilterComponent>;
+  let defaultProject: Project;
+  let filterChangeSpy: jasmine.Spy;
+
+  const createComponent = createRoutingFactory({
+    component: AudioRecordingsFilterComponent,
+    imports: [SharedModule, NgbCollapseModule, MockConfigModule, CacheModule],
+    declarations: [TimeComponent],
+    providers: [mockProvider(ToastrService)],
+  });
+
+  function setup(project: Project): void {
+    const resolvers = {};
+    const models = {};
+
+    resolvers["project"] = "resolver";
+    models["project"] = { model: project };
+
+    spectator = createComponent({
+      data: { resolvers, ...models },
+      detectChanges: false,
+    });
+
+    spectator.component.constructedFilters = new BehaviorSubject<
+      Filters<AudioRecording>
+    >({});
+    filterChangeSpy = spyOn(spectator.component.constructedFilters, "next");
+    updateForm();
+  }
+
+  beforeEach(fakeAsync(() => {
+    defaultProject = new Project(generateProject());
+
+    setup(defaultProject);
+    updateForm();
+  }));
+
+  // helper methods
+  function updateForm() {
+    spectator.detectChanges();
+    // Have to do this tick because of
+    // https://github.com/angular/angular/issues/22606#issuecomment-377390233
+    // Also because the observable which updates the filter has a debounce timer
+    spectator.tick(1000);
+    spectator.detectChanges();
+  }
+
+  function getElementByInnerText(innerText: string): HTMLElement {
+    return spectator.debugElement.query(
+      (el) => el.nativeElement.innerText === innerText
+    )?.nativeElement as HTMLButtonElement;
+  }
+
+  const getDateToggleInput = () =>
+    spectator.query<HTMLInputElement>("#date-filtering");
+  const getDateInputWrapper = () =>
+    spectator.query("#date-filters-wrapper", { read: NgbCollapse });
+  const getDateStartedAfterInput = () =>
+    spectator.query<HTMLInputElement>("#date-started-after");
+  const getDateFinishedBeforeInput = () =>
+    spectator.query<HTMLInputElement>("#date-finished-before");
+  const getTimeOfDayInputWrapper = () =>
+    spectator.query("#time-filters-wrapper", { read: NgbCollapse });
+  const getTimeOfDayToggleInput = () =>
+    spectator.query<HTMLInputElement>("#time-filtering");
+  const getIgnoreDaylightSavingsInput = () =>
+    spectator.query<HTMLInputElement>("#ignore-daylight-savings");
+  const getTimeOfDayStartedAfterInput = () =>
+    spectator.query<HTMLInputElement>("#time-started-after input");
+  const getTimeOfDayFinishedBeforeInput = () =>
+    spectator.query<HTMLInputElement>("#time-finished-before input");
+
+  function typeInElement(element: HTMLInputElement, value: string): void {
+    spectator.typeInElement(value, element);
+    spectator.dispatchTouchEvent(element, "blur");
+    updateForm();
+  }
+
+  function toggleTimeOfDayFilters() {
+    const input = getTimeOfDayToggleInput();
+    input.click();
+    input.dispatchEvent(new Event("input"));
+
+    updateForm();
+  }
+
+  function toggleDateFilters() {
+    const input = getDateToggleInput();
+    input.click();
+    input.dispatchEvent(new Event("input"));
+
+    updateForm();
+  }
+
+  function toggleIgnoreDaylightSavings() {
+    const input = getIgnoreDaylightSavingsInput();
+    input.click();
+    input.dispatchEvent(new Event("input"));
+
+    updateForm();
+  }
+
+  // start of assertion
+  it("should create", () => {
+    expect(spectator.component).toBeInstanceOf(AudioRecordingsFilterComponent);
+  });
+
+  describe("date filter input", () => {
+    it("should initially hide date filter input", fakeAsync(() => {
+      expect(getDateToggleInput()).toExist();
+      expect(getDateInputWrapper().collapsed).toBeTrue();
+    }));
+
+    it("should show date filter input when the date filter checkbox is set", fakeAsync(() => {
+      toggleDateFilters();
+      expect(getDateToggleInput()).toBeTruthy();
+      expect(getDateInputWrapper().collapsed).toBeFalse();
+    }));
+
+    it("should hide date filter input when the date filter checkbox is clicked twice", fakeAsync(() => {
+      toggleDateFilters();
+      toggleDateFilters();
+      expect(getDateToggleInput()).toExist();
+      expect(getDateInputWrapper().collapsed).toBeTrue();
+    }));
+  });
+
+  describe("time filter input", () => {
+    it("should initially hide time of day filters", fakeAsync(() => {
+      expect(getTimeOfDayToggleInput()).toExist();
+      expect(getTimeOfDayInputWrapper().collapsed).toBeTrue();
+    }));
+
+    it("should show time of day filter input when the time of day filter checkbox is set", fakeAsync(() => {
+      toggleTimeOfDayFilters();
+      expect(getTimeOfDayToggleInput()).toExist();
+      expect(getTimeOfDayInputWrapper().collapsed).toBeFalse();
+    }));
+
+    it("should hide timeOfDay filter input when the time of day filter checkbox is clicked twice", fakeAsync(() => {
+      toggleTimeOfDayFilters();
+      toggleTimeOfDayFilters();
+      expect(getTimeOfDayToggleInput()).toExist();
+      expect(getTimeOfDayInputWrapper().collapsed).toBeTrue();
+    }));
+
+    it("should initially have day light savings time enabled", fakeAsync(() => {
+      toggleTimeOfDayFilters();
+      expect(getIgnoreDaylightSavingsInput().checked).toBeTrue();
+    }));
+  });
+
+  describe("date format validation", () => {
+    const invalidDateErrorMessage =
+      "The date should follow the format yyyy-mm-dd, e.g. 2022-12-01";
+
+    beforeEach(fakeAsync(() => toggleDateFilters()));
+
+    it("should not show an invalid date error if the user has not input a date", fakeAsync(() => {
+      updateForm();
+      expect(getElementByInnerText(invalidDateErrorMessage)).not.toExist();
+    }));
+
+    it("should display an error if the user inputs an invalid date into the start date input", fakeAsync(() => {
+      const invalidDate = "-1999-13-02";
+      typeInElement(getDateStartedAfterInput(), invalidDate);
+      expect(getElementByInnerText(invalidDateErrorMessage)).toExist();
+    }));
+
+    it("should not display an error if the user inputs a valid date into the start date input", fakeAsync(() => {
+      const validDate = "2020-12-31";
+      typeInElement(getDateStartedAfterInput(), validDate);
+      expect(getElementByInnerText(invalidDateErrorMessage)).not.toExist();
+    }));
+
+    it("should display an error if the user types in an invalid date into the end date input", fakeAsync(() => {
+      // in this test, the user mixed their month and day, where the month (21) is greater than 12
+      const invalidDate = "2020-21-12";
+      typeInElement(getDateFinishedBeforeInput(), invalidDate);
+      expect(getElementByInnerText(invalidDateErrorMessage)).toExist();
+    }));
+
+    it("should not display an error if the user inputs a valid date into the end date input", fakeAsync(() => {
+      const validDate = "2020-12-31";
+      typeInElement(getDateFinishedBeforeInput(), validDate);
+      expect(getElementByInnerText(invalidDateErrorMessage)).not.toExist();
+    }));
+  });
+
+  describe("date bounds validation", () => {
+    const outOfBoundsDateErrorMessage =
+      "The filter dates are outside the date boundary. Ensure that the start date occurs before the end date.";
+
+    beforeEach(fakeAsync(() => toggleDateFilters()));
+
+    it("should show an error if the user inputs an out of bounds date filter", fakeAsync(() => {
+      const startDate = "2020-10-10";
+      const endDate = "2019-12-10";
+
+      typeInElement(getDateStartedAfterInput(), startDate);
+      typeInElement(getDateFinishedBeforeInput(), endDate);
+
+      expect(getElementByInnerText(outOfBoundsDateErrorMessage)).toExist();
+    }));
+
+    it("should not show an error if the user inputs a valid date range", fakeAsync(() => {
+      const startDate = "2019-12-10";
+      const endDate = "2020-10-10";
+
+      typeInElement(getDateStartedAfterInput(), startDate);
+      typeInElement(getDateFinishedBeforeInput(), endDate);
+
+      expect(getElementByInnerText(outOfBoundsDateErrorMessage)).not.toExist();
+    }));
+
+    it("should not show an out of bounds date error if the user has only input a start date", fakeAsync(() => {
+      const startDate = "2021-10-12";
+      typeInElement(getDateStartedAfterInput(), startDate);
+      expect(getElementByInnerText(outOfBoundsDateErrorMessage)).not.toExist();
+    }));
+
+    it("should not show an out of bounds date error if the user has only input an end date", fakeAsync(() => {
+      const endDate = "2021-09-01";
+      typeInElement(getDateFinishedBeforeInput(), endDate);
+      expect(getElementByInnerText(outOfBoundsDateErrorMessage)).not.toExist();
+    }));
+
+    it("should not show an out of bounds date error if the user has not input any dates", fakeAsync(() => {
+      updateForm();
+      expect(getElementByInnerText(outOfBoundsDateErrorMessage)).not.toExist();
+    }));
+
+    it("should not show an error if both the start and end dates are the same", fakeAsync(() => {
+      const startDate = "2020-10-10";
+      const endDate = "2020-10-10";
+
+      typeInElement(getDateStartedAfterInput(), startDate);
+      typeInElement(getDateFinishedBeforeInput(), endDate);
+
+      expect(getElementByInnerText(outOfBoundsDateErrorMessage)).not.toExist();
+    }));
+  });
+
+  describe("time bound validations", () => {
+    it("should allow a user to input a start time less than the end time without displaying an error", fakeAsync(() => {
+      const startTime = "01:00";
+      const endTime = "11:34";
+      const invalidInputClass = "is-invalid";
+
+      typeInElement(getTimeOfDayStartedAfterInput(), startTime);
+      typeInElement(getTimeOfDayFinishedBeforeInput(), endTime);
+
+      // is-invalid is a bootstrap validation class that will be added to an invalid time input (added by time.component.ts)
+      expect(getTimeOfDayStartedAfterInput()).not.toHaveClass(
+        invalidInputClass
+      );
+      expect(getTimeOfDayFinishedBeforeInput()).not.toHaveClass(
+        invalidInputClass
+      );
+    }));
+
+    it("should allow a user to input an end time less than the start time without displaying an error", fakeAsync(() => {
+      const startTime = "11:34";
+      const endTime = "01:56";
+      const invalidInputClass = "is-invalid";
+
+      typeInElement(getTimeOfDayStartedAfterInput(), startTime);
+      typeInElement(getTimeOfDayFinishedBeforeInput(), endTime);
+
+      expect(getTimeOfDayStartedAfterInput()).not.toHaveClass(
+        invalidInputClass
+      );
+      expect(getTimeOfDayFinishedBeforeInput()).not.toHaveClass(
+        invalidInputClass
+      );
+    }));
+  });
+
+  describe("filters events", () => {
+    function assertLastFilterUpdate(expectedFilter: Filters<AudioRecording>): void {
+      const mostRecentFilterUpdate = filterChangeSpy.calls.mostRecent().args[0];
+
+      const expectedFilterSerialized = JSON.stringify(expectedFilter);
+      const mostRecentFilterUpdateSerialized = JSON.stringify(mostRecentFilterUpdate);
+
+      expect(mostRecentFilterUpdateSerialized).toEqual(expectedFilterSerialized);
+    }
+
+    // date events where a filter update should not be emitted
+    it("should only emit one filter update event if the user hasn't input anything into the filter condition fields", fakeAsync(() => {
+      toggleDateFilters();
+      toggleTimeOfDayFilters();
+      // there is a filter update event emitted when the component is initialized
+      // this is done for the batch downloading component which needs the initial filter state for its batch downloading script
+      // therefore, if the user has not input any filters, there should only be the initial filter update
+      expect(filterChangeSpy).toHaveBeenCalledTimes(1);
+    }));
+
+    it("should not emit a filter update if the start date is a bad format", fakeAsync(() => {
+      // in this assertion, the user has typed a string into the date input. This will pass *some* assertions
+      // if this test is failing, the raw input from the input field is being passed to the date filter without its type being validated
+      const malformedStartDate = "testing";
+
+      expect(filterChangeSpy).toHaveBeenCalledTimes(1);
+
+      toggleDateFilters();
+      typeInElement(getDateStartedAfterInput(), malformedStartDate);
+      expect(filterChangeSpy).toHaveBeenCalledTimes(1);
+    }));
+
+    it("should not emit a filter update if the end date is a bad format", fakeAsync(() => {
+      // in this test, the day (12) and month (20) are swapped. It should not reformat the date or emit a filter update event in this case
+      const malformedEndDate = "2021-20-12";
+
+      expect(filterChangeSpy).toHaveBeenCalledTimes(1);
+      toggleDateFilters();
+      typeInElement(getDateStartedAfterInput(), malformedEndDate);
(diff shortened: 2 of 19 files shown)


src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+  todStartedAfter?: string;
+  todFinishedBefore?: string;
+}
+
+@Component({
+  selector: "baw-date-time-filter",
+  templateUrl: "filter.component.html",
+  styleUrls: ["filter.component.scss"]
+})
+export class DateTimeFilterComponent extends withUnsubscribe() implements AfterViewInit {
```
atruskie: file name does not reflect class name.

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+  dateStartedAfter?: NgbDate;
+  dateFinishedBefore?: NgbDate;
+  todFiltering?: boolean;
+  todIgnoreDst?: boolean;
+  todStartedAfter?: string;
+  todFinishedBefore?: string;
+}
+
+@Component({
+  selector: "baw-date-time-filter",
```
atruskie: Let's not arbitrarily limit this to dates and times; this a filter for recordings. Suggested: `baw-audio-recordings-filter`

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+    private formatter: NgbDateParserFormatter,
+    private calendar: NgbCalendar
+  ) { super(); }
+
+  @ViewChild(NgForm) public form: NgForm;
+
+  @Input() public project: Project;
+  @Input() public region: Region;
+  @Input() public site: Site;
+  @Output() public constructedFilters = new EventEmitter<Filters>();
```
atruskie: I wonder if we couldn't just accept the observable filter as an input; the component then call `.next` with the updated filter itself?

Much like we do for the pagination directive: https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/directives/datatable/pagination.directive.ts

Also I note that the pagination directive maintain sort order when the filter is updated... I'm assuming we're using this with the pagination directive so why the need to repeat sort order preservation as in https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-dc2d8f986bfbc9dcb229f29bbfff0e870eb6bf985baf35e4b2c0c7ce03b6ee17R131-R141 ?

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  private previousFilters: string;
+  public model: Model = { todIgnoreDst: true };
```
atruskie: do not abbreviate variable names

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+  public errors: {
+    todBoundaryError?: boolean;
```
atruskie: what is this used for?

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+    } else if (this.project) {
+      filter["projects.id"] = { eq: this.project.id };
+    }
+
+    this.setDateFilter(filter, model);
+    this.setTimeOfDayFilter(filter, model);
+
+    if (JSON.stringify(filter) === JSON.stringify({})) {
+      return null;
+    }
```
atruskie: i feel we could do much better than this.... count of own properties/

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    if (JSON.stringify({ filter }) !== this.previousFilters) {
+      this.previousFilters = JSON.stringify({filter });
+      return { filter };
+    }
```
atruskie: i wonder if we could solve this with some sort of structural equality

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+      return true;
+    }
+
+    // converting to a date object allows the comparison of dates using boolean operators
+    const startDate = new Date(formStartDate.year, formStartDate.month, formStartDate.day);
+    const endDate = new Date(formEndDate.year, formEndDate.month, formEndDate.day);
+
+    return startDate <= endDate;
+  }
+
```
atruskie: why aren't we using angular validations for these?

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+        `${this.formatter.format(model.dateStartedAfter)}T00:00:00.000Z`;
+    }
+
+    if (!this.isInvalidDate(model.dateFinishedBefore) && model.dateFinishedBefore) {
+      filter["recordedEndDate"] ??= {};
+      (filter["recordedEndDate"] as Comparisons).lessThanOrEqual =
+        // to allow users to input the same date as the starting and ending date, it is required that time is included
+        // if only the date is used (not with T24:00:00), it will default to T00:00:00, causing 1 day filters not to work
+        // if a time of day filter is specified, it will take priority over the time set in the date, causing no breakage in time filtering
+        `${this.formatter.format(model.dateFinishedBefore)}T23:59:59.000Z`;
```
atruskie: this is a dirty hack. Read up on intervals (https://en.wikipedia.org/wiki/Interval_(mathematics) ) and now that by default most natural language intervals are closed minimum and open maximum. Then you'll realise the filter can just be less than the end date and you won't need hacks like `23:59:59` ever again

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+      (!filterStartDate && !filterEndDate) ||
+      !this.dateIsInBoundary(filterStartDate, filterEndDate)
+    ) {
+      return;
+    }
+
+    if (!this.isInvalidDate(model.dateStartedAfter) && model.dateStartedAfter) {
+      filter["recordedDate"] ??= {};
+      filter["recordedDate"].greaterThanOrEqual =
+        `${this.formatter.format(model.dateStartedAfter)}T00:00:00.000Z`;
```
atruskie: seems like our formatter could emit the time parts yeah?

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (!this.isInvalidDate(model.dateFinishedBefore) && model.dateFinishedBefore) {
+      filter["recordedEndDate"] ??= {};
+      (filter["recordedEndDate"] as Comparisons).lessThanOrEqual =
+        // to allow users to input the same date as the starting and ending date, it is required that time is included
+        // if only the date is used (not with T24:00:00), it will default to T00:00:00, causing 1 day filters not to work
+        // if a time of day filter is specified, it will take priority over the time set in the date, causing no breakage in time filtering
+        `${this.formatter.format(model.dateFinishedBefore)}T23:59:59.000Z`;
+    }
+  }
+
```
atruskie: hint: the code written here was done before support for arrays was adding to the filtering API: https://github.com/QutEcoacoustics/baw-server/wiki/API:-Filtering

image

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+      .subscribe((model: Model): void => this.updateHref(model));
+  }
+
+  public updateHref(model: Model): void {
+    // validate that the dates are a valid format & are within bounds before sending an API filter request
+    // has the value of null if there is no update to the filters
+    const newFilters = this.generateFilters(model);
+
+    // if the new filters will be the same as the old filters, do not emit a filter update event
+    if (newFilters && this.model) {
```
atruskie: why are we testing if this.model is truthy?

src/app/components/shared/date-time-filter/filter.component.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (this.site) {
+      filter["sites.id"] = { eq: this.site.id };
+    } else if (this.region) {
+      filter["regions.id"] = { eq: this.region.id };
+    } else if (this.project) {
+      filter["projects.id"] = { eq: this.project.id };
+    }
+
+    this.setDateFilter(filter, model);
+    this.setTimeOfDayFilter(filter, model);
```
atruskie: expect to see an array condition here

src/app/components/shared/date-time-filter/filter.component.spec.ts
```diff
@@ -0,0 +1,535 @@
...
+        spectator.typeInElement(startTime, getTodStartedAfterInput());
+        spectator.typeInElement(endTime, getTodFinishedBeforeInput());
+        updateForm();
+
+        expect(getElementByInnerText(outOfBoundsTimeErrorMessage)).toExist();
+      }));
+    });
+  });
+
+  describe("filters events", () => {
```
atruskie: need to test both date range and time of day filters at the same time

atruskie: make most of these filter tests unit tests...

Perhaps even move filter creation out into it's own class?

atruskie (changes requested):
don't validate the number validate the format: use a regex to validate the year is at least 4 digits long. Something like `\d{4}\d*` - though `\d{4}` is good enough

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: `todBoundaryError` is set in the `setTimeOfDayFilter()` method and used as a predicate whether the start date is < than the end date (out of bounds date)

image

If the error `todBoundaryError` condition is true, the following error is emitted to the user.
image

In retrospective, we should either move this to a getter method (similar to how the date bound are validated), or add the date bound validations to this `errors` object.

In my opinion, exporting an filterErrors interface on the new proposed `Filters` class and adding the date boundary & format errors to this object is the most "correct" implementation and may prove useful in the future.

I will await your opinion though

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: I chose this approach because of two reasons

1. (From my knowledge) Objects are saved as references / pointers. Therefore,  would have to create and store two separate filter objects in memory; doubling memory (otherwise the when I update filters object, it would update previousFilters object too)

2. comparing two objects through traditional methods would compare the memory location not value e.g.

image

I could solve both these issues by instantiating two separate filter objects. But from my knowledge (may be incorrect), there is no quick way (that is less than `O(n)`) to key-value structurally compare two objects in JS

```js
// the following code doesn't work, but is an example of my idea of direct structural object equality
filter.entries(object1).forEach((key, value) => {
  if (this._previousFilter[key] !== value) {
    return;
  }
});

this._previousFilter = new Filter(filter)
return { filter }
```

Or we could compare hashes, but this has the same correctness issues as comparing by string (order matters and is still not a direct comparison).

Any ideas?

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: This method currently uses the NgBootstrap date formatter (it's not in the scope of the method)

However, adding the time to the date was a hacky trick that is no longer needed since we are no longer mixing date and time filter objects together (yay!), omitting time from date objects will default to `T00:00:00.00Z`, no longer mix date & time filters together & no longer need incorrect & hacky lines of code such as `T23:59:59.000Z`

src/app/components/shared/date-time-filter/filter.component.ts
atruskie: OK; while your analysis of iteration performance might be correct it doesn't really matter when n is counted in tens (~8 keys, sometimes 20).

The other thing to keep in mind is that json.stringify does not emit keys in a particular order. 
Trivial failure example:

```js
a = { b: 2, c: 3}
b = { c: 3, b: 2}
console.log(JSON.stringify(a) == JSON.stringify(b)) // false !
```

I feel the overhead of serializing an object just to get structural equality is more than any key count consideration.

> Therefore, would have to create and store two separate filter objects in memory; 

Your current implementation is storing the previous filters as a string. That's (bullshit guess) 3 to 4 times the memory of a raw object.

> comparing two objects through traditional methods would compare the memory location not value e.g.

Yes, that's what we call reference equality: comparing references to pointers.

> . But from my knowledge (may be incorrect), there is no quick way (that is less than O(n)) to key-value structurally compare two objects in JS

N isn't really the issue. 
There are implementations like: https://www.npmjs.com/package/lodash.isequal
But I think we already use immutablejs and it feels like it should support this. https://immutable-js.com/#nested-structures

image

But I'll tell you what, let's [measure it](https://jsbench.me/), not speculate it.

- compare ImmutableJS Maps
- to JSON.stringify (with ordered keys https://stackoverflow.com/a/64565203)

For a set of filters of various sizes.

atruskie: This is _not_ a fair test, but it is a good approximation:

image

src/app/components/shared/date-time-filter/filter.component.ts
atruskie: I am commenting on the naming. Suggested name: `timeOfDayStartNotLessThanEnd`. There are 6 other boundary conditions I can think of and the old name could refer to any of them.

But yeah,  like I said elsewhere kinda weird to see this here, I would have expected to see these entered as validations in the form.

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: Hmm, good points

Immutable.js's `.equals` method looks like it will be a very nice solution I skimmed over while solving this issue

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: This seems to have been an artifact from an old version of my code that looked something like this:

image


I have since removed it

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: This is a good point, I'll update the code accordingly

hudson-newey: I've had a play around with angular form validations for these and I can transfer the `dateIsInBoundary` method to ng form validation, but not the isInvalid date.

This is because of some trickery to do with how the bootstrap directives work.

When it does not match bootstrap form validation (`[0-9]+-\d{2}-\d{2}`), it returns a string, however, when it matches the RE above, it returns an `NgbDate` object. This means that using ngForm validation attributes such as `pattern="\d{4}-\d{2}-\d{2}"` does not work (because it's trying to compare to an object), and `minlength="10"` will return true even when the date (in string format) is < 10 characters because the object will be > 10.

src/app/components/shared/date-time-filter/filter.component.ts
hudson-newey: I have recreated the filter methods (below) using array conditions.

These helper methods are subject to change as I have already found one bug in them (and it is messy code), however, it provides insight into my approach to this solution using helper methods

image

hudson-newey:
~~This PR will also naturally fix another bug that I found while exploring the old client~~

~~The old client seems to be comparing the `startTime >= recordingEndTime` and `endTime <= recordingStartTime` in [these lines of code](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/audio-recordings/pages/download/download.component.ts#L195-L209)~~

~~This sometimes caused some recordings like the image below to be returned when they should not have been~~
image

hudson-newey:
Before review, please note that I have used [this](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-543caae93c289e882b1e8983185c25c0a4df75e865889c2e7a28d33a217667e8R81-R88) _slightly hacky_ block of code has remained and I have documented it's purpose

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
```diff
@@ -0,0 +1,169 @@
...
+    // since we are using angular form validation & bootstrap validation, the errors attribute of the form is updated with change detection
+    // this causes the form state to update with change detection so we need to add an extra change detection cycle
+    // to ensure the updated form state (with errors) is a part of the model at end of change detection
+    this.changeDetector.detectChanges();
+  }
+
+  public emitFilterUpdate(model: Model): void {
+    // we should only send new filter requests when the user has not input any "bad" / incorrect values into the input fields
+    // e.g. 2020-31-31 is not a valid date should display an error, and not send a new filter request
+    if (this.form.status !== "VALID") {
```
atruskie: is this constant defined anywhere? avoid magic values

exception: typescript type checks constrained string types so if it is not defined anywhere as a constant then this is fiine

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
```diff
@@ -0,0 +1,169 @@
...
+      return;
+    }
+
+    const newFilters = this.generateFilters(model);
+
+    // if the newFilters are null (representing there is no filter update), then do not emit a new filter event
+    if (newFilters) {
+      this.constructedFilters.next(newFilters);
+    }
```
atruskie: this API is messy

`const [changed, newFilters] = this.generateFilters(model)` - you could return changed state in a tuple

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
```diff
@@ -0,0 +1,169 @@
...
+    const modelEndTime = model?.timeFinishedBefore;
+
+    // we can save ourselves from calculating a new time filters (improving performance) if the user hasn't input a start time or end time
+    if (!model?.timeFiltering || (!modelStartTime && !modelEndTime)) {
+      return filters;
+    }
+
+    const timezoneExpressions: DateExpressions[] = model.ignoreDaylightSavings
+      ? ["local_offset", "time_of_day"]
+      : ["local_tz", "time_of_day"];
```
atruskie: this is filter detail that should belong in your filter builder method

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @returns A new filter with all the same conditions as the initial filter, with the date range conditions added
+ */
+export function addDateFilters(
+  startDate?: DateTime,
+  endDate?: DateTime,
+  filters?: InnerFilter
+): InnerFilter {
+  if (startDate) {
+    // the baw-api is expecting the date in the format (YYYY-MM-DD)
+    // ISO date format is used as it emits the date in the correct format and explicitly emits the time as 00:00:00
```
atruskie:
```suggestion
    // the baw-api is expecting the date in the ISO8601 format.
```

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,78 @@
...
+    const formattedStartDate = startDate.toISO();
```
atruskie: I'm surprised you have to do this; luxon DateTime implements `toJSON` itself which emits ISO8601 dates... 
image

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,78 @@
...
+    const startDateFilter: InnerFilter<AudioRecording> = {
+      recordedDate: { greaterThan: formattedStartDate },
```
atruskie: incorrect operator

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (endDate) {
+    const formattedEndDate = endDate.toISO();
+    const endDateFilters = {
+      recordedEndDate: { lessThanOrEqual: formattedEndDate },
+    };
+
+    filters = addFilter(filters, endDateFilters);
+  }
+
+  return filters;
```
atruskie: same points as above

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @param expressions Describes if the time filters use local (daylight savings) time
+ * @param startTime (optional) The minimum `recordedDate` allowed
+ * @param endTime (optional) The maximum `recordedEndDate` allowed
+ * @param filters (optional) An existing filter to add the time range conditions to. If no filter is provided, a new filter will be created
+ * @returns
+ */
+export function addTimeFilters(
+  expressions: DateExpressions[],
+  startTime?: string,
+  endTime?: string,
```
atruskie: weird that these are formatted as string when the other method takes a well-typed construct (`DateTime`) though I see luxon has no time only type... soo I guess there's no action to take here


maybe validate input value is time by regex??

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,78 @@
...
+  filters?: InnerFilter
+): InnerFilter {
+  if (startTime) {
+    const startTimeFilter = {
+      recordedEndDate: {
+        greaterThanOrEqual: { expressions, value: startTime },
```
atruskie: incorrect operator

src/app/helpers/filters/filters.ts
```diff
@@ -0,0 +1,63 @@
+import { InnerFilter } from "@baw-api/baw-api.service";
+import { isInstantiated } from "@helpers/isInstantiated/isInstantiated";
+import { AbstractModel } from "@models/AbstractModel";
+
+/**
+ * Adds a new filter condition to an existing filter
```
atruskie: it only add a filter to the root of a filter, and only in an and expression

src/app/helpers/filters/filters.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param filter The current existing filter conditions
+ * @param additionalFilter A new condition to add to the existing filter
+ * @returns A new filter which satisfies the intersection between the two filters
+ */
+export function addFilter(
```
atruskie: suggested naming convention:

`filterAnd` and usage would be `filterAnd(filter, newFilter)`

Carry that convention onto your other helper methods

atruskie (changes requested):
ok, while we're here lets sneak in support for times of day that span over midnight. i.e. "find all recordings after 8pm and until 4am"

the comparison operators for dates are incorrect, the lower value should be inclusive. The upper value exclusive.

okay a lot of neat changes here, but there's a few logic errors we can't pass

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
hudson-newey: It's a standard, but not a strongly contained TS string

image

src/app/helpers/filters/audioRecordingFilters.ts
hudson-newey: Time input is already validated through angular form validation on the `time.component.ts`, and is completely detached from the filter component

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
hudson-newey: An expansion upon this would be renaming `newFilters` to `currentFilters` and always emitting the current filter, but also emitting the `changed` status in the tuple.

At the moment; in my revision, I have retained the `newFilter` emitting `null` if there are no new filter changes

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
atruskie: from the snippet you showed me it is not clear that this is a literal string type, only a string, which means it is not constrained and type checked. Maybe the TS compiler is smarter? But I can't tell that from a code review.

https://github.com/angular/angular/blob/1d761bd7c74ac2b7dfe5f7dba0ff16fd6170afe2/packages/forms/src/model/abstract_model.ts#L67

src/app/helpers/filters/audioRecordingFilters.ts
atruskie: You miss the point: many well-typed structures serialize well by default. For example the Luxon DateTime object implements toJSON and emits an ISO8601 value automatically and also normalizes the input date.

This has nothing to do with validation.

hudson-newey:
@atruskie Changes since last review:

Form validation now uses [this.form.valid](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-543caae93c289e882b1e8983185c25c0a4df75e865889c2e7a28d33a217667e8R92)

The [`generateFilters()`](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-543caae93c289e882b1e8983185c25c0a4df75e865889c2e7a28d33a217667e8R103-R124) method now returns an array of `[changedState, newFilters]` and is used [here](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-543caae93c289e882b1e8983185c25c0a4df75e865889c2e7a28d33a217667e8R96-R100)

[Filter](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-8c0a96dc91803e0825235f8d388291fac65ecb83308e7604f8231942ef4ed708R12-R63) and [audio recording filter](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R14-R79) helper methods are now renamed to reflect their purpose.

[Start](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R22) and [end](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R30) time operators now use the correct closed and unclosed boundary conditions. [tests](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-b3b5715c341a2ba2bb5c4d9cb17ead5901f2be661eb5d91458685f1d5578e92dR618-R672) have been adjusted to reflect this change)

[Start](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R22) and [end](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R30) date filter helper functions now use Luxon's `.toJSON()` method

Daylight savings filter information is now added in the `audioRecordingFilter.ts` [time filter](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R54-R56) helper functions (this used to be done in the `audio-recordings-filter.component.ts`)

The time filter helper function now allows a [special condition](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-896def894b04ad1419922ed4e4678edcbd695b4519278d1b45edd75f1a648f16R63-R89) for filters past midnight.

The [time range tests](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-703bbaf9198ed685b2f23360dd66f1e7847f76071a1d7c16231ae43e128381dbR115-R147) now have the same input and output conditions as the [baw-api tests](https://github.com/QutEcoacoustics/baw-server/blob/b5a9f807d67a414610c665a499f39d50aa37d423/spec/requests/audio_recordings_spec.rb#L195-L208)

[Midnight time tests](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-703bbaf9198ed685b2f23360dd66f1e7847f76071a1d7c16231ae43e128381dbR149-R198) were created to match the [baw-api midnight tests](https://github.com/QutEcoacoustics/baw-server/blob/b5a9f807d67a414610c665a499f39d50aa37d423/spec/requests/audio_recordings_spec.rb#L281-L325).  
The same test conditions & assertions are also performed in the [`recordings-filter-component.spec.ts` tests](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-b3b5715c341a2ba2bb5c4d9cb17ead5901f2be661eb5d91458685f1d5578e92dR503-R563) to assert the correct filter update is emitted (_the link above may not work because the diff is large, it is collapsed by default. The test is named "an filter recordings past a day boundary with a composition of filters, 10PM to 1 AM with ignore day light savings" in `audio-recordings-filter.spec.ts` from lines 503-563_)

**Additional Change:**
The audio recordings list page now uses the `FilterAdd()` method for it's [`apiAction`](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-dc2d8f986bfbc9dcb229f29bbfff0e870eb6bf985baf35e4b2c0c7ce03b6ee17R149-R173) method

src/app/components/audio-recordings/pages/list/list.component.ts
```diff
@@ -143,6 +143,11 @@ class AudioRecordingsListComponent
       ...incomingFilters,
     };
 
+    // since paging is handled by the pagedTableTemplate, new filters always generate from page 1
+    // to ensure this is reflected client side, reset the page number to the first page (0), page index's start from 0 as page 1
+    // TODO: we should be able to remove this hack once https://github.com/QutEcoacoustics/workbench-client/issues/2054 has been resolved
+    this.pageNumber = 0;
+
```
atruskie: No, you just shouldn't add this hack to begin with. Remove it.

src/app/components/shared/date-time-filter/audio-recordings-filter.component.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+      const startDate = "2021-03-15";
+      const endDate = "2022-03-18";
+
+      toggleDateFilters();
+      typeInElement(getDateStartedAfterInput(), startDate);
+      typeInElement(getDateFinishedBeforeInput(), endDate);
+
+      toggleDateFilters();
+
+      expect(filterChangeSpy).toHaveBeenCalledWith({ filter: undefined });
```
atruskie: generally i prefer empty objects and empty arrays for describing nothing set, rather than null or undefined.

src/app/components/shared/date-time-filter/audio-recordings-filter.component.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+                  },
+                  {
+                    recordedEndDate: {
+                      lessThanOrEqual: { expressions, value: endTime },
+                    },
+                  },
+                  {
+                    recordedDate: {
+                      lessThan: { expressions, value: endTime },
+                    },
```
atruskie: redundant expression

src/app/components/shared/date-time-filter/audio-recordings-filter.component.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    const [changed, newFilters] = this.generateFilters(model);
+
+    if (changed) {
+      this.constructedFilters.next(newFilters);
+    }
+  }
+
+  private generateFilters(model: Model): [boolean, Filters] {
```
atruskie:
```suggestion
  private generateFilters(previous: Filters, model: Model): [boolean, Filters] {
```

embrace stateful side effects or functional methods with no side effects - don't mix the concepts

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,113 @@
...
+  filters?: InnerFilter
+): InnerFilter {
+  const expressions: DateExpressions[] = ignoreDayLightSavings
+    ? ["local_offset", "time_of_day"]
+    : ["local_tz", "time_of_day"];
+
+  // a midnight filter is a filter circumstance where the time range overlaps two days e.g. 10PM to 1AM
+  // midnight filters can only occur if both a start and end time are specified
+  // therefore, this statement should return false if both start and end times are not specified
+  const isMidnightFilter = startTime && endTime && startTime > endTime;
```
atruskie, line 69: this is incorrect - you're doing string comparisons on numbers. Thinking about this, it might hold actually (because the strings are always the same length) but regardless this is really bad practice.

src/app/helpers/filters/audioRecordingFilters.ts
atruskie: so start time and and time should be luxon durations i think

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,113 @@
...
+        {
+          recordedEndDate: {
+            lessThanOrEqual: { expressions, value: endTime },
+          },
+        },
+        {
+          recordedDate: {
+            lessThan: { expressions, value: endTime },
+          },
+        },
```
atruskie: this is a redundant expression with respect to the second expression

src/app/helpers/filters/audioRecordingFilters.ts
```diff
@@ -0,0 +1,113 @@
...
+      };
+
+      filters = filterAnd(filters, startTimeFilter);
+    }
+
+    if (endTime) {
+      const endTimeFilter = {
+        recordedDate: {
+          lessThanOrEqual: { expressions, value: endTime },
```
atruskie: after our call, we've decided that this should be less than.

src/app/helpers/filters/audioRecordingFilters.ts
atruskie, line 1: leave comments explaining why these condition are chosen - especially since they're so complicated

Main reason "we're looking for any overlap within our interval, but we don't want to include things that just touch on the ends"

src/app/helpers/filters/filters.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export function filterModel<T extends AbstractModel, U extends AbstractModel>(
+  key: string,
+  model: T,
+  currentFilter?: InnerFilter
+): InnerFilter<Writeable<U>> {
+  // all model filters condition on the id attribute. While it is very rare for a model to not have an id, it is possible
+  // this bailout is typically evoked if the model is undefined
+  if (!isInstantiated(model?.id)) {
+    return currentFilter;
```
atruskie, line 54: i would suggest returning an empty inner filter e.g. `{ }` denotes that no filters are applied

src/app/helpers/filters/filters.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param key The key identifier of the model, most likely the type
+ * @example key = "projects"
+ * @param model The model to filter by
+ * @param currentFilter (optional) A filter that model condition will be added to. If no filter is provided, a new filter will be created
+ * @returns A new filter with the model condition added
+ */
+export function filterModel<T extends AbstractModel, U extends AbstractModel>(
+  key: string,
+  model: T,
+  currentFilter?: InnerFilter
```
atruskie:
```suggestion
  currentFilter: InnerFilter
```

src/app/helpers/filters/filters.ts
atruskie: based on what you said this doesn't need to change because tests are artificially creating undefined filters

atruskie (changes requested):
so there's a little inconsistency with the date filters. The basic question we have to answer is what do our users want?

I'd generally say anyone who is filtering things is looking for the most data that satisfies their filters, which would include overlaps with the filter boundaries.

So that means for date filtering:

- start date: find any recordings that end after the start date
  `recordedEndDate > startFilterDate`
- end date: find any recordings that start before the end date
  `recordedDate < endFilterDate`

hudson-newey:
> So that means for date filtering:
> 
> * start date: find any recordings that end after the start date
>   `recordedEndDate > startFilterDate`
> * end date: find any recordings that start before the end date
>   `recordedDate < endFilterDate`

I'm assuming that the underlying logic for time filters should not be changed and will have the following logic

start time: `recordedEndTime > startFiltertime`
end time: `recordedStartTime <= endFilterTime`

image

atruskie (changes requested):
the case converter changes are not acceptable, revert them. Transforms should never mutate the original object.

hudson-newey:
> the case converter changes are not acceptable, revert them. Transforms should never mutate the original object.

@atruskie This problem has been resolved through [this code](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-e425c5f0f250fc4f3c04a30f430030e34fec00e61ec0a755c0e8c0fed637f7f0R54-R62)

I have however retained my case converter [idempotent unit test](https://github.com/QutEcoacoustics/workbench-client/pull/2030/files#diff-074af3367d976785fbf53c565bf48b4a24a45528beac7809b12a9c21d50cabf7R147-R156) as I believe this is a useful addition.

I am currently testing to code to ensure there are no regressions

atruskie (approved):
OK I think this is finally ready to go; though there are some failing tests on the CI

hudson-newey:
> OK I think this is finally ready to go; though there are some failing tests on the CI

These tests are failing because in a historical commit all models were added to filter conditions. This is obviously incorrect and redundant so this change was reverted.

I've reverted the tests to their old state (before the incorrect change) to fix the CI failures.

PR #2033, opened 2023-01-25
Harvest UI tweaks

### LINKED ISSUE #1970: Harvest: metadata review screen UI tweaks
UI tweaks suggested from user demo:

- [x] Make sure folder guides align on narrow page widths

Image

- [x] should not show expand chevron for one line problems

Image

- [ ] Show inherited values for mappings  instead of a change button

- [x] Change" Change Site or UTC for folder " to "Add" 
      (Combined with the inherited option from above, "change" should never be needed)
- [x] Do not show root mapping creation button unless there are harvest items in the root directory
- [x] Add tip into the instruction section that says "You don't have to change anything for files that have no problems"
- [x] "Upload more files" button should change to "Change or add more files"
- [x] Duplicate file tip should and a hyperlink to  "Change or add more files" button, and make terminology consistent.
  ```diff
  -You should go back to the upload stage and delete the duplicates.
  +You can go back and [change the files] to delete the duplicates
  ```
- [x] Differentiate between duplicates found in _this_ "current upload" and "previous uploads" in tips
- [x] Show correct site/point terminology in site column
- [x] Show site `id` on tooltip hover for select sites
- [x] UTC offset selector: have common values repeated near the top. E.g. 
   For site in NSW:

   ```
   +10:00
   +11::00
   ---
   <normal list of offsets here>
   ```

hudson-newey:
# Harvest UI tweaks

Following a user demo, users expressed that the Harvest UI needed some adjustments

## Changes

* Fixes bugs in folder grid alignment
* Drop down chevrons are now not shown if there is no expandable content
* Mappings are now inherited from their parent folders
* Root mapping button is now not shown if there are no Harvest items directly underneath it
* Changes terminology for a few UI elements
  * "Change mappings and UTC offset" is now "Add mappings and UTC offset"
  * Adds sidebar tip "You don't have to change anything for files that have no problems"
* Adds action buttons / hyperlinks to action on tips in sidebar
* Adds an extra tip to differentiate between duplicate files in the current upload and previous uploads
* Shows the site ID in a tooltip for mappings when the user hovers over
* UTC offset now shows relevant offsets to the site location at the top of the dropdown box
* Creates tests for folder row component
* Creates tests for file row component
* Creates tests for UTC offset component
* Adds tests to metadata review component to ensure mappings are inherited correctly
* Updates dependency `@angular-devkit/build-angular`
* Removes deprecated usage of `~` in scss imports
* Modified association decorator to not always use cached model. It now checks that the cached model has the same id as the requested model before returning the cached model. If the two id's do not match, the model is fetched from the API, and the cached model is updated (**Currently failing tests**)

## Problems

**This is a draft pull request, I am still debugging, refining, and searching for new bugs.**

## Issues

Fixes: #1970 

## Visual Changes

**Site id tool tip on hover:**
image

**Suggested UTC offsets in dropdown box:**
image

**Chevron does not show for single line errors:**
image

**New sidebar terminology & action hyperlinks:**
image

## Final Checklist

### DIFF
--- a/src/app/components/harvest/components/metadata-review/file-row.component.ts
+++ b/src/app/components/harvest/components/metadata-review/file-row.component.ts
@@ -1,6 +1,7 @@
 import {
-  ChangeDetectionStrategy,
+  AfterViewInit,
   Component,
+  ElementRef,
   Input,
   OnInit,
   ViewChild,
@@ -41,7 +42,7 @@ interface ValidationMessage {
       <div
         *ngIf="
           harvestItem.hasItemsInvalid &&
-          (shouldShowChevron(validationsContainer) || row.showValidations)
+          (areValidationsExpandable || row.showValidations)
         "
         class="dropdown-icon"
       >
@@ -93,16 +94,16 @@ interface ValidationMessage {
     </div>
   `,
   styleUrls: ["file-row.component.scss"],
-  // Nothing in this component can change without a change in the row
-  changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class FileRowComponent implements OnInit {
-  @Input() public row: MetaReviewFile;
+export class FileRowComponent implements OnInit, AfterViewInit {
   @ViewChild("validationsContainer")
-  public validationsContainer: HTMLDivElement;
+  public validationsContainer: ElementRef<HTMLDivElement>;
+  @Input()
+  public row: MetaReviewFile;
 
   public validationMessages: ValidationMessage[];
   public icons = metaReviewIcons;
+  public areValidationsExpandable = false;
 
   public get mapping(): HarvestMapping {
     return this.row.mapping;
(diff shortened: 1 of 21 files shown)


src/app/components/harvest/components/metadata-review/folder-row.component.ts
```diff
@@ -146,18 +150,38 @@ export class FolderRowComponent {
...
   }
 
+  public get hasHarvestItems(): boolean {
+    return !this.mapping && !(this.row.isRoot && !+this.row.harvestItem);
+  }
+
   public constructor(private injector: Injector) {}
 
+  public ngOnChanges(): void {
+    // to inherit mappings values among ancestor items, if the parent item gains a Harvest Mapping, then all the child items should too
```
hudson-newey: _This method is currently faulty_  
Since folder rows only have knowledge of their parents, it is not possible to make rows inherit their parents mappings before they are shown on the DOM.

Therefore, I either need to change this.rows to be a doubly linked list (with knowledge of the children, so that I can recurse down the tree when a mapping is changed on a parent item), or move this logic to where the mappings change request is sent to the api (where it does have knowledge of the full model).

src/app/components/harvest/screens/metadata-review/metadata-review.component.spec.ts
```diff
@@ -355,4 +355,7 @@ describe("MetadataReviewComponent", () => {
     discardPeriodicTasks();
   }));
 
+  it("should inherit mappings from parent item if harvest item mappings are not set", fakeAsync(() => {
```
hudson-newey: This unit test needs to be completed

src/app/components/harvest/components/metadata-review/file-row.component.ts
```diff
@@ -84,13 +93,19 @@ interface ValidationMessage {
         ></fa-icon>
       </div>
     </div>
+    <ng-container *if="updateChevronState(validationsContainer)"></ng-container>
```
hudson-newey: This is a hacky workaround I had to implement because change detection wasn't firing for `*ngFor` changes

hudson-newey (commented):
_This PR is currently a WIP_

What I have to complete before review

src/app/components/harvest/components/metadata-review/file-row.component.ts
hudson-newey: This has now been resolved. It seems to have been an issue with the logic in my `updateChevronState` method

hudson-newey:
I'll refactor this into some atomic PR's so that it's easier to review and so that I can review the work I did in this PR a few years ago.

Closing although I will use it as a base for other PR's.

PR #2034, opened 2023-01-30
Unique Route titles

### LINKED ISSUE #1980: Unique title per route
The title of the website is always `config.settings.brand.short`. This means that when users are attempting to navigate back multiple pages, it is impossible to determine which step was which:

image

Angular 14 has introduced setting the title of the current route based off the current route using a `TitleStrategy`: https://blog.angular.io/angular-v14-is-now-available-391a6db736af.

hudson-newey:
# Unique Route titles

At the moment, navigating through the workbench client does not change the user agent title.

Users have expressed this as a point of issue when they are navigating through their browser history trying to find a web page.

### e.g.

**Navigating through the website using the back navigation button**
image
  
**Trying to find a page through browser history**
image

## Changes

* Adds the `title` callback method to component page info
* Adds unit tests for page routes at the app level, not component level. Tests for route titles at the component can be completed at request, however, this will require an estimated 1000 lines of code, and I do not see a beneficial ROI for this to be tested by CI.
* Adds a `TitleStrategy` injectable at the app level

## Problems

None

## Issues

Fixes: #1980 

## Visual Changes

**For components directly under the root app (example.com/projects) `<<BrandName>> | Projects`:**
image

**For nested components (example.com/projects/123) `<<BrandName>> | Projects | 2022 Symposium Example`:**
image

**For the root app `<<BrandName>>`:**
image

For components that do not have a title, it will construct the title to the lowest level possible (example.com/projects/321/audio-recordings/1) (in a hypothetical case where audio-recordings component does not have a title specified in its `PageInfo`) `<<BrandName>> | Projects | 321`

## Additional Information

We may want to consider using the Angular `TitleCasePipe` for each title. This pipe ensures a string has a capital letter after each space. However, I have decided against it because I believe that it will cause titles to not be exactly the same as components.

e.g. a project named "hello world" would have the browser title "Hello World"

## Final Checklist

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -1,4 +1,3 @@
-import { Title } from "@angular/platform-browser";
 import { RouterTestingModule } from "@angular/router/testing";
 import { MockBawApiModule } from "@baw-api/baw-apiMock.module";
 import { Writeable } from "@helpers/advancedTypes";
@@ -16,8 +15,9 @@ import {
   SpectatorRouting,
 } from "@ngneat/spectator";
 import { LoadingBarComponent, LoadingBarService } from "@ngx-loading-bar/core";
-import { ConfigService } from "@services/config/config.service";
+import { Title } from "@angular/platform-browser";
 import { MenuService } from "@services/menu/menu.service";
+import { BreadcrumbComponent } from "@shared/breadcrumb/breadcrumb.component";
 import { SharedActivatedRouteService } from "@services/shared-activated-route/shared-activated-route.service";
 import { FooterComponent } from "@shared/footer/footer.component";
 import { HeaderComponent } from "@shared/menu/header/header.component";
@@ -26,7 +26,13 @@ import { generatePageInfo } from "@test/fakes/PageInfo";
 import { generatePageInfoResolvers } from "@test/helpers/general";
 import { MockComponent } from "ng-mocks";
 import { Subject } from "rxjs";
-import { AppComponent } from "./app.component";
+import { ConfigService } from "@services/config/config.service";
+import { RouterStateSnapshot } from "@angular/router";
+import { modelData } from "@test/helpers/faker";
+import { generateMenuRoute } from "@test/fakes/MenuItem";
+import { MenuRoute } from "@interfaces/menusInterfaces";
+import { AppComponent, PageTitleStrategy } from "./app.component";
+import { CommonRouteTitles } from "./stringConstants";
 
 describe("AppComponent", () => {
   const eventSubject = new Subject<IPageInfo>();
@@ -41,10 +47,13 @@ describe("AppComponent", () => {
       MockComponent(PrimaryMenuComponent),
       MockComponent(SecondaryMenuComponent),
       MockComponent(ActionMenuComponent),
+      MockComponent(BreadcrumbComponent),
     ],
     providers: [
       { provide: DEFAULT_MENU, useValue: mockDefaultMenu },
       mockProvider(LoadingBarService, { value$: new Subject() }),
+      Title,
+      PageTitleStrategy,
     ],
     imports: [RouterTestingModule, MockBawApiModule],
   });
@@ -84,13 +93,6 @@ describe("AppComponent", () => {
   it("should create loading bar", () => {
     spec.detectChanges();
     expect(spec.query("ngx-loading-bar")).toBeTruthy();
-  });
-
-  it("should change the title to match the environment title", () => {
-    const title = spec.inject(Title);
-    const config = spec.inject(ConfigService);
-    spec.detectChanges();
-    expect(title.getTitle()).toBe(config.settings.brand.short);
   });
 
   describe("side nav", () => {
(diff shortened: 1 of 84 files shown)


atruskie:
Might as well test the route titles for each component. See if you can find ways to cut down the crud. Make a helper function? Most of these pages already have fakes ready to go...

I saw in one of your screenshots that 'projects' was not capitalised... Probably should be capitalised.

hudson-newey:
Thanks for your input @atruskie, at the moment all of the components are all lowercase; I will be able to adjust this easily.

I'll also have a look into how I can use a helper function to test each component's page info / title.

hudson-newey:
There is some "less than optimal" code that I'm using in the tests.
e.g. [https://github.com/QutEcoacoustics/workbench-client/pull/2034/files#diff-b4ce2a94a9d3840eff713a0c671c5706d4a3238538dbd09d9aa7d28258d4e195R125-R160](https://github.com/QutEcoacoustics/workbench-client/pull/2034/files#diff-b4ce2a94a9d3840eff713a0c671c5706d4a3238538dbd09d9aa7d28258d4e195R125-R160).

I'll be attempting to move this into the `assertPageInfo` helper function [here](https://github.com/QutEcoacoustics/workbench-client/pull/2034/files#diff-ac84e2f0dd198ae2f77c52a5f39fa7e5612b6ea1d8f10db057d7b08cfd1df296R5-R20)

src/app/app.component.spec.ts
```diff
@@ -207,5 +205,94 @@ describe("AppComponent", () => {
...
+            },
+          },
+        },
+      });
+
+      titleStrategyInjectable.updateTitle(mockRouterState);
+
+      const appTitle = titleService.getTitle();
+
+      expect(appTitle).toEqual(configService.settings.brand.short);
```
atruskie: is this the best default behaviour?

Perhaps we can use the menu item's category as an interim label?

src/app/app.component.ts
```diff
@@ -83,3 +81,53 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
...
+
+  /**
+   * Recursively iterates through the nested components to construct a title
+   * This is needed for components that are mounted in multiple locations
+   *
+   * @param componentState The component as defined in the route data
+   * @param routeState a `RouterStateSnapshot` representing the router state. This value should not change throughout calls.
+   * @returns A `string` representing the breadcrumb steps that the user took
+   */
+  private constructBreadcrumbTitle(componentState, routeState: RouterStateSnapshot): string {
```
atruskie: avoid the term _breadcrumb_ - i'd say it is misleading in this context

src/app/app.component.ts
```diff
@@ -83,3 +81,53 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
...
+    const componentPageInfos = getPageInfos(componentState?.route.pageComponent);
+
+    // in the edge case where a page doesn't have page info, or a title stop searching through
+    // after it comes across a component that does not have a title and use the highest level of actitlescuracy possible
```
atruskie: `actitlescuracy`?

src/app/app.component.ts
```diff
@@ -83,3 +81,53 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
...
+    if (!componentPageInfos || !componentPageInfos[0]?.title) {
+      return "";
+    }
```
atruskie: Ok if something is missing a title, I think that:

a. We shouldn't give up constructing the rest of the title.
b. We should use a fallback (like category). Even empty title segments might be clearer.

src/app/app.component.ts
```diff
@@ -83,3 +81,53 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
...
+
+    const componentPageInfo = componentPageInfos[0];
+
+    // the Angular title case pipe ensures that all titles start with a capital letter after each space
+    const componentTitle = componentPageInfo.title(routeState);
```
atruskie: my gut feeling is that this will cause a problem somewhere down the line. But I don't mind it  for now until we see that it  causes a problem

src/app/components/sites/pages/wizard/wizard.component.ts
```diff
@@ -86,6 +86,7 @@ WizardComponent.linkToRoute({
   pageRoute: newSiteMenuItem,
   menus: { actions: List(projectMenuItemActions) },
   resolvers: { [projectKey]: projectResolvers.show },
+  title: () => "New Region / Site",
```
atruskie:
```suggestion
  title: () => "New Site or Point",
```

Remember, the term _region_ is a technical concept - it should be exposed nowhere in the UI.

src/app/components/sites/pages/edit/edit.component.ts
```diff
@@ -84,6 +84,7 @@ SiteEditComponent.linkToRoute({
     [projectKey]: projectResolvers.show,
     [siteKey]: siteResolvers.show,
   },
+  title: () => "Edit",
```
atruskie: Seems like we'd probably use the title "edit" a few times. Refactor common titles into some constants?

src/app/components/sites/pages/delete/delete.component.ts
```diff
@@ -84,6 +84,7 @@ SiteDeleteComponent.linkToRoute({
     [projectKey]: projectResolvers.show,
     [siteKey]: siteResolvers.show,
   },
+  title: () => "Delete",
```
atruskie: add to list of common titles

src/app/components/regions/pages/new/new.component.ts
```diff
@@ -71,6 +71,7 @@ NewComponent.linkToRoute({
   resolvers: {
     [projectKey]: projectResolvers.show,
   },
+  title: () => "New Region",
```
atruskie: _region_ is not a UI term:

```suggestion
  title: () => "New Site",
```

src/app/components/regions/pages/list/list.component.ts
```diff
@@ -79,6 +79,7 @@ ListComponent.linkToRoute({
   category: shallowRegionsCategory,
   pageRoute: shallowRegionsMenuItem,
   menus: { actions: List(regionsMenuItemActions) },
+  title: () => "Regions",
```
atruskie:
```suggestion
  title: () => "Sites",
```

src/app/components/projects/pages/details/details.component.spec.ts
```diff
@@ -43,9 +45,16 @@ describe("ProjectDetailsComponent", () => {
     imports: [SharedModule, MockBawApiModule],
     declarations: [mock.map, mock.card],
     component: DetailsComponent,
+    providers: [PageTitleStrategy],
   });
   const emptyResponse = [[]];
 
+  assertPageInfo(DetailsComponent, "test name", {
+    project: {
+      model: new Site(generateSite({ name: "test name" }))
```
atruskie: this is a project details page? Why are we generating a site model?

src/app/test/helpers/page-route.ts
```diff
@@ -0,0 +1,40 @@
+import { Type } from "@angular/core";
+import { RouterStateSnapshot } from "@angular/router";
+import { getPageInfos, IPageComponent } from "@helpers/page/pageComponent";
+import { PageInfo } from "@helpers/page/pageInfo";
+
+export function assertPageInfo<T>(
```
atruskie: document this.

Something subtle I just noticed is you're only testing these routes at their top level.

E.g. for harvest details page your test is essentially:

`test title is "test name"`

and not

`test title is "Ecosounds | Projects | Samford | Uploads | test name"`.

The approach you've taken is you don't need to test every segment in the title if you have tests for every title in the parent hierarchy by itself.
However, in theory a title generator can use any of the route data available to it to generate a title. So that is never tested currently. 

At the minimum document about these limitations and how these tests do not reflect actual generation of titles.

atruskie (approved):
OK, looks really good.

Just needs one more cleanup pass.

Strive to deduplicate titles. Define some constants in the menu files for each page that you can reuse.

src/app/app.component.ts
hudson-newey: Seems to be a typo for Title Accuracy

src/app/app.component.ts
hudson-newey: I have renamed this method to the most descriptive `buildHierarchicalTitle` name

src/app/components/projects/pages/details/details.component.spec.ts
hudson-newey: This is a good catch. Surprisingly, test tests didn't fail because of it.

I'll probably have to add some form of type checking to the page info / route title assertion tests to ensure this gets picked up in the future

_Edit:_
I was able to make this helper method strongly typed through the following code. This should prevent similar issues in the future

image

src/app/components/sites/pages/delete/delete.component.ts
hudson-newey: With the addition of delete modals in #2042, there will be no need for the route title anymore

src/app/test/helpers/page-route.ts
hudson-newey: There's an assertion that the correct route data and page segments are used to construct the route title [here](https://github.com/QutEcoacoustics/workbench-client/pull/2034/files/38f8dbeb1411fd4e81459797451ef77324a634a1#diff-002e1a20416b0120fbe06390465dd02fb2d00c771c5ae171766211daef759308R244-R270)

However, this test does not cover pages with multiple mounting points e.g. sites / point details component, I'll add another assertion to this tests helper method to assert the correct title is used at the different mounting points.

src/app/app.component.spec.ts
hudson-newey: I've implemented the use of the menu item category as a fallback label

By removing the `title` attribute from the project details page `PageInfo`, I was able to produce the result below.

image

Alternatively, if I enforce a title on PageInfo (through TypeScript) we can be sure that a page will never have a title

src/app/app.component.ts
hudson-newey: I have decided that using the category as a fallback placeholder will be more descriptive an have implemented the appropriate code. ([comment](https://github.com/QutEcoacoustics/workbench-client/pull/2034#discussion_r1139695891))

src/app/app.component.ts
hudson-newey: 100%. I have found where this will cause problems.

In the case where there are multiple mounting points for a page (e.g. Sites / Points), multiple `PageInfo` objects are returned.

I'll be fixing this and adding some tests to assert that it works as expected

src/app/app.component.ts
```diff
@@ -86,7 +96,55 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
...
+  private buildHierarchicalTitle(subRoute: MenuRoute): string {
+    // If the page route has an explicit way to construct the title, use the title callback
+    // if there is no `title` callback defined in the menuRoute, use the category label as a fallback
+    const componentTitle = " | " + (
+      subRoute?.title ?
+        subRoute.title(this.routerState) :
+        // since category labels are not title cased (first letter after space capitalized), we need to title case them
+        // since explicit route titles commonly include model names which are case sensitive, explicit titles should not change casing
+        // e.g. Project name "Tasmanian wetlands" != "Tasmanian Wetlands" as the user has explicitly not capitalized "Wetlands"
+        titleCase(subRoute.label)
```
atruskie: two points:

- this is a complex multiline conditional; it will read better if you just break it into a few steps with an if/else
- you should probably catch and handle exceptions (and log the error on the console) thrown by the title function

atruskie (approved):
Great, ready for this to be done.

src/app/app.component.ts
hudson-newey: I have made the requested code changes

image

PR #2041, opened 2023-02-15
Fix library navbar inheriting component route params

### LINKED ISSUE #2028: Strong route query params are retained from current view when clicking on library menu item
**Describe the bug**
All `rotueParams` are retained from the current view to the library page if the user clicks on the "Library" navbar menu item.

**To Reproduce**

1. Click on Projects
2. Go to page 3 / (1,n]
3. Click on a project listed on page 3
4. Click the back button on your web browser
5. Click on "Library" menu item in the top menu bar
6. Observe that you are seeing results from page 3

OR **using url bar navigation**

1. Navigate to [https://ecosounds.org/projects?page=3](https://www.ecosounds.org/projects/1107?page=3)
2. Click on Library
3. Observe that you are seeing results from page 3

OR **Using project that doesn't have multiple pages**
1. Navigate to [https://www.ecosounds.org/projects/1107?page=3](https://www.ecosounds.org/projects/1107?page=3)
2. Click on "Library"
3. Observe that you are seeing results from page 3

Or **Using a project that does have multiple pages**
1. Navigate to [https://ecosounds.org/](ecosounds.org)
2. Click on Projects
3. Click on "Tasmanian Wetlands" (https://www.ecosounds.org/projects/1084)
4. Click on page 2
6. Press the refresh button in your browser
7. Click on Library
8. Observe that you are on page 2 of the library

OR **Using audio recording id**
1. Navigate to [https://www.ecosounds.org/projects/1107/audio_recordings/579698](https://www.ecosounds.org/projects/1107/audio_recordings/579698)
9. Click on "Library"
10. Observe that you are viewing the audio recording `579698` on the library page

**Expected behavior**
When the user clicks on the Library navbar menu item, it should always act as if the user is going to the home page of the library view (the user should always see page 1, without any filters or selected audio recordings)

- OS: Linux & Windows
- Browser Chromium, Firefox, Edge
- Version Chromium: 108.0.5359.128, Firefox: 109.0, Edge: 108/1462.76

**Additional Behavior:**
From looking through the code it seems to be related to [library.module.ts#L11](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/library/library.module.ts#L11)

hudson-newey:
# Fix library navbar inheriting component routeParams from view model

The library navbar menu item inherited the current view models `routeParams`. This caused a wide range of bugs that wouldn't normally be spotted unless the user inspects the URL.

e.g. Navigating to page 3 of the projects page then clicking on "Library" would navigate the user to page 3 of the library page.

## Changes

- Adds a `libraryNavRoute` `StrongRoute` to `library.menu.ts`
- Adds a unit test to assert that navbar menu items do not use component `routeParams` in their routes
- (Minor Change) adds method return type `void` to method `setHeaderLinks` _This change is optional, however, I found this useful while looking for solutions, and other developers may find the same_

See my full justification on why I chose this solution over others [here](https://github.com/QutEcoacoustics/workbench-client/issues/2028#issuecomment-1430744696)

### Assumptions

I have fixed the issue related to the library item by creating a new libraryNav route. This is similar to how the projects nav menu item works.

However, an alternative solution that I have thought of is "sanitizing" / cleaning all baw-primary-menu > baw-header, baw-header-dropdown input strong routes (Maybe using a directive?) so they have no context of the current view. This would be more "correct" if all nav bar items are supposed to act as a navigation to a completely new view / session.

I have chosen not to use this solution as I believe that menu items routes should follow the same format / conventions as other parts of the client, and should not have their own special rules associated, and that menu routes should instead be explicitly defined as not carrying component route information at the route level, not baw-primary-menu level. By explicitly defining no routeParams in the library nav menu item, I have also conserved the possibility of a future nav bar item that can navigate with knowledge of the current route / component.

I will be adding code comments to document this for future developers and will add tests asserting that this change does not get accidentally reverted in future revisions.

## Problems

None

## Issues

Fixes: #2028 

## Visual Changes

None

## Final Checklist

### DIFF
--- a/src/app/components/library/library.menus.ts
+++ b/src/app/components/library/library.menus.ts
@@ -1,6 +1,10 @@
 import { Category, menuRoute } from "@interfaces/menusInterfaces";
 import { StrongRoute } from "@interfaces/strongRoute";
 
+// there is a clear distinction between the nav route and the library route because the nav route should never have knowledge of the current
+// component route params. The library nav route should instead always act as if its creating a new instance of the library view.
+export const libraryNavRoute = StrongRoute.newRoot().add("library");
+
 export const libraryRoute = StrongRoute.newRoot().add("library", (params) =>
   !params
     ? {}
@@ -29,7 +33,7 @@ export const libraryCategory: Category = {
 export const libraryMenuItem = menuRoute({
   icon: libraryCategory.icon,
   label: "Library",
-  route: libraryRoute,
+  route: libraryNavRoute,
   order: 6,
   tooltip: () => "Library of annotations",
 });
--- a/src/app/components/library/library.module.ts
+++ b/src/app/components/library/library.module.ts
@@ -2,13 +2,13 @@ import { NgModule } from "@angular/core";
 import { RouterModule } from "@angular/router";
 import { getRouteConfigForPage } from "@helpers/page/pageRouting";
 import { SharedModule } from "@shared/shared.module";
-import { libraryRoute } from "./library.menus";
+import { libraryNavRoute } from "./library.menus";
 import { AnnotationComponent } from "./pages/details/details.component";
 import { LibraryComponent } from "./pages/list/list.component";
 
 const components = [LibraryComponent, AnnotationComponent];
 
-const routes = libraryRoute.compileRoutes(getRouteConfigForPage);
+const routes = libraryNavRoute.compileRoutes(getRouteConfigForPage);
 
 @NgModule({
   declarations: components,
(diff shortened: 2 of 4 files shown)


atruskie:
Hmm, so a good cue that you're solving the wrong problem is when we have two strong route objects for the same route. We should only have one per unique route.

The real issue seems to be strong route: https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/directives/strongRoute/strong-route.directive.ts

It is designed to incorporate current route and query string parameters into the URL when it templates on of our links.

The real fix here is to decide what the proper behavior of the directive should be. This change may have been introduced in https://github.com/QutEcoacoustics/workbench-client/pull/651

PR #2042, opened 2023-02-16
Change delete model page to a delete confirmation modal

### LINKED ISSUE #496: Delete model should use modal
The method of deleting models is a bit overkill. Instead of redirecting the user to a separate page, look into displaying a modal pop-up instead which handles the same process.

Issue generated from the following comment: https://github.com/QutEcoacoustics/workbench-client/pull/488#pullrequestreview-498515737

>the delete warning should not be a page, but a bootstrap modal shown on the details page
>
>    and shouldn't use a h1 for text
>    and shouldn't be a form

hudson-newey:
# Change delete model page to a delete modal

At the moment, when deleting a project, region or site model, the client navigates to an entirely new page route to confirm. This is overkill and should instead use a modal.

## Changes

- Changes project delete page to a delete modal `modalMenu` item
- Changes region delete page to a delete modal `modalMenu` item
- Changes site / region delete page to a delete modal `modalMenu` item
- Adds tests for all the above

**Minor Changes**
- Adds rounded corners to `LinkMenuItem`'s. The rounded corners were already present on all other menu items, and applied to the `LinkMenuItem` icon, indicating that this was merely a "typo" / missed bootstrap class on the menu link container. (before and after examples attached below)
- links mock regions to projects in `details.component.spec.ts` tests
- links mock sites to region / site in `details.component.spec.ts` tests
- Adds the ability for modal menu items to be disabled

**Moves menu item inheritance to the `MenuItem` interface**  
This interface is used by all menu items, allowing for all menu items to inherit each other now.

Historically, `MenuRoute`'s could only be children of other `MenuRoute`'s. This caused all the different menu items to cluster together into groups. e.g. all `MenuRoute`'s would be clustered and sorted together. This process would then be repeated for the `MenuModal`'s and `MenuAction`'s.

This has now been changed so that `MenuAction`, `MenuModal` and `MenuRoute`'s can now inherit and be children of each other (causing them to sort as a hierarchical cluster).

**Minor changes / bug fixes unrelated to ticket:**
- Under certain conditions on Chrome, there was a bug where a border would not show around the annotation batch uploading file input

## Problems

1. I have noticed that the menu items are not all of the same casing and is inconsistent across the client. We may want to consider using the Angular [`TitleCasePipe`](https://angular.io/api/common/TitleCasePipe) to standardize menu casing
2. The `tag_groups` route has regressed in functionality and there is currently no way to verify that the delete modal is working on the `tag_groups` view. I have verified that the client has not regressed in functionality due to the changes made in this PR
3. The `delete` `Tag` endpoint is not currently functional on the server side e.g. `api.staging.ecosounds.org/admin/tags/2883`

## Issues

Fixes: #496 

### Delete modals

**Delete project modal:**
image

### Batch upload annotations visual changes
The file input box did not have a border around it on Chrome

**From**
image

**To**
image

## Final Checklist


atruskie (changes requested):
Ok; I don't think we need three separate components that are identical in implementation.

Let's have one component and we can customize it for each page.

The trick will be invoking a callback on the component..

hudson-newey (commented):
@atruskie _OUTDATED_ Before you review, there is a major assumption [here](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-884f7f49640e5923f6bcac4c51d90340330a178f662defbe61e5f5aac1c512deR87-R96).
image

That the best practice for updating the page component `PageInfo` can be done through the following steps.

**When the router navigates to a new view component**
1. Get the shared route `IPageInfo` (this `IPageInfo` is a partial implementation of `PageInfo` does not have any knowledge about the state of the view or component. It only has knowledge on type of view component & app level `IpageInfo` e.g. fullscreen, route, etc...)

2. Get the `PageInfo` of the current component. The `PageInfo` object is an instantiated `PageInfo` object with knowledge of the current component. This functionality is set when the `PageComponent` is initialized [source](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/helpers/page/pageComponent.ts#L26-L31).
3. Add the component instance to the `PageComponent`'s `PageInfo` as the value for the `componentInstance` attribute

This approach of adding the component instance to the `PageComponent` `PageInfo` was chosen because I believe that the `PageComponent` `PageInfo` should hold all the information about the constructed component.

~~Another possible approach I have thought of is doing the same steps, but on the `PageComponent`, and defining a new getter on the `PageComponent` class (e.g. `instance()`) that returns the instance.~~

~~This would lead to the following code instead~~
```ts
this.sharedRoute.pageInfo
      .subscribe((pageInfo) => pageInfo.pageRoute.route.pageComponent.setInstance());
```

Apart from these assumptions that are giving me the shivers, the PR is ready for review

src/app/components/regions/pages/details/details.component.ts
```diff
@@ -134,6 +139,17 @@ class DetailsComponent extends PaginationTemplate<Site> implements OnInit {
...
     return this.sites.size > 0;
   }
+
+  public deleteModel(): void {
+    this.regionsApi.destroy(this.region, this.project)
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe({
+        complete: () => {
+          this.notifications.success(defaultSuccessMsg("destroyed", this.region.name));
+          this.router.navigateByUrl(this.project.viewUrl);
```
atruskie: when `hideProjects` is `true` this should probably redirect to the sites list page instead

src/app/components/shared/delete-modal/delete-modal.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  public constructor(
+    private sharedRoute: SharedActivatedRouteService,
+  ) { super(); }
+
+  @Input() public successCallback: (component: PageComponent) => void;
+  public closeModal: (result: any) => void;
+  public dismissModal: (reason: any) => void;
+
+  protected deleteModel(): void {
+    this.successCallback(this.sharedRoute.pageComponentInstance);
```
atruskie: you just need to invoke success callback, shared route should be invoking the menuitem callback with the parameter

delete modals shouldn't know about shared menus.

if they don't know then we could put the delete button in the page itself, or in the menu widget list. either works!

src/app/components/shared/delete-modal/delete-modal.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    <div class="modal-footer">
+      <a class="btn btn-outline-primary" (click)="dismiss()">Cancel</a>
+      <a class="btn btn-danger text-white" (click)="deleteModel()">Delete</a>
+    </div>
+  `,
+})
+export class DeleteModalComponent extends withUnsubscribe() implements ModalComponent {
+  public constructor(
+    private sharedRoute: SharedActivatedRouteService,
```
atruskie: hmmmm code smell

src/app/components/sites/pages/details/details.component.ts
```diff
@@ -89,6 +95,17 @@ class SiteDetailsComponent extends PageComponent implements OnInit {
...
     this.site = models[siteKey] as Site;
   }
+
+  public deleteModel(): void {
+    this.sitesApi.destroy(this.site, this.project)
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe({
+        complete: () => {
+          this.notifications.success(defaultSuccessMsg("destroyed", this.site?.name));
+          this.router.navigateByUrl((this.region ?? this.project).viewUrl);
```
atruskie: again we should cater for hideProjects

src/app/helpers/page/pageInfo.ts
```diff
@@ -56,6 +57,10 @@ export class PageInfo implements IPageInfo {
     this.route.pageComponent = target;
   }
 
+  public setComponentInstance(target: PageComponent): void {
+    this.componentInstance = target;
+  }
+
```
atruskie: this is no longer needed

atruskie (changes requested):
This is looking really good now. 

My only concern is that `pageComponentInstance` could be holding a reference to a page that is no longer active (between the time the component is removed, and then a new page is created and attached to the outlet). 

Suggest: using `RouterOutlet.deactivate` to clear the shared route's `pageComponentInstance`

Also: I'm not sure about this, but perhaps we should bind to `detach` and `attach` instead? Or both? I'm concerned that if we ever take use of `RouteReuseStrategy` this system will break.

src/app/components/shared/delete-modal/delete-modal.component.ts
hudson-newey: As the `delete-modal` no longer uses route information in the `successCallback`, the `sharedRoute` is no longer needed and I have therefore been able to remove the constructor

hudson-newey:
It would be interesting to use the `RouteReuseStrategy` in the future, however, [from what I can see](https://angular.io/api/router/RouterOutlet) the `RouterOutlet` `attach` and `detach` events are only useful when there is a `RouteReuseStrategy` in place because it triggers when a reusable route is attached / detached.

If we end up using `RouteReuseStrategy` in the future is should be as simple as calling the existing `onRouterOutlet` method from the `attach` and `detach` `routerOutelet` events.

src/app/components/shared/delete-modal/delete-modal.component.ts
hudson-newey: I have implemented this through the [menu service](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-ab6a0ee0e398612e0fbfa14ca21cf089c65fb17ad1335d99e1069a8f583616d4R165-R173) that adds the component information to the callback.

And 2 lines of additional code to the [menu components](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-c9315700578bd5e26c449a32c33915231c2c45e8e4246f2767f46491f448eaafR136-R138) `createModalAction` method (this method gives structure & action to modal menu items).

atruskie (approved):
Deal with the attach/detach problem as outlined already, otherwise this is good to go.

src/app/components/shared/delete-modal/delete-modal.component.ts
```diff
@@ -28,16 +28,22 @@ import { ModalComponent } from "@menu/widget.component";
   `,
 })
 export class DeleteModalComponent extends withUnsubscribe() implements ModalComponent {
-  @Input() public successCallback: () => void;
-  public closeModal: (result: any) => void;
-  public dismissModal: (reason: any) => void;
+  @Input() public successCallback?: () => void;
+  @Input() public closeModal: (result: any) => void;
+  @Input() public dismissModal: (reason: any) => void;
```
atruskie, line 36: what's the difference between close and dismiss? they're synonyms to me

hudson-newey: Dismiss and close are inherited from NgbModals

Dismiss is used when the user presses escape to dismiss the modal without action or otherwise clicks away from it.

Close is used to close the modal with action (e.g. the user confirmed). This is typically useful for when there's some sort of input box in the modal that should be returned when submitted. But for our implementation, it's used to return if the user presses the confirm / delete action button.

Now that I think about it, we should probably only use close, and let Ngb handle dismiss.

atruskie: Fair enough. If that's ngb conventions document the inputs then to relay to the reader what they mean

hudson-newey:
Since your last review, I have added the delete modals to several more pages. I have outlined the changes since your last review below:

### Tag Groups

[Added a tag group delete modal menu item](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-d02010a2edbbfce38d0c2d354a3567be1265eaf47f46cd258617bba301c904e9R7-R15). This menu item is used on the edit tag groups page.

[Added a method to delete tag groups from the list page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-f6db513b4fdcad22c1a109c1fc7cf35f587f813931231539859cbf30af6170a0R52-R63). This method allows for a singular tag-group to be deleted through a modal from the list page if the "delete" button next to a tag is clicked. Since this button is embedded into the page, this async method uses the result value of the modal to assert if the user confirmed the deletion.

[Added a method to delete tag groups from the edit page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-429f2592dc51458f6b131c954a4282b20e27c242325839ed279e7b24bdd6e0e9R69-R79). This method is called when the "Delete tag" modal menu item is clicked and confirmed.

[Added tests to assert that tag groups are correctly deleted from the list page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-7053220bd2baafbd8ffd9deb741d3e17b85b347abfe20a2fb1ec580bd30130adR61-R94)

[Added tests to assert that tag groups are correctly deleted from the edit page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-9d6b2b9f46477243f8830059b7a979061a8a861e881354575dc7bfd83944468dR116-R136)

### Singular Tags
[Added tag delete method on the tag edit page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-ff1253ebaaeff9e258999fb5b94521619222df4f22b5628cdf314baaae5eb9e2R77-R87). This method is called by the tag delete modal menu item.

  [Added tests for the tag delete methods found on the tag edit page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-9d6b2b9f46477243f8830059b7a979061a8a861e881354575dc7bfd83944468dR116-R136)

[Added a tag deletion method to the tag list page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-8ebb0989a7940a29cbd15de68373051c669f7a56681e1bf1080b1c3ebbad43f4R58-R69). This method allows for a singular tag to be deleted through a modal if the "delete" button next to a tag is clicked. Since this button is embedded into the page, this async method uses the `result` value of the modal to assert if the user confirmed the deletion.

[Added test for the tags list page](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-f6ede45b3abcf07b2aa8684efddd5cbc8b9868a002e16824508128e9e10a17e1R58-R91). These tests assert that a singular tag can be deleted if the "delete" button next to a tag is clicked.

[Added a tag delete modal menu item](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-a85c5c199792ec2b7441978d704da8cf5f1d6f2ab7779e6b67e8b886dcc6de97R7-R15). This menu item is used on the tag edit page.

### Account cancellation
[Added a `cancelAccount` method to the `my-profile` page component](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-f8d65ddc50a44e5a5d2c6ee17451f71ff8755b71350978b71812e1d7c11e005fR153-R163). This method is called when the "Cancel my account" menu button is clicked and the delete modal is confirmed.

[Added tests to assert that the cancel account method works as expected](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-f946588827efc2c863a964f3db46c2f5c3ae8a34ea31f628f1f6e521ea04fb2cR155-R179)

[Added a "Cancel my account" modal menu item](https://github.com/QutEcoacoustics/workbench-client/pull/2042/files#diff-b3bc37cd276de0214025b2cb8743ac9049dcd4e489c0b9f943e241144aee2d63R7-R16). This menu item is shown on the `my-profile` page component.

atruskie (approved):

PR #2062, opened 2023-05-25
Fix: project shown in titles when projects hidden

### LINKED ISSUE #2052: Route titles should not show project if projects are hidden
**Describe the bug**
If projects are hidden, the project name is shown in the route title

**To Reproduce**

1. Go to [data.acousticobservatory.org/projects/1/regions/54](https://data.acousticobservatory.org/projects/1/regions/54)
4. Observe that the route title includes the name of _project 1_ (the hidden project)

**Expected behavior**
The project name should not be emitted in the route title if projects are hidden

**Screenshots**
image

- OS: Windows 10
- Browser Firefox
- Version 102.9.0esr

**Additional context**
This bug is only visual and does not produce any errors

hudson-newey:
# Bug Fixs: project shown in titles when projects hidden

Since regions still technically have a parent project item, when projects were hidden, the project model name was still emitted in the route title.

This is incorrect as users should not see any project information if projects are hidden

## Changes

* Adds condition in route title builder to use a shallow region route instead of project route if projects are hidden
* Adds test to assert that project names are not emitted when projects are hidden

## Problems

Adds a _TODO_ stating that we may want to refactor the route title builder in the future to use the menu service's `rootToMenuRoute()` method. This method is not currently exposed as public, and causes circular dependencies in the current architecture.

## Issues

Fixes: #2052 

## Visual Changes

From:
image

To:
image

## Final Checklist

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -243,7 +243,7 @@ describe("AppComponent", () => {
     });
 
     it("should have an injected page title strategy", () => {
-      expect(titleStrategyInjectable).toBeTruthy();
+      expect(titleStrategyInjectable).toBeInstanceOf(PageTitleStrategy);
     });
 
     // a route without a hierarchy is a route with no `parent` property
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -18,11 +18,12 @@ import {
 } from "@baw-api/resolver-common";
 import { titleCase } from "@helpers/case-converter/case-converter";
 import { PageComponent } from "@helpers/page/pageComponent";
-import { MenuRoute } from "@interfaces/menusInterfaces";
+import { MenuRoute, TitleOptionsHash } from "@interfaces/menusInterfaces";
 import { GlobalsService } from "@services/globals/globals.service";
 import { MenuService } from "@services/menu/menu.service";
 import { SharedActivatedRouteService } from "@services/shared-activated-route/shared-activated-route.service";
 import { filter, Observable, takeUntil } from "rxjs";
+import { isInstantiated } from "@helpers/isInstantiated/isInstantiated";
 import { IS_SERVER_PLATFORM } from "./app.helper";
 import { withUnsubscribe } from "./helpers/unsubscribe/unsubscribe";
 import { ConfigService } from "./services/config/config.service";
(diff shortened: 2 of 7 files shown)


atruskie (changes requested):
Isn't there a better way to achieve this?

The title is used by the menuroute attached to the page... shouldn't all the project hidden routes/menus be in their own hierarchy which doesn't include project as a parent?

I checked and it doesn't look this is how we've setup our menu routes.

So:

- we could make sure these sub-pages use a menuroute hierarchy that is distinct (which naturally avoids the problem but could explode menu route comlexity)
- or we could put the decision into whether or not to show a title in the hands of the title maker. I.e. the project menu route title function should do something like ` hideProjects ? null : 'normal title'`. We'd probably inject an options hash into the title lambda then.

Either way the logic here should belong outside the app.component, perhaps except for injecting the required options.

hudson-newey:
> or we could put the decision into whether or not to show a title in the hands of the title maker. I.e. the project menu route title function should do something like ` hideProjects ? null : 'normal title'`. We'd probably inject an options hash into the title lambda then.

I've implemented this, however, I've realized that it generates route titles that are not consistent with the rest of the client.

E.g.

image
_Project route title_

image
_Site route title when projects are hidden_

I expect the result to be "<\<brandName\>> | Sites | aaa" not "<\<brandName\>> | aaa"

This arises because sites should actually be mounted under the `shallowRegion` route when projects are hidden (as is with the breadcumbs and secondary menu).

Therefore, we should probably go with the second proposed option

> we could make sure these sub-pages use a menuroute hierarchy that is distinct (which naturally avoids the problem but could explode menu route comlexity)

This will explode the scope of the project, but if done correctly might actually decrease the complexity, because there won't be two routes active (as is with the current client)

atruskie (approved):
Looks great!

hudson-newey:
Will be fixing test dependency error before merging

hudson-newey:
I've updated the route title tests to [test each route title](https://github.com/QutEcoacoustics/workbench-client/pull/2062/files#diff-a7029529e750e168b2942a952c12426370cf9b2c8588419250a2b5930b09ef19R70-R78) under a [pre-determined array of title option states](https://github.com/QutEcoacoustics/workbench-client/pull/2062/files#diff-a7029529e750e168b2942a952c12426370cf9b2c8588419250a2b5930b09ef19R25-R28)

This change was chosen as it allowed to test route titles under different conditions while not breaking or needing to refactor ever route fragments unit tests.

I've also updated to _master_, and rebased into one commit

PR #2065, opened 2023-06-14
[DRAFT] [Testing needed] Update Angular and its dependencies

hudson-newey:
# Updated Angular and its dependencies

Update Angular and its dependencies

[Angular 15 Breaking & Deprecations](https://angular.io/guide/update-to-version-15#breaking-changes-in-angular-v15)
[Angular 16 Breaking & Deprecations](https://angular.io/guide/update-to-version-16#highlighted-breaking-changes-in-angular-v16)

## Changes

* The use of `~` in SCSS imports (which used to represent `node_modules`) is now deprecated as has been removed. To ensure all imports still work as expected the `node_modules` path has been added to the `stylePreprocessorOptions` paths
* Updated version of EsLint showed errors that `Immutable.Collection<unknown, unknown>` could hypothetically never end in object depth. all `Immutable.Collection<unknown, unknown>` are now strongly typed using the correct `FromJS` type
* `RouterLinkWithHref` is now deprecated as all its functionality is now implemented by `RouterLink` and is now an "alias type" (incorrect word). I've changed all `RouterLinkWithHref` types to `RouterLink`
* Faker.js `address` module is now deprecated in favour of `location`. I have updated the code base accordingly
	- datatype.number() is now deprecated in favour of `number.int()` or `number.float()` I have adjusted the code base accordingly
	- Faker.js imageUrl is now replaced with `image.url()`
	- Faker.js `random.alphanumeric()` replaced with `string.alphanumeric()`
	- Faker.js `faker.name` replaced with `faker.person`
	- Faker.js `faker.random.word()` replaced with `faker.lorem.word()`
* TypeScript can now trivially infer the types of function parameters from their default values. If the parameter type cannot be overloaded with multiple types, TypeScript now throws an error. I have therefore removed trivially inferred parameter types
* Removed defaultProject workspace extension as per [this issue](https://github.com/angular/angular-cli/issues/11111)
* Since we are now using esbuild, `namedChunks` and `vendorChunk` are no longer needed in the angular config. Therefore I have omitted them to reduce console pollution

## Problems

* Must use NodeJS version 16 or 18 from this version forward
* Removes support for KaiOS and Opera Mini mobile browsers (this is due to the higher TS target)

## Visual Changes

None

## Final Checklist


atruskie (changes requested):
We should make use of the new [esbuild](https://angular.io/guide/update-to-version-16#faster-builds-with-the-esbuild-developer-preview) option.

Why are `id` and `recaptchaTokens` removed from a bunch of the models?

hudson-newey:
> Why are `id` and `recaptchaTokens` removed from a bunch of the models?

the `id` attribute is implemented by the `AbstractModel` class (which most models extend), but don't need to override since they don't add additional functionality (an exception is the [permissions](https://github.com/QutEcoacoustics/workbench-client/pull/2065/files#diff-e473233fdabe191c77a8ccde569a0252a1cea428cd8b31849e853417a0a449e4R28) model).

The above was causing the new version of EsLint to throw errors.

Additionally, `recaptchaToken` is implemented by the `FormComponent` which some page components extended. I have a suspicion that `recaptchaToken` might be needed on some `FormComponent`'s so I'll have to do some testing to see if there are any regressions.

Alternatively, as it doesn't actually affect the performance in any way, I'm happy to disable this EsLint rule in the name of being verbose

hudson-newey:
Superseded by #2085

PR #2074, opened 2023-09-08
Site/point deletion: warn users what will happen with audio recordings upon deletion

### LINKED ISSUE #2067: site/point deletion: warn users what will happen with audio recordings

It is currently unclear what happens to audio recordings if a site/point is deleted. 

This should be explained in the user interface. Something like:

> when this point/site is deleted it will be made invisible. For data safety: all audio recordings will no longer be accessible but will be recoverable. If you need to recover these recordings after they have been deleted, please \<contact us\>.

True deletion? not supported server side currently.

**Additional context**

Now that users can upload recordings themselves, they're using exposed options more frequently. A user deleted a point and then tried to reupload the same recordings, which of course failed because the recordings still exist and duplicate detection kicks in.

hudson-newey:
# Site/point deletion: warn users what will happen with audio recordings upon deletion

As users are now uploading audio recordings via the Harvest webpages, they sometimes run into issues when uploading duplicate audio recordings.

Currently when a point/site is deleted, the audio recordings are disassociated from the site/point, but they still exist (they are hidden). So if a user uploads an audio recording -> deletes a site -> creates a new site -> uploads the same audio recording, it will be blocked from Harvest as a duplicate upload.

We should explain this behavior to users in "user friendly" language when the site is deleted via a warning in the deletion modal window.

## Changes

* Adds a specialised delete modal for sites/points
* Adds the ability to pass component options to modals
* Adds tests for new sites/points delete modal

## Problems

None

## Issues

Fixes: #2067

## Visual Changes

image
**New deletion modal with warning text**

## Final Checklist

### DIFF
--- a/src/app/components/shared/menu/widgets/widget.component.ts
+++ b/src/app/components/shared/menu/widgets/widget.component.ts
@@ -7,4 +7,5 @@ export interface ModalComponent extends WidgetComponent {
   closeModal: (result: any) => void;
   dismissModal?: (reason: any) => void;
   successCallback?: () => void;
+  options?: Record<string, unknown>;
 }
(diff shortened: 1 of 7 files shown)


atruskie (approved):
:shipit:

Issue #2060, opened 2023-05-23
Add creator/updater cards to harvest pages


atruskie:
**Is your feature request related to a problem? Please describe.**

It's hard to tell who created an upload/harvest.

**Describe the solution you'd like**

- [ ] Add out standard created/updated by cards in the harvest details page
- [ ] Add creator column to the table on the harvest list page
- [ ] Remove the created/updated cards on the harvest list page from the secondary action menu: these cards reflect the information from the project which is confusing in this context.

PR #2075, opened 2023-09-12
Added creator/updated cards to Harvest pages

hudson-newey:
# Add creator/updater cards to harvest pages

**I found a bug in my implementation after I opened this PR. Not ready for review**

The current created by / updated by cards refer to the project. These should instead refer to the harvest.

## Changes

* Adds the "Created by / Updated by" menu cards to the harvest details page
* Adds a "creator" column to Harvest list pages
* Removes the "Created by / Updated by" menu cards from the harvest list page
* Fixes typo in Harvest batch uploading page
* `permissions-shield.component.ts` (menu bar "Created by / Updated by" card now displays Harvest information if available)

## Problems

None

## Issues

#2060 

## Visual Changes

image
_Harvest list page with "Creator" column and no "Created by / Updated by" menu cards_

image
_Harvest details page with "Created by / Updated by" menu card and fixed typo_

## Final Checklist

### DIFF
--- a/src/app/components/harvest/pages/list/list.component.ts
+++ b/src/app/components/harvest/pages/list/list.component.ts
@@ -10,7 +10,6 @@ import {
   newHarvestMenuItem,
 } from "@components/harvest/harvest.menus";
 import { PageComponent } from "@helpers/page/pageComponent";
-import { permissionsWidgetMenuItem } from "@menu/widget.menus";
 import { Harvest } from "@models/Harvest";
 import { Project } from "@models/Project";
 import { List } from "immutable";
@@ -101,7 +100,6 @@ ListComponent.linkToRoute({
   category: harvestsCategory,
   menus: {
     actions: List(harvestsMenuItemActions),
-    actionWidgets: List([permissionsWidgetMenuItem]),
   },
   pageRoute: harvestsMenuItem,
   resolvers: {
--- a/src/app/components/harvest/screens/uploading/batch-uploading.component.html
+++ b/src/app/components/harvest/screens/uploading/batch-uploading.component.html
@@ -58,11 +58,11 @@
           Cyberduck and connect to our cloud.
         </li>
         <li>
-          On your first connection to our cloud, Cyberduck may prompt you with 
-          an "unknown fingerprint" message. Select "Allow".  
+          On your first connection to our cloud, Cyberduck may prompt you with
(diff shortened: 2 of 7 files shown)


src/app/components/harvest/pages/list/list.component.ts
```diff
@@ -92,6 +92,10 @@ class ListComponent extends PageComponent implements OnInit {
     return model;
   }
 
+  public harvestCreator(model: Harvest): User {
+    return model.creator;
+  }
+
```
atruskie: does angular change detection still penalize methods?

Regardless seems like a little too much ceremony to have a whole function just to access one property

src/app/components/harvest/screens/uploading/batch-uploading.component.html
atruskie, line 1: ensure the whitespace stripping done by the ide does not affect the formatting of the commands in the page.

atruskie (approved):

src/app/components/harvest/screens/uploading/batch-uploading.component.html
hudson-newey: I got worried by this too and checked

The changes did not impact the visuals

src/app/components/harvest/pages/list/list.component.ts
hudson-newey: I agree. I initially tried to use the `model.creator` directly in the template, but it didn't seem to work because change detection wouldn't fire when the associated model would change from an `UnresolvedModel` to a resolved model.

Using a wrapper function like this made change detection work as expected and updated the templated model when the associated model changed.

If I can't use `model.creator` directly in the template, what would you recommend for the best path forward? Fix the underlying issue in the association decorator to make change detection fire when we change from an `UnresolvedModel` -> Resolve model?

atruskie: Yes, we should investigate the underlying cause. The point of the system is that this happens automatically.

File a new issue. Start cataloguing places to update (e.g. this spot) once we work how to fix it. I wonder if we could use one of these new fangled signals to trigger the change.

hudson-newey:
Blocked by #2076

hudson-newey:
With the ngx-datatable props fix we found on Tuesday, this PR is ready to merge. I will be merging shortly

PR #2085, opened 2023-10-17
Upgrade Angular and Its Dependencies

### LINKED ISSUE #2077: Change builder to esbuild
We are currently using the angular builder in our `angular.json`

To improve the performance of the builder, we should switch to [esbuild](https://angular.io/guide/esbuild) and ensure that there are no regressions through testing.

```
"architect": {
  "build": {
    "builder": "@angular-devkit/build-angular:browser"
    }
}
```

hudson-newey:
# Upgrade Angular and Its Dependencies

We have several outdated dependencies, this pull request updates all the dependencies

You can see all the breaking changes due to our migration from ngx 14 -> 16 [here](https://update.angular.io/?v=14.0-16.0)

Although, a lot of tests were modified due to breaking changes in ng-mocks (see below)

## Changes

- Added tests for time.component.ts
- Fixed bug in time.component.ts where errors weren't shown if a propper time wasn't input first
- Commented out empty TODO tests (empty tests now causes a Karma error)
- When fetching an element mocked with ng-mock, we can no longer fetch it with it's mocked instance, you now have to fetch it using the component which was mocked. This caused tests to break after upgrading
- ng-mock MockRender now requires an empty object as the second parameter
- Ran formatter on all files (this was automatically done by the `ng update` command)
- We now use the "declare" keyword in certain places when overriding abstract class methods
- Importing ngx-datatable styles is now done through the `@use` keyword instead of `@import`
- Using the ~ as an alias for node_modules is no longer supported for scss imports. node_modules was added as a default import location for scss imports, and the use of the "~" was removed
- Changed builder to es-build 🥳
- Angular testing now recognises .spec.ts as the file extension for test files. This assertion was therefore removed from our test.ts
- Stronger type checking for ng-template's required me to update the Typeahead input template output context to include the search term & term formatter
- The collapsed property is no longer exposed in ngx-boostrap DOM elements. This required tests to be changed to query classess not DOM object properties
- Added "work in progress" menu item predicate to add better semantic meaning as to why only admins can view a menu item
- Officially dropped support for kaios, and opera mini browsers (as they do not support ES6 and are mobile browsers with small market share)
- RouterLinkWithHerf is no longer implemented in Angular and all functionality has been transfered to RouterLink. Therefore, I've changed all instances of RouterLinkWithHerf with RouterLink (ensuring that I don't break strong route)
- In [cards.component.spec.ts](src\app\components\shared\model-cards\cards\cards.component.spec.ts), the CardsComponent was being mocked twice each test, causing errors
- TypeScript return type of NodeJS setInterval was renamed to NodeJS.Timeout
- "plugin:@angular-eslint/ng-cli-compat" es-lint extensions are now deprecated (has been merged with @angular-eslint/recommended)
- eslint plugin "plugin:@angular-eslint/ng-cli-compat--formatting-add-on" was deprecated and merged with @angular-eslint/recommended
- Upgraded to NodeJS version 18 (LTS) (in docker container and dev environment)

## Problems

None

## Issues

Fixes: #2077 

## Visual Changes

None

## Final Checklist


src/app/app.menus.ts
```diff
@@ -47,3 +47,11 @@ export const isProjectEditorPredicate = (
...
 export const isAdminPredicate = (user: User): boolean =>
   isLoggedInPredicate(user) && user.isAdmin;
+
+/**
+ * Returns true only if the user is an admin and can see work in progress items
+ * This predicate differs from the `isAdminPredicate` for semantic reasons
+ * 
+ * @param user User session data. This will be used to check if the user is an admin
+ */
+export const isWorkInProgressPredicate = isAdminPredicate;
```
atruskie: well if it is meant to mirror the WIP component, then it should be something like `env.isDev || isAdminPredicate()`

https://github.com/QutEcoacoustics/workbench-client/blob/152d6dba544751639e61dce4c056e9743cbad958/src/app/components/shared/wip/wip.component.ts#L63-L66

src/app/components/projects/projects.menus.ts
```diff
@@ -111,8 +112,8 @@ export const uploadAnnotationsProjectMenuItem = menuRoute({
   icon: ["fas", "file-import"],
   label: "Batch Upload Annotations",
   parent: projectMenuItem,
-  // TODO Change to isProjectEditorPredicate
-  predicate: isAdminPredicate,
+  // TODO: Once functionality is implemented, this should be changed to isProjectEditorPredicate
```
atruskie: should it?

you should rely on a capability from the server if it exists...

src/app/directives/url/url-active.directive.ts
```diff
@@ -35,9 +35,8 @@ export class UrlActiveDirective extends RouterLinkActive {
     _element: ElementRef,
     _renderer: Renderer2,
     _cdr: ChangeDetectorRef,
-    @Optional() _link?: RouterLink,
     @Optional() _linkWithHref?: UrlDirective
   ) {
-    super(_router, _element, _renderer, _cdr, _link, _linkWithHref);
+    super(_router, _element, _renderer, _cdr, _linkWithHref);
```
atruskie: Inconsistent parameter naming compared to https://github.com/QutEcoacoustics/workbench-client/pull/2085/files#diff-6c89349913cac853641bb63226dabd1f09947f41a4711a4306a400d5e2f9f686L47

src/app/guards/input/input.guard.ts
```diff
@@ -12,7 +12,7 @@ export interface UnsavedInputCheckingComponent {
  */
 @Injectable()
 export class UnsavedInputGuard
-  implements CanDeactivate<UnsavedInputCheckingComponent>
+
```
atruskie: I can't tell from the code, but does this canDeactivate guard still work? Worth checking?

atruskie (approved):

src/app/components/projects/projects.menus.ts
hudson-newey: From my understanding, this is how the `isProjectEditorPredicate` predicate works

image

Where `canEdit` comes from the server capabilities

image

Note: Because this PR is going to be merged before the annotation uploads, I've put this behind the `wipPredicate`

src/app/guards/input/input.guard.ts
hudson-newey: This same change was made to the `form.guard.ts`, and I can confirm that it still works as expected. I've also verified that the tests are working as expected for the `form.guard.ts` (they would have failed if it no longer worked).

Shockingly, the `input.guard.ts` wasn't even working before the upgrade. The only place this guard is used is on the Harvest Metadata review page, where it'll show a warning if the user tries to navigate away before a mapping is complete.

image

Placing some debugger breakpoints & `console.log`'s showed that the check was never triggered, even before the upgrade

src/app/components/projects/projects.menus.ts
atruskie: My point rather is that assuming someone can edit a project means that they can upload annotations is not correct.

From my memory of the servers permissions: only owners can edit projects but anyone with writer can upload annotations.

hudson-newey: > My point rather is that assuming someone can edit a project means that they can upload annotations is not correct.
> 
> From my memory of the servers permissions: only owners can edit projects but anyone with writer can upload annotations.

Understandable, I've backfilled a new predicate in preparation

image

hudson-newey:
I've done some testing using staging data and there are no regressions

Given that this PR will impact almost every function of the web client, are you all good for me to merge after CI passes?

PR #2091, opened 2023-11-14
Add Annotations Upload

### LINKED ISSUE #1887: Annotation upload user interface

Users want to be able to upload annotations which they've generated (either manually or by machine) for purposes of sharing or validation.

Expose a UI for managing this. 

The server and client will need to negotiate on responsibilities.

We don't currently store enough metadata to bulk import annotations in our current audio events tables without polluting the data. 

**Additional context**

User story, important

Blocked by: https://github.com/QutEcoacoustics/baw-server/issues/573

hudson-newey:
# Add Annotations Upload

Creates pages for users to create annotation imports and upload annotations

## Changes

- Add annotation import list, edit, delete, and detail pages
- Add tests for created pages
- Add models for audio event imports
- Add models for audio event import files (read and write)
- Add service for audio event imports
- Add support for sub-models through the `@bawSubModelCollection` and `@bawSubModel` decorators
- Add readonly [AudioEventError](https://github.com/QutEcoacoustics/workbench-client/pull/2091/files#diff-5a598ef5eb0c475aa08e663e234d1bb94ed176f3b7d49b695e2507391a19bda9R26-R29) submodel to `AudioEvent` model
- Add support for serialization of FormData arrays and `FileList`'s
- Add options object to `DateTime` pipe so that they are more verbosely expressed in templates e.g. `value | dateTime: true : false` is now `value | dateTime: { includeTime: true, localTime: false }`
- Migrated event summary report model's use of `Id[] | Ids` to the `CollectionIds` type alias
- Fixed bug in `isWorkInProgressPredicate` where it was only shown in production (now only shows in development)
- Fixed some TS typing issues inside the report generation component
- Re-enabled eslint maximum line length (it was incorrectly disabled during the angular upgrade)
- **Important**: It was previously assumed that `FormData` would always be sent as part of a multipart request. However, it is now possible to send `FormData` requests without an associated PATCH/json request. I've changed the [`baw-api.service.ts`](https://github.com/QutEcoacoustics/workbench-client/pull/2091/files#diff-2e5a0799ecfae901aaf8f95aa03ccce27de670e1faf96896372e2992493d13a3) so that it can now handle `FormData` that is not a part of a multipart request

## Problems

Future planned changes to the API require models and services to change.
See [baw-server#664](https://github.com/QutEcoacoustics/baw-server/issues/664) for more information

## Issues

Fixes: #1887 

## Visual Changes

image
**New Menu Item for "Batch Import Annotations" in secondary menu**

image
**Annotation Import List Page**

image
**New Annotation Import List Page**

image
**Edit Annotation Import Page**

image
**Delete Annotation Import Modal (from list page)**

image
**Annotation Import Details Page**

image
**Importing several files (some with errors)**

## Final Checklist


src/app/app.menus.ts
```diff
@@ -70,4 +70,4 @@ export const isAdminPredicate = (user: User): boolean =>
  * @param user User session data. This will be used to check if the user is an admin
  */
 export const isWorkInProgressPredicate = (user: User): boolean =>
-  environment.production && isAdminPredicate(user);
+  !environment.production && isAdminPredicate(user);
```
atruskie:
```suggestion
  !environment.production || isAdminPredicate(user);
```

The predicate is meant to show WIP items if we're in not in production, or if we are, then only admin can see them (so we can test them out on prod but not release to the general public)

src/app/components/upload-annotations/details/details.component.html
```diff
@@ -0,0 +1,239 @@
...
+      </ng-container>
+    </div>
+  </div>
+
+  <span
+    [ngbTooltip]="
+      isBufferValid()
+        ? 'Upload all new annotations'
+        : 'Please fill out all required fields or fix errors'
+    "
```
atruskie: multiline attributes always seem icky to me. code smell?

also what is a buffer in this context?

src/app/components/upload-annotations/details/details.component.html
```diff
@@ -0,0 +1,239 @@
...
+            ></baw-inline-list>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </section>
+</div>
+
+<section>
+  <h2>Upload new annotations</h2>
```
atruskie:
```suggestion
  <h2>Add more annotations</h2>
```

src/app/components/upload-annotations/details/details.component.html
```diff
@@ -0,0 +1,239 @@
...
+      Upload All
+    </button>
+  </span>
+</section>
+
+<ng-template #uploadGroupTemplate let-model="model" let-index="index">
+  <section
+    [ngClass]="
+      model?.errors?.length > 0 ? 'border-danger' : 'border-secondary-light'
+    "
```
atruskie: use semantic classes

```suggestion
    [ngClass]="{'form-invalid': model?.errors?.length}"
```

src/app/components/upload-annotations/details/details.component.html
```diff
@@ -0,0 +1,239 @@
...
+        <ul>
+          <li *ngFor="let error of model.errors">
+            {{ error }}
+          </li>
+        </ul>
+      </div>
+    </div>
+
+    <label for="annotationsCsv" class="form-label required">
+      Annotations CSV File(s)
```
atruskie: inaccurate: could be JSON, could be a raven file

src/app/components/upload-annotations/details/details.component.html
```diff
@@ -0,0 +1,239 @@
...
+</section>
+
+<ng-template #uploadGroupTemplate let-model="model" let-index="index">
+  <section
+    [ngClass]="
+      model?.errors?.length > 0 ? 'border-danger' : 'border-secondary-light'
+    "
+    class="upload-group rounded border p-4 my-3"
+  >
+    <h3>Upload Group {{ index + 1 }}</h3>
```
atruskie: 'group' terminology is weird - why not just keep referring to these as files. You'd maybe have an argument if we could upload multiple files at once?

atruskie (changes requested):
Generally: I think our date time pipe is getting too complex and subsuming the role of the component we specked out in #1736. We ended up not accepting #1736 but we should replace it with a more modern solution. Let's add this to our backlog.

We've also got inconsistent terminology around this feature: `upload` vs `import`. Let's not overload `upload` (since we already upload recordings) and settle on `import`. Update all terminology.

Can we upload multiple files at a time? Or does each file have to be uploaded in it's own 'group'? We really need to be able to upload multiple at once.

OK I didn't get very far in this review before I had to do something else. But there are already significant changes that are needed.

src/app/components/upload-annotations/details/details.component.html
hudson-newey: Each "upload group" can have multiple files, but it's an easy way to assign one (or many) associated tag to many files.

In short: each upload group has different associated tags

src/app/components/upload-annotations/details/details.component.html
hudson-newey: A buffer is all the "upload groups" pending for upload. Each of which unique files (multiple), errors, associated tags, and additional tags.

([view model](https://github.com/QutEcoacoustics/workbench-client/blob/annotations-upload/src/app/components/upload-annotations/details/details.component.ts#L45-L58))

src/app/components/import-annotations/details/details.component.html
```diff
@@ -0,0 +1,243 @@
...
+              else audioRecordingTemplate
+            "
+            size="sm"
+          ></baw-loading>
+
+          <ng-template #audioRecordingTemplate>
+            <a [bawUrl]="value.audioRecording.viewUrl">
+              {{ value.audioRecording.id }}
+            </a>
+          </ng-template>
```
atruskie: it occurs to me that we could have loading component that is so much smarter than our current one:

```suggestion
          <baw-loading value="value.audioRecording">
                <a [bawUrl]="value.audioRecording.viewUrl">
                      {{ value.audioRecording.id }}
                </a>
          </baw-loading>
```

src/app/components/import-annotations/details/details.component.scss
```diff
@@ -0,0 +1,8 @@
+section {
+  margin-top: 2rem;
+}
+
+.required::after {
+  content: "*";
+  color: var(--baw-danger);
+}
```
atruskie: code smell: custom form styling for just one component?

You're honestly telling me we haven't set a class for a form input with an error before?

src/app/components/import-annotations/details/details.component.html
```diff
@@ -0,0 +1,243 @@
...
+        </ul>
+      </div>
+    </div>
+
+    <label for="annotationsCsv" class="form-label required">
+      Annotations File(s)
+    </label>
+    <input
+      type="file"
+      accept=".csv, .selections.txt"
```
atruskie: no JSON format?

src/app/components/import-annotations/details/details.component.html
```diff
@@ -0,0 +1,243 @@
...
+      </ng-container>
+    </div>
+  </div>
+
+  <span
+    [ngbTooltip]="
+      isBufferValid()
+        ? 'Import all new annotations'
+        : 'Please fill out all required fields or fix errors'
+    "
```
atruskie: eww

src/app/components/import-annotations/details/details.component.html
```diff
@@ -0,0 +1,243 @@
...
+      class="form-control border"
+      [files]="model?.files"
+      (change)="pushToBuffer(model, $event)"
+      multiple
+    />
+
+    <div class="mt-4">
+      <baw-typeahead-input
+        label="Associated Tags"
+        [inputPlaceholder]="'Required'"
```
atruskie: the placeholder shouldn't be used to indicate required

it should say what the input does, e.g. `search for tags by name`

but also additional tags aren't required???

src/app/components/import-annotations/details/details.component.html
```diff
@@ -0,0 +1,243 @@
...
+        <ul>
+          <li *ngFor="let error of model.errors">
+            {{ error }}
+          </li>
+        </ul>
+      </div>
+    </div>
+
+    <label for="annotationsCsv" class="form-label required">
+      Annotations File(s)
```
atruskie: if I am understanding correctly then the semantic danger colour is being used to

a) incorrectly indicate a field is required (elsewhere we use an asterisk to indicate required fields) 
and b) incorrectly indicated an error state on the form's first display?

src/app/components/import-annotations/details/details.component.html
```diff
@@ -0,0 +1,243 @@
...
+      <div class="alert alert-danger">
+        <ul>
+          <li *ngFor="let error of model.errors">
+            {{ error }}
+          </li>
+        </ul>
+      </div>
+    </div>
+
+    <label for="annotationsCsv" class="form-label required">
```
atruskie: a) inappropriate name (not just csv)
b) no input with that name exists?

src/app/components/import-annotations/details/details.component.ts
```diff
@@ -0,0 +1,333 @@
...
+const audioEventImportKey = "audioEventImport";
+
+interface ImportGroup {
+  /** The list of files to be imported */
+  files: FileList;
+  /** The list of errors encountered during a dry run. */
+  errors: string[];
+  /**
+   * The list of additional tag IDs that are not found within the imported file to be associated with the imported files
+   * This is separate from the identified events because the api doesn't accept additional tags that are already within the file
```
atruskie: incorrect: https://github.com/QutEcoacoustics/baw-server/blob/349ff8d284eb7624d94a035a94bcdaca55525889/app/modules/api/audio_event_parser/tag_transformer.rb#L30-L32

It simply allows a user to tag everything in an import with an additional tag. e.g. "testing" or "training"

src/app/components/import-annotations/details/details.component.ts
```diff
@@ -0,0 +1,333 @@
...
+    private tagsApi: TagsService,
+    private eventsApi: ShallowAudioEventsService,
+    private eventImportsApi: AudioEventImportService,
+    private notifications: ToastrService,
+    private router: Router
+  ) {
+    super();
+  }
+
+  public importBuffer: ImportGroup[] = [this.emptyImportGroup];
```
atruskie: buffer has a fairly technical meaning. I'm not sure we want to use it here, Why not `importGroups`

src/app/components/import-annotations/details/details.component.ts
```diff
@@ -0,0 +1,333 @@
...
+
+    // if the user updates an existing import group, we don't want to create a new one
+    // however, if the user uses the last empty import group, we want to create a new empty one
+    // that they can use to create a new import group
+    if (this.isImportBufferFull()) {
+      this.importBuffer.push(this.emptyImportGroup);
+    }
+  }
+
+  // uses a pointer to the ImportGroup heap object and updates the additional tag ids property
```
atruskie:
```suggestion
  // uses a reference to the ImportGroup object and update the additional tag ids property
```
calm down hackerman

src/app/components/import-annotations/details/details.component.ts
```diff
@@ -0,0 +1,333 @@
...
+    // we should therefore filter out any invalid models before importing
+    // however, this should be caught by form validation, preventing a user to import if there are errors
+    const validImportBuffers = this.importBuffer.filter(
+      (model) => model.errors.length === 0 && model.files?.length > 0
+    );
+
+    validImportBuffers.forEach((model: ImportGroup, i: number) =>
+      this.importEventGroup(model, i)
+    );
+  }
```
atruskie: we probably just shouldn't send anything is there are invalid models? not sure

src/app/components/import-annotations/details/details.component.ts
```diff
@@ -0,0 +1,333 @@
...
+            this.removeFromBuffer(i);
+            this.refreshTables();
+          },
+          error: (error: BawApiError) => {
+            // some of the default error messages are ambiguous on this page
+            // e.g. "you do not have access to this page" means that the user doesn't have access to the audio recording
+            if (error.status === FORBIDDEN) {
+              this.notifications.error(
+                "You do not have access to all the audio recordings or tags in your files."
+              );
```
atruskie: we're effectively squashing an error here

- make sure the error message passed back from the server is included
- file a bug against the API
- cross link issue reference

src/app/components/import-annotations/details/details.component.ts
```diff
@@ -0,0 +1,333 @@
...
+          },
+        });
+    });
+  }
+
+  // since the typeahead input returns an array of models, but the api wants the associated tags as an array of id's
+  // we use this helper function to convert the array of models to an array of id's that can be sent in the api request
+  protected getIdsFromAbstractModelArray(items: object[]): Id[] {
+    return items.map((item: AbstractModel): Id => item.id);
+  }
```
atruskie: seems like a job for a model association?

src/app/components/import-annotations/edit/edit.component.ts
```diff
@@ -0,0 +1,69 @@
...
+{
+  public constructor(
+    private api: AudioEventImportService,
+    route: ActivatedRoute,
+    notifications: ToastrService,
+    router: Router
+  ) {
+    super(notifications, route, router, {
+      getModel: (models) => models[audioEventImportKey] as AudioEventImport,
+      successMsg: (model) => defaultSuccessMsg("updated", model.name),
```
atruskie, line 46: don't abbreviate

src/app/components/import-annotations/new/new.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  annotationsImportCategory,
+} from "../import-annotations.menu";
+
+export const newAnnotationMenuItemActions = [
+  annotationsImportMenuItem,
+  newAnnotationImportMenuItem,
+];
+
+@Component({
+  selector: "baw-new-annotation",
```
atruskie: incorrect name

src/app/models/AbstractModel.ts
```diff
@@ -114,22 +114,41 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
       const snakeCaseAttr = snakeCase(attr);
-      output.append(`${modelName}[${snakeCaseAttr}]`, data[attr]);
+
+      // file lists are commonly uploaded via FormData
+      // however, a FileList is actually an object and not an array that implements an iterator
+      // we therefore need to convert it to an array before appending it to the FormData
+      if (dataValue instanceof FileList) {
+        Array.from(dataValue).forEach((value) =>
+          output.append(`${modelName}[${snakeCaseAttr}]`, value)
+        );
```
atruskie: why not just:
```suggestion
       for (const value of dataValue) {
          output.append(`${modelName}[${snakeCaseAttr}]`, value)
        );
```

src/app/models/AudioEvent.ts
```diff
@@ -33,6 +41,7 @@ export interface IAudioEvent extends HasAllUsers {
   isReference?: boolean;
   taggings?: ITagging[] | Tagging[];
   provenanceId?: Id;
+  errors?: AudioEventError[];
```
atruskie: this is not a property of an audio event. You'll need a different solution here.

I realize the API returns it like this. But yeah, a property for a general model that's only available at one endpoint seems prone to cause further confusion

src/app/models/AudioEvent.ts
```diff
@@ -63,6 +72,8 @@ export class AudioEvent
   @bawDateTime()
   public readonly deletedAt?: DateTimeTimezone;
   public readonly provenanceId?: Id;
+  @bawReadonlyConvertCase(false)
+  public readonly errors?: AudioEventError[];
```
atruskie: same as above

atruskie (changes requested):

src/app/components/import-annotations/details/details.component.html
hudson-newey: Removed

src/app/components/import-annotations/details/details.component.html
hudson-newey: Changed to `annotationsFileInput`

src/app/components/import-annotations/details/details.component.ts
hudson-newey: This makes sense to me. Fixed

image

src/app/components/import-annotations/edit/edit.component.ts
hudson-newey: Easy to fix. Just letting you know: this will change a large number of files as this is defined in our [formTemplate.ts](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/helpers/formTemplate/formTemplate.ts#L20-L32)

image

src/app/components/import-annotations/new/new.component.ts
hudson-newey: Revised name: `baw-new-annotation-import`

src/app/models/AudioEvent.ts
hudson-newey: This is a sub-model, so I'll be creating a new submodel on `AudioEventImport` that implements and extends the `AudioEvent` class/interface and extends it with errors

src/app/components/import-annotations/details/details.component.scss
hudson-newey: I was trying to mimic the formly required star (I got mistaken and thought it was baw-danger)

image

Surprisingly, we don't have a class (nor does Bootstrap) for a purely stylistic required symbol ("*").

We could certainly add this astrix in the HTML template

src/app/components/import-annotations/details/details.component.ts
hudson-newey: > make sure the error message passed back from the server is included

Done

---

> file a bug against the API

Updated https://github.com/QutEcoacoustics/baw-server/issues/664 with the bug

src/app/components/import-annotations/details/details.component.html
hudson-newey: Not quite sure what you are looking for here

Would the correct approach be to condense multiline attribute into a single line?
image
(ruler lines drawn at 80 and 120 characters)

src/app/components/import-annotations/details/details.component.html
hudson-newey: > incorrectly indicate a field is required (elsewhere we use an asterisk to indicate required fields)

Removed

> incorrectly indicated an error state on the form's first display

My bad, I've changed form validation errors from `--baw-danger` to the bootstrap form validation `is-invalid` class

src/app/components/import-annotations/details/details.component.scss
atruskie: ok sounds like a general style to add regardless (and without the danger color)?

src/app/components/import-annotations/details/details.component.html
atruskie: i'm not sure but I really don't like multiline attributes.

- I'm fine with long strings on isolated lines
- ngbtooltip accepts template references
- the most correct is if we had all strings in a translations file 🙃 but that is a project for another time

src/app/components/import-annotations/edit/edit.component.ts
atruskie: separate PR then

src/app/components/import-annotations/details/details.component.html
hudson-newey: Agreed, a translations file is the most propper approach, but in the interim, I prefer the "long strings on isolated lines"

As I think using a template here would unnecessarily obfuscate the tooltip contents / value

src/app/components/import-annotations/details/details.component.ts
hudson-newey: This function is needed because the typeahead-input returns a list of tag models, while the event import requires an array of tag id's.

I'm not sure how I'd use a model association for this??

src/app/components/import-annotations/details/details.component.html
hudson-newey: I like this syntax.

Should we add this to the scope of this PR or do it in another "code cleanup" PR with the `successMsg` and `errorMsg` form template refactoring (change name from `successMsg` -> `successMessage` & `errorMsg` -> `errorMessage`)?

src/app/components/import-annotations/details/details.component.ts
atruskie: oh right. Our associations code translates an array of ids to an array of resolved objects.

When we update an array of resolved objects, we should update the original array of ids automatically. 
Thus, when the serialization happens the ids are already there.

~Might be out of scope of this PR~ is out of scope. Our associations are one-way only. nvm

src/app/components/import-annotations/details/details.component.html
atruskie: Separate PR

src/app/services/baw-api/baw-api.service.ts
atruskie, line 1: I just noticed you only changed `update`, we should change `create` as well for parity.

Also I can't help but wonder if there's a better way to orchestrate this function.

Can we chain these observables together?

hudson-newey: > I just noticed you only changed update, we should change create as well for parity.

Done

> Can we chain these observables together?

image

atruskie (approved):

PR #2096, opened 2023-12-06
Create Date/Time Component

### LINKED ISSUE #2092: New DateTime Component
We should add a DateTime component to display date/times in a standardised ISO 8601 format.

Relates to https://github.com/QutEcoacoustics/workbench-client/issues/346, https://github.com/QutEcoacoustics/workbench-client/issues/345

Charles tried a previous implementation in #1736, however, it has diverged significantly from `master` and therefore should be rejected and started from scratch.

We should use a component over a template pipe as it allows styling to indicate dateTime's that are not relative to the users timezone.

image
**The component should have a dotted underline styling for `localTime` dateTime's**

---

In short: all datesTimes should be relative to the user group that created them

* Recordings and events should always use the timezone that the location the recording/event was captured. (note: the creation date/time of the event should be in the users local time zone)
* User generated events such as harvest uploads and edits should use user local time

e.g. If a recording was captured at 11:11 am (UTC+ 10), the client should show 11:11 am with a tooltip on however showing the timezone (AEST UTC + 10).

If an event is created, the `createdAt` dateTime should be relative to the user. e.g. if a user in UTC + 8 creates an event at `T00:00Z+08:00`, a user viewing the event at UTC+10 should see the `createdAt` value as `2:00 am` with a tooltip on hover showing `T00:00Z+08:00`.  
However, the events dateTime should always be relative to the recording and not be effected by user locals.

---

### Prefered spec

```html
<baw-date [options]="{ showTime: true, localTime: true }"></baw-date>
```

where `DateTimeOptions` is an interface that looks like

```ts
interface {
	showTime: boolean;
	localTime: boolean;
}
```

### Alternative spec

```html
<baw-date [showTime]="true" [localTime]="true"></baw-date>
```

hudson-newey:
# Create Date/Time Component

Given we use so many Date/Times formats throughout the client, it can be ambiguous to a user if the date/time is localised to their timezone, and what timezone a date/time is captured in.

## Changes

- Created new `baw-datetime` component to show localised instants
- Created new `baw-zoned-datetime` component to show non-localised events
- Created new `baw-duration` component to show times & duration's in sexagesimal, ISO 8601, and humanized formats
- Created new `baw-time-since` component to show what period of time has passed since a date/time (or just pass in a Luxon Duration)
- Created admin page to test new date/time components
- Created tests for all the above
- Changed time mocking in [src/app/components/harvest/pages/list/list.component.spec.ts](src/app/components/harvest/pages/list/list.component.spec.ts) to use the new `withDefaultZone` helper function (`Settings.defaultZone`) was not being unset propperly, leaking into other tests and causing them to fail
- Added new `assertTooltip` test helper function that asserts the tooltip contents of an element
- Added new `withDefaultZone` test helper function that mocks the test runners timezone / `Settings.defaultZone`
- Enabled [`@typescript-eslint/prefer-as-const`](https://typescript-eslint.io/rules/prefer-as-const/) eslint rule to bring us more into line with the Angular recommended TS rules
- Upgrade Luxon dependency from 4.3.3 -> 4.3.4
- Upgrade CI workflows (GitHub Actions) to Node.js 18 (in-line with the reset of the client)

## Problems

_Implementation on live pages was not completed as part of this PR_

## Issues

Fixes: #2092 
Closes: #1736 

### Pages

image
**New Date & Time examples admin page**

---

### Components

image
**configuration for all visuals under the "Components" section**

---

image
**`baw-datetime`**

- Default dotted underline that comes from using our new `.has-context` CSS class
- Notice: 2 hours have been added, because the DateTime object input is from Australia/Perth (UTC+08:00), and these tests were run in Australia/Brisbane (UTC+10:00)

---

image
**`baw-zoned-datetime` (Implicit)**

- Uses the `.has-context` CSS class
- Notice: The dashed underline instead of the dotted underline
- Notice: The date/time emitted is the date/time in the original timezone (Australia/Perth) and the date/time has not been localised to the timezone where the tests were run in (Australia/Brisbane, UTC+10:00)

---

image
**`baw-zoned-datetime` (Explicit)**

- Uses the `.has-context` CSS class
- Notice: The dashed underline instead of the dotted underline
- Notice: The original implicit dateTime is from _Australia/Perth_ (UTC+08:00), and I have set the site timezone to _Australia/Darwin_ (UTC+09:30. Therefore, the _Australia/Perth_ time has had 01:30 hours added to it.

---

image
**`baw-duration`**

- Uses the `.has-context` CSS class

---

image
**`baw-time-since`**

- Uses the `.has-context` CSS class
- Dotted underline (default provided by `.has-context`) using `--baw-highlight` colour

## Final Checklist


src/app/components/shared/datetime/abstract-datetime.component.html
```diff
@@ -0,0 +1,7 @@
+<!--
+  This is a shared template used by datetime, zoned datetime, duration, and time since components
+  and should be used in conjunction with the AbstractDatetimeComponent.
+-->
+<time [dateTime]="rawDateTime" [ngbTooltip]="tooltipValue()">
+  {{ formattedValue() }} {{ suffix() }}
```
atruskie: it would be better if we could bind to fields here rather than functions

src/app/components/shared/datetime/abstract-datetime.component.ts
```diff
@@ -0,0 +1,28 @@
+export interface IAbstractDatetimeComponent {
+  formattedValue: () => string;
+  tooltipValue: () => string;
+  rawDateTime: () => string;
+  suffix: () => string;
+}
```
atruskie: this interface should not be needed

src/app/components/shared/datetime/abstract-datetime.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  public constructor() {}
+
+  public abstract formattedValue(): string;
+  public abstract tooltipValue(): string;
+  public abstract rawDateTime(): string;
+
+  // I have implemented this method so that each datetime component will default to having no suffix
+  public suffix(): string {
+    return "";
+  }
```
atruskie: i'm going to suggest redefining the shape of this class.

```suggestion
  public formattedValue: string;
  public tooltipValue: string;
  public rawDateTime: string = "";

 public abstract update() { };
```

src/app/components/shared/datetime/datetime/datetime.component.scss
```diff
@@ -0,0 +1,3 @@
+time {
+  border-bottom: 1px dotted;
+}
```
atruskie: so `<abbr>` uses dotted bottom border too.

Perhaps rather than a component specific style we want add some semantics to this. Something like `.has-context {}` in our shared styles section.

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    imports: [SharedModule, MockBawApiModule],
+  });
+
+  function setup(): void {
+    // to ensure that tests pass for all timezones, we set the default timezone to Australia/Perth
+    // this mocks the test runners timezone to be Australia/Perth so that they will assert correctly in all timezones
+    Settings.defaultZone = "Australia/Perth";
+    Date.prototype.getTimezoneOffset = () => {
+      return 8 * 60; // +08:00 hours in minutes (8 * 60)
+    };
```
atruskie: I don't see where you are cleaning up these changes. Won't this impact other tests?

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    // utc datetime should be localized to "Australia/Perth"
+    let fakeDateTime = DateTime.fromISO("2020-01-01T12:10:11.000Z");
+    fakeDateTime = fakeDateTime.toUTC();
```
atruskie: why are we setting the components value here, when you override it in each test?

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
atruskie, line 1: i found these tests difficult to follow. There is a lot of boiler plate here that makes it confusing. 

Have you considered a more data oriented approach?

```
const testCases = [
  { name: '',  value: new Date(...), expectedText: '...', expectedTooltip: '...', dateOnly: false, timeOnly: false }
  { name: '',  value: new DateTime(..., 'Australia/Syndey'), expectedText: '...', expectedTooltip: '...', dateOnly: false, timeOnly: false }
]

testCases.forEach((testCase( => {

  describe (testCase.name, () => {
    before(()=>    spectator.component.value = testCase.value; spectator.detectChanges(); );

     it('outputs the expected text', ()=> { ... } );

     it('has the correct tooltip', ()=> { ... } );
  });
})
```

You can resuse this structure for each of your concrete components

atruskie (changes requested):
I've done a partial review and have suggested significant architectural changes. 

The tooltips for the `baw-duration` components should probably not use the ISO8601 format. Maybe inlcude sexagesimal as well as ISO8601?  Or a humanized form? I'm not sure what's best, but although unambiguous, i think most users would find that tooltip does not add much context.

The `baw-time-since` tooltip should show a date in the tooltip. Since it's the duration elapsed from an instant, the context we need to know is what instant is our reference point.

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
hudson-newey: Yes, oversight on my part

src/app/components/shared/datetime/datetime/datetime.component.scss
hudson-newey: I like the idea of a shared class

However, Using a shared class on the shared `abstract-datetime.component.html` template means that each datetime component will have the same `.has-context` indicator style.

eg. The zoned-datetime.component.ts will have the same visual and semantic styles as the datetime.component.ts

We are currently visually distinguishing these classes by changing the style and colour of the `bottom-border`, as each component has its own associated style sheet.

To maintain the differences in indicator styles, we could either:

- Retain the separate stylesheets for each component (and override with `!important` the bottom-border style for duration's since we we a dashed border style instead of dotted) and add colour to the borders in each component
- OR make another attribute on our `abstract-datetime.component.ts` called `indicator-class` (or the like) (not ideal)

hudson-newey:
> The tooltips for the baw-duration components should probably not use the ISO8601 format. Maybe inlcude sexagesimal as well as ISO8601? Or a humanized form? I'm not sure what's best, but although unambiguous, i think most users would find that tooltip does not add much context.

Mock-up of sexagesimal + ISO8601

image

---

Mock-up of humanized + ISO8601

image

I think I prefer _sexagesimal + ISO8601_

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    expectedDateTimeAttribute: "2020-01-01T09:10:11.000+08:00",
+  },
+  {
+    name: "should increment date if timezone localization causes the date to increment",
+    value: DateTime.fromISO("2020-01-01T23:10:11.000Z"),
+    expectedText: "2020-01-02 07:10:11",
+    expectedTooltip: "2020-01-02 07:10:11.000 Australia/Perth +08:00",
+    expectedDateTimeAttribute: "2020-01-02T07:10:11.000+08:00",
+  },
+  {
```
atruskie, line 95: Do you really need test this? Isn't this basically testing the luxon library, which is already tested?

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    name: "should display full datetime if both dateOnly and timeOnly are set",
+    value: DateTime.fromISO("2020-01-01T12:10:11.000Z"),
+    expectedText: "2020-01-01 20:10:11",
+    expectedTooltip: "2020-01-01 20:10:11.000 Australia/Perth +08:00",
+    dateOnly: true,
+    timeOnly: true,
+    expectedDateTimeAttribute: "2020-01-01T20:10:11.000+08:00",
+  },
+  {
```
atruskie: this combination of attributes doesn't make sense - semantically. It should either

- be an error to combine these attributes
- or we should rename them to something that makes more sense

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
atruskie: do we really need to test this? Isn't this effectively just texting luxon itself?

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    name: "should localize a Luxon DateTime object with a timezone offset",
```
atruskie: How is this test different from https://github.com/QutEcoacoustics/workbench-client/blob/1716a708e17677a62c0e488fd7d26c0130a7451e/src/app/components/shared/datetime/datetime/datetime.component.spec.ts#L68-L69

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    expectedDateTimeAttribute: "2020-01-01T20:10:11.000+08:00",
+  },
+  {
+    name: "should localize a JavaScript date/time with an offset",
+    value: new Date("2020-01-01T12:10:11.000+01:00"),
+    expectedText: "2020-01-01 19:10:11",
+    expectedTooltip: "2020-01-01 19:10:11.000 Australia/Perth +08:00",
+    expectedDateTimeAttribute: "2020-01-01T19:10:11.000+08:00",
+  },
+];
```
atruskie: missing test cases for negative offsets and western timezones

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    expectedDateTimeAttribute: "2020-01-01T19:10:11.000+08:00",
+  },
+  {
+    name: "should handle a date/time object with fractional seconds",
+    value: DateTime.fromISO("2020-01-01T12:10:11.123Z"),
+    expectedText: "2020-01-01 20:10:11",
+    expectedTooltip: "2020-01-01 20:10:11.123 Australia/Perth +08:00",
+    expectedDateTimeAttribute: "2020-01-01T20:10:11.123+08:00",
+  },
+  {
```
atruskie: this case seems redunant, only in that if you added fractional seconds to all inputs you test, you'd cover all cases for fractional seconds and then this test wouldn't be needed

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+// we re-mock the test runners timezone to a more complex one when the tests are run
+// however, during test setup it is easier to read the DateTime objects if we create
+// them as if the test runner is in utc+00:00
+Settings.defaultZone = "utc";
+
+// in the following test cases, all the dateTime objects are created in UTC+00:00
+// therefore, because we are mocking the test runners timezone to be Australia/Perth (UTC+08:00)
+// we should see that the date time text, tooltips and machine readable content is localized
+// to UTC+08:00. eg. We should see the hours increase by 8
+const testCases: TestCase[] = [
```
atruskie: greatly simplify the names of these test cases.

Data driven test cases don't usually have names. They're a grouping only with useful diagnostic information. If they're long sentences the reader had to parse them to understand the differences.

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+import { isInstantiated } from "@helpers/isInstantiated/isInstantiated";
+import { TimezoneInformation } from "@interfaces/apiInterfaces";
+import { DateTime, IANAZone, Zone } from "luxon";
+import { AbstractDatetimeComponent } from "../abstract-datetime.component";
+
+type BawTimezoneUnion = Zone | TimezoneInformation | string;
+
+/**
+ * Displays a datetime (possibly from another timezone) in the users locale and timezone
+ * Commonly used for displaying client side timestamps e.g. Project created at: xyz
```
atruskie:
```suggestion
 * Commonly used for displaying instants (real dates that happened in no particular timezone)  e.g. timestamps like `createdAt` and `updatedAt`.
```

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * This component does not assume that all dates input are utc, and will localize using ISO8601 offset information if possible
```
atruskie: This comment seems redunant.

```suggestion
 * This component always shows a date in the user's  local timezone.
```

src/app/components/shared/datetime/abstract-datetime.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  standalone: true,
+  imports: [NgbTooltipModule],
+})
+export abstract class AbstractDatetimeComponent implements OnChanges {
+  public constructor() {}
+
+  public formattedValue: string;
+  public tooltipValue: string;
+  public rawDateTime: string;
+  // I have implemented this method so that each datetime component will default to having no suffix
```
atruskie:
```suggestion
  // Each child component will default to no suffix.
```

src/app/components/shared/datetime/abstract-datetime.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  public suffix = "";
+
+  /**
+   * A method that should update formattedValue, tooltipValue, and rawDateTime
+   * after during each change detection cycle the component updated as part of
```
atruskie:
```suggestion
   * before each change detection cycle
```

src/app/components/shared/datetime/abstract-datetime.component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { NgbTooltipModule } from "@ng-bootstrap/ng-bootstrap";
+
+// use use a standalone component here so that it implements all its
+@Component({
+  selector: "baw-abstract-datetime",
+  templateUrl: "./abstract-datetime.component.html",
+  standalone: true,
+  imports: [NgbTooltipModule],
+})
+export abstract class AbstractDatetimeComponent implements OnChanges {
```
atruskie: comment truncated?

why is this a component at all? It's never used anywhere? ~~Just remove the decorator?~~

So it does need a decorator, but the extra information here is essentially unused? So don't repeat it.

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+@Component({
+  selector: "baw-datetime",
+  templateUrl: "../abstract-datetime.component.html",
+})
+export class DatetimeComponent extends AbstractDatetimeComponent {
+  public constructor() {
+    super();
+  }
+
+  @Input({ required: true }) public value!: DateTime | Date;
```
atruskie: shouldn't this be pushed up to the base class?

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+  protected get luxonDateTime(): DateTime {
+    const luxonDate: DateTime =
+      this.value instanceof Date
+        ? DateTime.fromJSDate(this.value, { zone: "utc" })
+        : this.value;
+
+    // localize the date/time to the users timezone using toLocal()
+    return luxonDate.toLocal();
+  }
+
```
atruskie: wouldn't this be better modelled as a setter on `value`? If it is read more than once we're doing unnecessary work

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+  @Input() public dateOnly: string | boolean;
+  @Input() public timeOnly: string | boolean;
+  protected timezone?: BawTimezoneUnion;
+
+  // we create a getter for the a luxon DateTime object so that we can get values from the @Input() value
+  // without type guards, shape, and instance checks
+  protected get luxonDateTime(): DateTime {
+    const luxonDate: DateTime =
+      this.value instanceof Date
+        ? DateTime.fromJSDate(this.value, { zone: "utc" })
```
atruskie: `"utc"` is not the name of the Zone. Why is it lower cased? Can we use `FixedOffsetZone.utcInstance` instead?

atruskie: I'm also wondering why we have to specify zone at all - especially since we just `.toLocal` 4 lines down...

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+        : this.value;
+
+    // localize the date/time to the users timezone using toLocal()
+    return luxonDate.toLocal();
+  }
+
+  public update(): void {
+    // the fullDateTime is used regardless of format as it is used in the tooltip
+    // it emits as much information as possible
+    const fullDateTime = this.luxonDateTime.toFormat("yyyy-MM-dd HH:mm:ss.SSS");
```
atruskie: this can't be the only place this format string is used?
Refactor to constant, maybe onto abstract component?

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (isInstantiated(this.dateOnly) && this.dateOnly !== false) {
+      return "yyyy-MM-dd";
+    } else if (isInstantiated(this.timeOnly) && this.timeOnly !== false) {
+      return "HH:mm:ss";
+    }
+
+    // include both date and time if neither dateOnly or timeOnly are specified
+    return "yyyy-MM-dd HH:mm:ss";
+  }
+
```
atruskie: why "we should prefer not to use this functionality"?

also:
- constantize the strings
- check values only once
- simplify the branches.

```
const showDate = !(this.dateOnly === "" || this.dateOnly === true)
const showtime = !(this.timeOnly === "" || this.timeOnly === true)

if (showDate && showTime ) {
    return FULL_DATE;
}
else if (showDate) {
    return DATE_ONLY;
}
else {
    return TIME_ONLY;
}
```

and if we take advantage of the new `transform` option for `@Input` we can do even better: https://stackoverflow.com/a/76537839/224512

atruskie: and again, I haven't looked yet but this seems like functionality that belongs in the abstract class

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+})
+export class DatetimeComponent extends AbstractDatetimeComponent {
+  public constructor() {
+    super();
+  }
+
+  @Input({ required: true }) public value!: DateTime | Date;
+  @Input() public dateOnly: string | boolean;
+  @Input() public timeOnly: string | boolean;
+  protected timezone?: BawTimezoneUnion;
```
atruskie: why is this here? it makes no sense...

src/app/components/shared/datetime/datetime/datetime.component.ts
```diff
@@ -0,0 +1,132 @@
...
+  // timezoneName from the tooltip in this case
+  private isIanaTimezone(timezoneName: string): boolean {
+    return !(
+      timezoneName.startsWith("UTC") ||
+      timezoneName === "utc" ||
+      timezoneName === "system" ||
+      timezoneName === "local"
+    );
+  }
+}
```
atruskie: this logic seems overly complex and also not related to this class?

atruskie (changes requested):
Ok, we need to take another pass at this. 

There is a lot of complexity here that seems to be misplaced.

atruskie (changes requested):
Ok, we need to take another pass at this. 

There is a lot of complexity here that seems to be misplaced.

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: In retrospect, this could also be implemented as a simple Fizz Buzz method

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
hudson-newey: Good point, no; we don't need this test

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: Yes, fixed.

I originally did this for stricter TypeScript typing (so that each component could define it's value type)

However, by making the AbstractDatetimeComponent a generic abstract class, I was able to define the TS type on the abstract component

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
hudson-newey: Throwing a runtime console error works for me.

My original justification as to why we might want both was:

Imagine we have a place in the website where we want to conditionally show date, time or both date & time.  
We should be able to do something like `<baw-datetime [timeOnly]="true" [dateOnly]="true" />` (where `true` is set reactively by view model values)

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
hudson-newey: It's supposed to be different because it contains an offset

However, as stated in a following comment: By adding offsets to all tests, we remove the need for this test.

I have removed this test

hudson-newey: Good catch, removed

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: I've moved all timezone logic to the `zoned-datetime.component.ts`

---

This was originally here because the `datetime.component.ts` needs to have knowledge of the users local timezone anyways (for tooltips).

Therefore, I made the `zoned-datetime.component.ts` and `datetime.component.rs` share the same timezone logic. The `value` (fetched from `luxonDateTime`) was localized in the `datetime.component.ts`, but not the `zoned-datetime.component.ts`.

This meant that the `datetime.component.ts` could actually just using the implicit timezone logic when creating the tooltip

Essentially the only difference between the `zoned-datetime.component.ts` and `datetime.component.ts` was that you could overwrite what timezone it localized to  using the `timezone` `@Input()` (if it was undefined it'd localize to the users local timezone).

_In hindsight, setting the initialization value of `timezone` in the `datetime.component.ts` to `DateTime.now().zone` would make more sense than binding functionality to `undefined`._

src/app/components/shared/datetime/abstract-datetime.component.ts
hudson-newey: Sadly, there is no way to remove the templateUrl: "" as all angular components must have a template (there is no such thing as an abstract component without a template).

Therefore, the best I can do is
image

Without hacking something together using abstract directives

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: I've moved most of the timezone logic to the `zoned-datetime.component.ts`. However, since the `datetime.component.ts` and `time-since.component.ts` both require the users local timezone, I'm moving methods to get the users local timezone to the abstract template coponent

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: We can't use a setter since we're using a props `@Input()`, however, we can apply a transform function to the value (similar to the functionality of a setter)

image

In my mind: This satisfies the objective of not having to reshape the object every time we use the value in the component

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: By the looks of it, we don't need the { zone: "utc" }

---

> "utc" is not the name of the Zone

Funny enough, this luxon `{ zone }` [options object doesn't actually have to be a zone (or even an IANA identifier!)](https://github.com/moment/luxon/blob/3125686af82d9a25c7267a1cf1eb838a3d41144f/docs/zones.md?plain=1#L55-L58) (weird in my opinion)

This is also the reason why I've had to implement a helper function to check if the `zoneName` returned by a zone object is actually an IANA identifier, or if Luxon has used one of their custom identifiers

image

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
atruskie: Nah I'm really suggesting changing the design spec semantics of these components. (changing the names)

- if no modifier attributes are set then show standard format
- if `date` is set show the date
- if `time` is set show the time
- if `<baw-datetime value=... date time seconds zone>` then show all of those components

Proposed change allows for later extensions (`zone`, `seconds`) down the line without reading non-sensically (`<baw-datetime value=... dateOnly timeOnly secondsOnly zoneOnly>`) and still allows for binding (and it even makes more sense since the booleans aren't negated)

src/app/components/shared/datetime/datetime/datetime.component.ts
atruskie: discussed out of band: transform function is for text/template set/attribute values only and not appropriate here

src/app/components/shared/datetime/datetime/datetime.component.ts
atruskie: I didn't get time to understand that isIanaTimezone function but it was a huge code smell. My gut instinct (without fully understanding everything here) is that we should just leave this to luxon.

src/app/components/shared/datetime/datetime/datetime.component.ts
atruskie: I understand your reasoning but from a functional composition perspective you still had one type import and store a property of a type it itself could never use.

src/app/components/shared/datetime/datetime/datetime.component.spec.ts
hudson-newey: Done

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: ditto above

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: Done

image

src/app/components/shared/datetime/datetime/datetime.component.ts
hudson-newey: I was able to replace this function with [`IANAZone.IsValidZone`](https://github.com/moment/luxon/blob/3125686af82d9a25c7267a1cf1eb838a3d41144f/src/zones/IANAZone.js#L101-L111)

Alternatively, I could have used [`Info.isValidIANAZone`](https://github.com/moment/luxon/blob/3125686af82d9a25c7267a1cf1eb838a3d41144f/src/info.js#L29-L31)

src/app/components/shared/datetime-formats/abstract-template.component.html
```diff
@@ -0,0 +1,8 @@
+<!--
+  This is a shared template used by datetime, zoned datetime, duration, and time-since components
+  and should be used in conjunction with the abstract-template.component.ts
+-->
+<time class="has-context" [dateTime]="isoDateTime">
+  <!-- We place a span inside the time element because [NgbTooltip] cannot bind to a time element -->
```
atruskie: So close to answering my question: but why can't it? Link me?

src/app/components/shared/datetime-formats/abstract-template.component.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+  });
+
+  it("should create", () => {
+    expect(abstractTemplateComponent).toBeTruthy();
+  });
+});
+
+class AbstractTemplateComponentMock extends AbstractTemplateComponent<any> {
+  public update(): void {}
+}
```
atruskie: these tests are useless - they don't test anything novel (the other controls test inheritability themselves)

src/app/components/shared/datetime-formats/abstract-template.component.ts
```diff
@@ -0,0 +1,49 @@
...
+})
+export class AbstractTemplateComponent<InputType, InternalValueType = InputType> implements OnChanges {
+  public constructor() {}
+
+  // we use getters and setters on the [input] prop so that we can change the shape of the input
+  // to a uniform type. This allows us to take multiple types as input and convert them into a single type
+  // eg. the [input] can take a Luxon DateTime, JS Date, or string and convert this.value into a Luxon DateTime
+  @Input({ required: true })
+  public set value(value: InputType) {
+    this._value = value as any;
```
atruskie: this setter doesn't actually do what your comment says.

Is the coercion done is the child classes?

In which case this setter should never be used? Should probably throw a not implemented error and then that would avoid this any cast

src/app/components/shared/datetime-formats/abstract-template.component.ts
```diff
@@ -0,0 +1,49 @@
+import { Component, Input, OnChanges } from "@angular/core";
+
+@Component({
+  templateUrl: "./abstract-template.component.html",
+})
+export class AbstractTemplateComponent<InputType, InternalValueType = InputType> implements OnChanges {
```
atruskie: Name suggestion for `InternalValueType`:

- `NormalizedType`
- `CoercedType`

atruskie: Also why does it default to `InternalValueType = InputType` ? Surely we want to either leave the internal type undefined, so the implementer has to define it? Or just set it to the luxon type(s) we're expecting...

src/app/components/shared/datetime-formats/abstract-template.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this._value;
+  }
+
+  // the ISO dateTime is used in the <time> elements "datetime" attribute
+  // this attribute is used by screen readers and web scrapers to determine the date and time
+  public isoDateTime: string;
+  public documentText: string;
+  public tooltipText: string;
+  public suffix = "";
+  protected _value: InternalValueType;
```
atruskie: hmm should this not be private? we have a public getter

atruskie (changes requested):

src/app/components/admin/admin.module.ts
```diff
@@ -28,7 +28,7 @@ const components = [
   AdminUserListComponent,
   AdminThemeTemplateComponent,
   AllUploadsComponent,
-  DateTimeExampleComponent,
+DateTimeExampleComponent,
```
atruskie:
```suggestion
  DateTimeExampleComponent,
```

src/app/components/shared/datetime-formats/datetime/abstract-datetime.component.ts
```diff
@@ -1,67 +1,85 @@
...
+  @Input({ transform: booleanAttribute }) public date?: boolean;
+  @Input({ transform: booleanAttribute }) public time?: boolean;
 
-  public override get value(): DateTime {
-    return this._value as DateTime;
-  }
+  protected abstract extractTimezone(): Zone;
+
+  // this method returns the timezone name and offset
+  // eg. "Australia/Perth +08:00"
```
atruskie:
```suggestion
  // eg. "+08:00 Australia/Perth"
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.ts
```diff
@@ -11,42 +12,22 @@ import { AbstractDatetimeComponent } from "../abstract-datetime.component";
...
-      return offsetValue;
-    }
+  public override extractTimezone(): Zone {
+    return SystemZone.instance;
+  }
 
-    return `${timezoneName} ${offsetValue}`;
+  protected override normalizeValue(value: InputTypes) {
+    const unlocalizedValue = super.normalizeValue(value);
+    return unlocalizedValue.setZone(SystemZone.instance);
```
atruskie: this should be redundant now given we always apply a zone in `update`, see `valueInZone`

src/app/components/shared/datetime-formats/time-since/time-since.component.ts
```diff
@@ -1,56 +1,66 @@
...
+    // this is because all the functionality is handled by takeUntilDestroyed
+    // eslint-disable-next-line rxjs-angular/prefer-takeuntil
+    this.tick.subscribe(() => {
+      const elapsedTime = this.value.diffNow().rescale();
 
-    this.startDate = DateTime.now().minus(durationDifference);
+      // we want to update the "time since" every second if the time is less than a minute
+      // if the time since is greater than a minute, we want to update the value every minute
+      // because the observable will trigger every second, using elapsedTime.seconds < 1
+      // will cause the value to update every
```
atruskie:
```suggestion
      // will cause the value to update roughly every minute
```

src/app/components/harvest/pages/list/list.component.spec.ts
```diff
@@ -118,6 +119,8 @@ describe("ListComponent", () => {
     )?.nativeElement as T;
   }
 
+  beforeAll(() => testRunnersDefaultTimezone = Settings.defaultZone);
+
```
atruskie: Ok, this mocking behavior is repeated in several files. I wonder if we can't do better?

```javascript
// Helpers.ts

const withDefaultZone = function (zone, f) {
  const fullName = `With defaultZone: ${zone}`;
  describe(fullName, () => {
     beforeEach(() => {
         Settings.defaultZone = zone;
     });

    afterEach(() => {
        Settings.defaultZone = null;
    });
    
    f.call(this);
  });
};

// usage

Helpers.withDefaultZone(FixedOffsetZone.utcInstance, () => {

   it('blah blah blah' , () => { ... } );

});
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+const originalTestRunnerZone = Settings.defaultZone;
+Settings.defaultZone = "utc";
+
+// in the following test cases, all the dateTime objects are created in UTC+00:00
+// therefore, because we are mocking the test runners timezone to be Australia/Perth (UTC+08:00)
+// we should see that the date time text, tooltips and machine readable content is localized
+// to UTC+08:00. eg. We should see the hours increase by 8
+const testCases: TestCase[] = [
+  {
+    name: "Luxon DateTime outside test runners timezone",
```
atruskie:
```suggestion
    name: "UTC DateTime",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    // we should see this date be localized to UTC+08:00 (the test runners timezone)
+    // we do not call setZone() here like the other tests because we want to create the DateTime object in UTC+00:00
+    // so that we can see it localized to UTC+08:00
+    value: DateTime.fromISO("2020-01-01T12:10:11.123Z"),
+    expectedText: "2020-01-01 20:10:11",
+    expectedTooltip: "2020-01-01 20:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T20:10:11.123+08:00",
+  },
+  {
+    name: "Luxon DateTime in test runners timezone",
```
atruskie:
```suggestion
    name: "local DateTime",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// by setting the defaultZone to utc when we create DateTime objects from ISO strings
+// it doesn't implicitly localize the DateTime object to another timezone
+// this makes the object produced by the iso string predictable
+// eg. an iso string "2020-10-10T12:12:12Z" will always produce the object
+//     { year: 2020, month: 10, day: 10, hour: 12, minute: 12, second: 12 }
+// we re-mock the test runners timezone to a more complex one when the tests are run
+// however, during test setup it is easier to read the DateTime objects if we create
+// them as if the test runner is in utc+00:00
```
atruskie: i'm not at all convinced this is worthwhile - especially given it does not simulate a real world environment

additionally your tests should be unambiguous in their definitions

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+  {
+    name: "Luxon DateTime in test runners timezone",
+    // by setting the timezone to Australia/Perth, we should see that the time value is not changed
+    // because the Luxon DateTime object is already in the same timezone as the test runner
+    // by using { keepLocalTime: true }, we ensure that the time is not changed when we change the timezone
+    // without using { keepLocalTime: true }, the DateTime object would be implicitly created as 20:10:11 +08:00
+    value: DateTime.fromISO("2020-01-01T12:10:11.321Z").setZone(
+      "Australia/Perth",
+      { keepLocalTime: true }
+    ),
```
atruskie:
```suggestion
    value: DateTime.fromISO("2020-01-01T12:10:11.321+08:00").setZone("Australia/Perth"),
```

Setting up your data as it was is actually just incorrect. Sure you massage it into what you mean, but why write a UTC date, when you can just write a date in the correct offset? Any time I see `keepLocalTime` and you're not using it for massaging user input, then you're breaking the date.

atruskie: Further if this was really a test of a local time you should be constructing a local date from a local string!

```suggestion
    value: DateTime.fromISO("2020-01-01T12:10:11.321")
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    expectedText: "2020-01-01 12:10:11",
+    expectedTooltip: "2020-01-01 12:10:11.321 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T12:10:11.321+08:00",
+  },
+  {
+    // most of the tests create the DateTime objects in UTC+00:00
+    // for this test, we create th DateTime object in UTC+02:00
+    // we should therefore see the hours increase by 6
+    name: "Luxon DateTime with offset (no timezone)",
```
atruskie:
```suggestion
    name: "DateTime with offset",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    // in the users local timezone
+    value: DateTime.fromISO("2020-01-01T12:10:11.123+02:00"),
+    expectedText: "2020-01-01 18:10:11",
+    // notice that the tooltips are still in the test runners mock timezone
+    // this is because the emitted date/time has been localized to the test runners timezone
+    expectedTooltip: "2020-01-01 18:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T18:10:11.123+08:00",
+  },
+  {
+    name: "Luxon DateTime with dateOnly",
```
atruskie:
```suggestion
    name: "DateTime with attribute date",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    value: DateTime.fromISO("2020-01-01T12:10:11.123Z"),
+    expectedText: "2020-01-01",
+    expectedTooltip: "2020-01-01 20:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T20:10:11.123+08:00",
+    date: true,
+  },
+  {
+    name: "Luxon DateTime with timeOnly",
```
atruskie:
```suggestion
    name: " DateTime with attribute time",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    value: DateTime.fromISO("2020-01-01T12:10:11.123Z"),
+    expectedText: "20:10:11",
+    expectedTooltip: "2020-01-01 20:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T20:10:11.123+08:00",
+    time: true,
+  },
```
atruskie: missing test with both date and time attributes

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+  {
+    name: "Luxon DateTime with a negative offset (no timezone)",
```
atruskie:
```suggestion
    name: "DateTime with a negative offset",
```
Also move this up to be next to the related test

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+interface TestCase {
+  name: string;
+  value: Date | DateTime;
+  expectedText: string;
+  expectedTooltip: string;
+  expectedDateTimeAttribute: string;
+  date?: boolean;
+  time?: boolean;
+}
+
```
atruskie, line 20: If you make a function:

```
function testCase(name, value, expectedText, expectedTooltip, expectedAttribute, date, time) {
  return { name:, value:, expectedText:, expectedTooltip:, expectedAttribute:, date:, time: };
}
```

You could massive reduce the boiler plate in the following test case definitions

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+  },
+  {
+    name: "Luxon DateTime with a negative offset (no timezone)",
+    value: DateTime.fromISO("2020-01-01T12:10:11.123-01:00"),
+    expectedText: "2020-01-01 21:10:11",
+    expectedTooltip: "2020-01-01 21:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T21:10:11.123+08:00",
+  },
+  {
+    name: "Luxon DateTime with a western timezone",
```
atruskie:
```suggestion
    name: "DateTime with a western timezone",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    value: DateTime.fromISO("2020-01-01T12:10:11.123Z").setZone(
+      "America/Los_Angeles",
+      { keepLocalTime: true }
+    ),
```
atruskie: smell: keepLocalTime

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    expectedText: "2020-01-02 04:10:11",
+    expectedTooltip: "2020-01-02 04:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-02T04:10:11.123+08:00",
+  },
+  {
+    name: "JavaScript Date with no offset",
```
atruskie:
```suggestion
    name: "UTC Date",
```

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  function setup(): void {
+    // to ensure that tests pass for all timezones, we set the default timezone to Australia/Perth
+    // this mocks the test runners timezone to be Australia/Perth so that they will assert correctly in all timezones
+    Settings.defaultZone = "Australia/Perth";
+    mockSystemZone = "Australia/Perth";
+
+    spyOnProperty(SystemZone, "instance").and.callFake(() =>
+      IANAZone.create(mockSystemZone)
+    );
```
atruskie: why is this needed?

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    expectedDateTimeAttribute: "2020-01-01T19:10:11.123+08:00",
+  },
+  {
+    name: "JavaScript Date with a negative offset",
+    value: new Date("2020-01-01T12:10:11.123-01:00"),
+    expectedText: "2020-01-01 21:10:11",
+    expectedTooltip: "2020-01-01 21:10:11.123 +08:00 Australia/Perth",
+    expectedDateTimeAttribute: "2020-01-01T21:10:11.123+08:00",
+  },
+];
```
atruskie: haven't tested local Dates e.g. `new Date("2020-01-01T12:10:11.123")`

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
atruskie: furhter why is it in the general setup if it is only used in one test

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    const mockDateTime = DateTime.fromISO("2020-01-01T12:10:11.321Z");
+
+    spectator.component.value = mockDateTime;
+    update();
+
+    assertTooltip(timeElement(), expectedTooltip);
+  });
+
+  testCases.forEach((testCase) => {
+    describe(testCase.name, () => {
```
atruskie:
```suggestion
    describe('with ' + testCase.name, () => {
```

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+  expectedExplicitText: string;
+  expectedExplicitTooltip: string;
+  expectedExplicitDateTimeAttribute: string;
+
+  // implicit timezones should be added to the "value" field on the test case object
+  // this is typically done by using the Luxon setZone method
+  explicitTimezone: string;
+  date?: boolean;
+  time?: boolean;
+}
```
atruskie: DRY

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+  name: string;
+  value: DateTime | Date;
+
+  expectedImplicitText: string;
+  expectedImplicitTooltip: string;
+  expectedImplicitDateTimeAttribute: string;
+
+  expectedExplicitText: string;
+  expectedExplicitTooltip: string;
+  expectedExplicitDateTimeAttribute: string;
```
atruskie: separate your concerns, these should be individual test cases

atruskie (changes requested):
OK, I've reviewed one test file. The changes suggested should be applied to the other test files. Pay particular attention to naming.

Make use of the suggested timezone helper.

Cut down your comments. Some are outdated, some are incorrect.

I'll reveiw again when all the tests are of the same standard.

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
hudson-newey: Agreed, but I'd argue that it's harder to read

eg.

```js
{
    name: "DateTime with attribute time",
    value: DateTime.fromISO("2020-01-01T12:10:11.123Z"),
    expectedText: "2020-01-01",
    expectedTooltip: "2020-01-01 20:10:11.123 +08:00 Australia/Perth",
    expectedDateTimeAttribute: "2020-01-01T20:10:11.123+08:00",
    time: true,
},
```

would become:

```js
testCase(
	"DateTime with attribute time",
	DateTime.fromISO("2020-01-01T12:10:11.123Z"),
	"20:10:11.123",
	"2020-01-01 20:10:11.123 +08:00 Australia/Perth",
	"2020-01-01T20:10:11.123+08:00",
	false,
	true
)
```

OR

```js
testCase("DateTime with attribute time", DateTime.fromISO("2020-01-01T12:10:11.123Z"), "20:10:11.123", "2020-01-01 20:10:11.123 +08:00 Australia/Perth", "2020-01-01T20:10:11.123+08:00", false, true)
```

It'd remove the need for us to specify keys in our test cases, but I don't see this as a win for readability

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
hudson-newey: Removed

image

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
atruskie: You're not wrong. The trade-off is compactness making it easier to compare test cases.

atruskie: Your choice

hudson-newey: A dev preview of an implementation using the proposed factory function

image

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
hudson-newey: Agreed, I have removed these from my test declarations

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
hudson-newey: Since we're using `SystemZone.instance` to fetch the users local timezone in the `dateTime.component.ts`, we want to mock this getter so that it returns the mock timezone we are using for the `datetime.component.spec.ts` testbed.

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.ts
hudson-newey: Removed. This seems to have been a problem with how I was mocking `DateTime.now()`.

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -17,153 +18,73 @@ interface TestCase {
...
-  {
-    name: "JavaScript Date with a negative offset",
-    value: new Date("2020-01-01T12:10:11.123-01:00"),
-    expectedText: "2020-01-01 21:10:11",
-    expectedTooltip: "2020-01-01 21:10:11.123 +08:00 Australia/Perth",
-    expectedDateTimeAttribute: "2020-01-01T21:10:11.123+08:00",
-  },
+  test("UTC DateTime", DateTime.fromISO("2020-01-01T12:10:11.123Z"), "2020-01-01 20:10:11", "2020-01-01 20:10:11.123 +08:00 Australia/Perth", "2020-01-01T20:10:11.123+08:00"),
+
+  test("local DateTime", DateTime.fromISO("2020-01-01T12:10:11.321+08:00"), "2020-01-01 12:10:11", "2020-01-01 12:10:11.321 +08:00 Australia/Perth", "2020-01-01T12:10:11.321+08:00"),
```
atruskie: ok this test isn't actually testing a local date time though is it?

the datetime you constructed is an offset date time that happens to be in the same timezone as your tests are being test in.

if you move the case definitions into part of the code that already has a mocked local timezone then you can actually supply a local datetime and test it is parsed correctly

src/app/components/shared/datetime-formats/datetime/datetime/datetime.component.spec.ts
```diff
@@ -17,153 +18,73 @@ interface TestCase {
...
+
+  test("DateTime with attribute date", DateTime.fromISO("2020-01-01T12:10:11.123Z"), "2020-01-01", "2020-01-01 20:10:11.123 +08:00 Australia/Perth", "2020-01-01T20:10:11.123+08:00", true),
+
+  test("DateTime with attribute time", DateTime.fromISO("2020-01-01T12:10:11.123Z"), "20:10:11", "2020-01-01 20:10:11.123 +08:00 Australia/Perth", "2020-01-01T20:10:11.123+08:00", false, true),
+
+  test("DateTime with attributes date and time", DateTime.fromISO("2020-01-01T12:10:11.123Z"), "2020-01-01 20:10:11", "2020-01-01 20:10:11.123 +08:00 Australia/Perth", "2020-01-01T20:10:11.123+08:00", true, true),
+
+  test("UTC Date", new Date("2020-01-01T12:10:11.123Z"), "2020-01-01 20:10:11", "2020-01-01 20:10:11.123 +08:00 Australia/Perth", "2020-01-01T20:10:11.123+08:00"),
+
+  test("Local Date", new Date("2020-01-01T12:10:11.123+08:00"), "2020-01-01 12:10:11", "2020-01-01 12:10:11.123 +08:00 Australia/Perth", "2020-01-01T12:10:11.123+08:00"),
```
atruskie: same comment as above - not actually testing a local date and this test is essentially equivalent to the one below

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
```diff
@@ -1,214 +1,106 @@
...
-    expectedImplicitDateTimeAttribute: "2020-01-01T12:10:11.123Z",
-
-    expectedExplicitText: "21:40:11",
-    expectedExplicitTooltip: "2020-01-01 21:40:11.123 +09:30 Australia/Darwin",
-    expectedExplicitDateTimeAttribute: "2020-01-01T21:40:11.123+09:30",
-
-    time: true,
-  },
-];
+  test("Local DateTime", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2020-01-01 12:10:11", "2020-01-01 12:10:11.123 +09:30 Australia/Darwin", "2020-01-01T12:10:11.123+09:30"),
```
atruskie: same as previous

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
```diff
@@ -1,214 +1,106 @@
...
+
+  // explicit offset
+  test("DateTime with offset", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2020-01-01 03:40:11", "2020-01-01 03:40:11.123 +01:00", "2020-01-01T03:40:11.123+01:00", false, false, "+01:00"),
+
+  test("DateTime with negative offset", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2019-12-31 18:40:11", "2019-12-31 18:40:11.123 -08:00", "2019-12-31T18:40:11.123-08:00", false, false, "-08:00"),
+
+  test("DateTime with offset and time attribute", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2020-01-01 10:40:11", "2020-01-01 10:40:11.123 +08:00", "2020-01-01T10:40:11.123+08:00", false, false, "+08:00"),
+
+  test("DateTime with offset and time attribute", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2020-01-01 10:40:11", "2020-01-01 10:40:11.123 +08:00", "2020-01-01T10:40:11.123+08:00", false, false, "+08:00"),
```
atruskie: redundant cases or a mistake.

it seems from the name you want to test attribute cases but both cases are `false, false`. 

Either you need a third case and to change them to `true, false`, `false, true`, and `true, true` 

Or remove the redundant tests.

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
```diff
@@ -1,214 +1,106 @@
...
+  // explicit timezone
+  test("DateTime with timezone", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2020-01-01 10:40:11", "2020-01-01 10:40:11.123 +08:00 Australia/Perth", "2020-01-01T10:40:11.123+08:00", false, false, "Australia/Perth"),
+
+  test("DateTime with western timezone", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2019-12-31 18:40:11", "2019-12-31 18:40:11.123 -08:00 America/Los_Angeles", "2019-12-31T18:40:11.123-08:00", false, false, "America/Los_Angeles"),
+
+  test("DateTime with timezone and date attribute", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "2020-01-01", "2020-01-01 10:40:11.123 +08:00 Australia/Perth", "2020-01-01T10:40:11.123+08:00", true, false, "Australia/Perth"),
+
+  test("DateTime with timezone and time attribute", DateTime.fromISO("2020-01-01T12:10:11.123+09:30").setZone("Australia/Darwin"), "10:40:11", "2020-01-01 10:40:11.123 +08:00 Australia/Perth", "2020-01-01T10:40:11.123+08:00", false, true, "Australia/Perth"),
+
+  // implicit JavaScript Date objects
```
atruskie:
```suggestion
  // implicit JavaScript Date objects
  // JavaScript automatically converts any parsed date to a date represented in the local timezone
  // so all implicit timezones/offsets for Dates are in local time
```

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
```diff
@@ -217,100 +109,106 @@ describe("ZonedDateTimeComponent", () => {
...
-
-          describe(`${TimezoneType[timezoneType]} timezone type`, () => {
+      spectator.component.timezone = "Australia/Perth";
+      update();
+
+      expect(timeElement()).toHaveExactTrimmedText("2020-01-01 10:40:11");
+    });
+
+    for (const testCase of testCases) {
+      const componentTimezones = (): any[] => {
```
atruskie:
```suggestion
      // our component accepts time zones in different formats and we want to test each automatically
      // so generate extra cases for IANA time zone strings, luxon time zone objects, 
      // and our standard API time zone info objects
      const componentTimezones = (): any[] => {
```

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.ts
```diff
@@ -18,10 +19,16 @@ export class ZonedDateTimeComponent extends AbstractDatetimeComponent {
     super();
   }
 
+  // if we set the timezone to null or undefined, we want to remove the explicit timezone
+  // by setting the explicit timezone to the implicit timezone
```
atruskie: that's not an accurate comment is it? because the code sets the explicit tz to null, not to the implicit
```suggestion
  // which will result in any implicit time zone taking effect
```

src/app/components/shared/datetime-formats/duration/duration.component.spec.ts
```diff
@@ -19,42 +20,44 @@ interface TestCase {
...
+
+  test("negative explicit sexagesimal", { minutes: -10, seconds: -1.5 }, "-00:10:01.500", "-00:10:01.500 (-PT10M1.5S)", "-PT10M1.5S", false, false, true),
+
+  test("iso 8601", { minutes: 10, seconds: 1.5 }, "PT10M1.5S", "00:10:01.500 (PT10M1.5S)", "PT10M1.5S", true),
+
+  test("negative iso 8601", { minutes: -10, seconds: -1.5 }, "-PT10M1.5S", "-00:10:01.500 (-PT10M1.5S)", "-PT10M1.5S", true),
+
+  test("human readable", { minutes: 10, seconds: 1.5 }, "10 minutes 2 seconds", "00:10:01.500 (PT10M1.5S)", "PT10M1.5S", false, true),
+
+  test("negative human readable", { minutes: -10, seconds: -1.5 }, "-10 minutes 2 seconds", "-00:10:01.500 (-PT10M1.5S)", "-PT10M1.5S", false, true),
```
atruskie, line 62: we should add a test to make sure duration is rebalanced. e.g. `{  minutes: 300, seconds: 3600 } ` should show `06:00:00`

src/app/components/shared/datetime-formats/time-since/time-since.component.spec.ts
```diff
@@ -16,116 +18,116 @@ interface TestCase {
...
-    name: "JavaScript Date",
-    value: new Date("2019-11-13T14:17:58.500Z"),
-    expectedText: "1 month 3 weeks ago",
-    expectedTooltip: "2019-11-13 14:17:58.500 +00:00 Europe/London",
-    expectedDateTimeAttribute: "P-1M-2W-6DT-9H-42M-1.5S",
-  },
+  test("Luxon Duration", Duration.fromObject({ hours: 4, minutes: 42, seconds: 29, milliseconds: 500, }), "4 hours 42 minutes from now", "2020-01-01 04:42:29.500 +09:30 Australia/Darwin", "PT4H42M29.5S"),
+
+  test("Negative Luxon Duration", Duration.fromObject({ hours: -2, minutes: -10 }), "2 hours 10 minutes ago", "2019-12-31 21:50:00.000 +09:30 Australia/Darwin", "-PT2H10M"),
+
```
atruskie: we should add a rebalance test in this set as well

src/app/components/shared/datetime-formats/time-since/time-since.component.spec.ts
```diff
@@ -16,116 +18,116 @@ interface TestCase {
...
+  return {
+    name,
+    value,
+    expectedText,
+    expectedTooltip,
+    expectedDateTimeAttribute,
+  };
+}
 
 const testCases: TestCase[] = [
```
atruskie: these test definitions could do with being moved into the suite as well, because we the reader would see the clock mock definition first and understand what "now" is being fixed at

src/app/components/shared/datetime-formats/time-since/time-since.component.spec.ts
```diff
@@ -16,116 +18,116 @@ interface TestCase {
...
+      updateSpy.calls.reset();
+
+      expect(timeElement()).toHaveExactTrimmedText("30 seconds ago");
+
+      tick(secondInMilliseconds);
+      update();
+
+      expect(updateSpy).toHaveBeenCalledTimes(1);
+      expect(timeElement()).toHaveExactTrimmedText("31 seconds ago");
+    }));
```
atruskie: great test, but you should extend it to also test what happens after a minute

e.g. updates at 59 seconds to 1 minute ago

then does not update per second

but will update after a minute has passed

src/app/components/shared/datetime-formats/time-since/time-since.component.ts
```diff
@@ -36,7 +36,7 @@ export class TimeSinceComponent
       // we want to update the "time since" every second if the time is less than a minute
```
atruskie, line 38: i thought of a better algorithm for this:

- on subscribe, store the lastChange date in a private field in the component
- on each tick
    - let now be Date.now()
    - let valueDelta be the difference between now and value
    - let lastChangeDelta be the difference between now and lastChange
    - compare the absolute delta and decide if we're lessThanAMinute or not
    - if we're lessThanAMinute
        - then update()
        - and set lastChange to now
    - else if lastChangeDelta >= 60
        - then update()
        - and set lastChange to now
    - else 
        -  // do nothing

atruskie (approved):
Great work!

A few more comments and test cases to cover, but we're pretty much done

src/app/components/shared/datetime-formats/datetime/zoned-datetime/zoned-datetime.component.spec.ts
hudson-newey: One of these tests is supposed to be testing the "DateTime with offset and date attribute". I've adjusted the tests accordingly

src/app/components/shared/datetime-formats/time-since/time-since.component.ts
hudson-newey: Slightly simplified implementation that consolidates `update()` calls

image

Where `update()` modifieds `this.lastChange`

atruskie: it's clean, but the reason we pinned date.now is because we will see very minor drift between when we calculate lastChange and when we set it.

PR #2105, opened 2024-01-18
Display warning message for "bad" website status

### LINKED ISSUE #2102: Proactively display warning message for "bad" website status
In a recent pull request, we added the ability to view the website status / API status under the `/status` route.

However, at the moment, the website status isn't exposed to the user unless they access this route.

Instead, we should proactively show warning messages if a feature the user is about is use is not healthy.

### Suggested changes

- We should add a warning icon (with --baw-warning colours) in the menu bar (next to the user icon) if the overall website status is not "good". Hovering over this icon should show a tooltip, and clicking on it should take the user to the `/status` page
- We should add banner alerts to the recording list / download pages if the storage is not healthy
- We should add banner alerts to the harvest list & details pages if the user uploads or database are not healthy

Relates to: #2100

hudson-newey:
# Display warning message for "bad" website status

At the moment, when the server is down / has partial services, the only way users can diagnose the issue as a server side issue is by going to [https://www.ecosounds.org/status](https://www.ecosounds.org/status).

Because this is unintuitive to some users, we should surface the issue sooner as a server issue by adding warnings showing that the service the user is trying to use is non-functional.

**Important note**: I am not displaying an error indicator in the navbar for a timeout error.

This is because:

- We are already displaying a toast message that covers timeout messages
- If the user doesn't have internet, it tells them that they do not have internet and to check their internet / VPN connection
- The warning indicator is supposed portray the message "we are not fully functional right now, and it is not a problem with your internet / client. Try again later". Adding a timeout error case to this icon will change the semantic meaning of this icon to "something failed. It might be you; it might be us."

## Changes

- Add website status indicator to the navigational bar (that only shows when the website status is "bad" / unhealthy)
- Add warning banner to the harvest list & detail pages if the api sftp server is down
- Add warning banner to the audio recording list & download pages if the api storage is down
- Add tests for all the above
- Menu links now always emit a `<hr>` at the base of the menu links container (even when there are no widgets)

## Problems

None

## Issues

Fixes: #2102 

## Visual Changes

image
_API Status indicator icon and tooltip (will only be shown if the server status is `"bad"`)_

---

image
_Warning widget on the harvest list page_

image
_Warning widget on the harvest page_

image
_Warning widget on the recording list page_

image
_Warning widget on the recording list page_

image
_Warning widget on the batch download page_

image
_Warning widget on the audio recording details page_

image
_New `<hr>` element at the bottom of menu links without widgets_

## Final Checklist

### DIFF
--- a/src/app/components/audio-recordings/pages/details/details.component.ts
+++ b/src/app/components/audio-recordings/pages/details/details.component.ts
@@ -22,6 +22,8 @@ import { Project } from "@models/Project";
 import { Region } from "@models/Region";
 import { Site } from "@models/Site";
 import { List } from "immutable";
+import { WidgetMenuItem } from "@menu/widgetItem";
+import { WebsiteStatusWarningComponent } from "@menu/website-status-warning/website-status-warning.component";
 import schema from "./audio-recording.schema.json";
 
 const audioRecordingKey = "audioRecording";
@@ -78,7 +80,17 @@ function getPageInfo(
   return {
     pageRoute: audioRecordingMenuItems.details[subRoute],
     category: audioRecordingsCategory,
-    menus: { actions: List([downloadAudioRecordingMenuItem, downloadAudioRecordingAnalysesMenuItem]) },
+    menus: {
+      actions: List([downloadAudioRecordingMenuItem, downloadAudioRecordingAnalysesMenuItem]),
+      actionWidgets: List([
+        new WidgetMenuItem(WebsiteStatusWarningComponent, undefined, {
+          message: `
+            Downloading audio is temporarily unavailable.
+            Please try again later.
+          `,
+        }),
+      ]),
+    },
     resolvers: {
       [audioRecordingKey]: audioRecordingResolvers.show,
       [projectKey]: projectResolvers.showOptional,
--- a/src/app/components/audio-recordings/pages/download/download.component.ts
+++ b/src/app/components/audio-recordings/pages/download/download.component.ts
@@ -9,16 +9,22 @@ import { regionResolvers } from "@baw-api/region/regions.service";
 import { ResolvedModelList, retrieveResolvers } from "@baw-api/resolver-common";
 import { siteResolvers } from "@baw-api/site/sites.service";
 import { contactUsMenuItem } from "@components/about/about.menus";
-import { audioRecordingMenuItems, audioRecordingsCategory } from "@components/audio-recordings/audio-recording.menus";
+import {
+  audioRecordingMenuItems,
+  audioRecordingsCategory,
+} from "@components/audio-recordings/audio-recording.menus";
 import { myAccountMenuItem } from "@components/profile/profile.menus";
 import { PageComponent } from "@helpers/page/pageComponent";
 import { IPageInfo } from "@helpers/page/pageInfo";
 import { StrongRoute } from "@interfaces/strongRoute";
+import { WebsiteStatusWarningComponent } from "@menu/website-status-warning/website-status-warning.component";
+import { WidgetMenuItem } from "@menu/widgetItem";
 import { AudioRecording } from "@models/AudioRecording";
 import { Project } from "@models/Project";
 import { Region } from "@models/Region";
 import { Site } from "@models/Site";
 import { NgbDate } from "@ng-bootstrap/ng-bootstrap";
+import { List } from "immutable";
 import { BehaviorSubject, takeUntil } from "rxjs";
 import { loginMenuItem } from "src/app/components/security/security.menus";
 
@@ -33,7 +39,8 @@ const siteKey = "site";
 class DownloadAudioRecordingsComponent extends PageComponent implements OnInit {
   @ViewChild(NgForm) public form: NgForm;
 
-  public filters$: BehaviorSubject<Filters<AudioRecording>> = new BehaviorSubject({});
+  public filters$: BehaviorSubject<Filters<AudioRecording>> =
+    new BehaviorSubject({});
 
   public contactUs = contactUsMenuItem;
   public href = "";
@@ -48,7 +55,7 @@ class DownloadAudioRecordingsComponent extends PageComponent implements OnInit {
   public constructor(
     public session: BawSessionService,
     private route: ActivatedRoute,
-    private recordingsApi: AudioRecordingsService,
+    private recordingsApi: AudioRecordingsService
   ) {
     super();
   }
@@ -92,8 +99,18 @@ function getPageInfo(
   subRoute: keyof typeof audioRecordingMenuItems.batch
 ): IPageInfo {
   return {
-    pageRoute: audioRecordingMenuItems.batch[subRoute],
     category: audioRecordingsCategory,
+    pageRoute: audioRecordingMenuItems.batch[subRoute],
+    menus: {
+      actionWidgets: List([
+        new WidgetMenuItem(WebsiteStatusWarningComponent, () => true, {
+          message: `
+            Downloading audio is temporarily unavailable.
+            Please try again later.
+          `,
+        }),
+      ]),
+    },
     resolvers: {
       [projectKey]: projectResolvers.show,
       [regionKey]: regionResolvers.showOptional,
--- a/src/app/components/audio-recordings/pages/list/list.component.ts
+++ b/src/app/components/audio-recordings/pages/list/list.component.ts
@@ -14,6 +14,8 @@ import { filterModel } from "@helpers/filters/filters";
 import { IPageInfo } from "@helpers/page/pageInfo";
 import { PagedTableTemplate } from "@helpers/tableTemplate/pagedTableTemplate";
 import { toRelative } from "@interfaces/apiInterfaces";
+import { WebsiteStatusWarningComponent } from "@menu/website-status-warning/website-status-warning.component";
+import { WidgetMenuItem } from "@menu/widgetItem";
 import { AudioRecording } from "@models/AudioRecording";
 import { Project } from "@models/Project";
 import { Region } from "@models/Region";
@@ -176,7 +178,17 @@ function getPageInfo(subRoute: keyof typeof menuItems): IPageInfo {
   return {
     category: audioRecordingsCategory,
     pageRoute: menuItems[subRoute],
-    menus: { actions: List([downloadMenuItems[subRoute], visualizeMenuItem]) },
+    menus: {
+      actions: List([downloadMenuItems[subRoute], visualizeMenuItem]),
+      actionWidgets: List([
+        new WidgetMenuItem(WebsiteStatusWarningComponent, undefined, {
+          message: `
+            Downloading and playing audio is temporarily unavailable.
+            Please try again later.
+          `,
+        }),
+      ]),
+    },
     resolvers: {
       [projectKey]: projectResolvers.showOptional,
       [regionKey]: regionResolvers.showOptional,
--- a/src/app/components/harvest/pages/details/details.component.ts
+++ b/src/app/components/harvest/pages/details/details.component.ts
@@ -13,7 +13,11 @@ import { withUnsubscribe } from "@helpers/unsubscribe/unsubscribe";
 import { Harvest } from "@models/Harvest";
 import { Project } from "@models/Project";
 import { List } from "immutable";
-import { permissionsWidgetMenuItem } from "@menu/widget.menus";
+import {
+  permissionsWidgetMenuItem,
+} from "@menu/widget.menus";
+import { WidgetMenuItem } from "@menu/widgetItem";
+import { WebsiteStatusWarningComponent } from "@menu/website-status-warning/website-status-warning.component";
 import { HarvestStagesService } from "../../services/harvest-stages.service";
 import { harvestsMenuItemActions } from "../list/list.component";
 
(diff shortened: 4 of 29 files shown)


src/app/components/audio-recordings/pages/download/download.component.html
```diff
@@ -1,5 +1,10 @@
 <h1>Batch Download Audio Recordings</h1>
 
+<baw-website-status-warning
+  feature="isStorageHealthy"
+  websiteCapabilityText="download audio"
+></baw-website-status-warning>
+
```
atruskie: i would have thought we would put this in the action menu?

atruskie: applies to other placements too

src/app/components/shared/menu/website-status-indicator/website-status-indicator.component.ts
```diff
@@ -0,0 +1,44 @@
...
+@Component({
+  selector: "baw-website-status-indicator",
+  template: `
+    <a
+      *ngIf="!isWebsiteHealthy"
+      class="block nav-link text-danger text-nowrap"
+      [strongRoute]="websiteStatusRoute"
+      [ngbTooltip]="
+        'The website is currently experiencing a temporary loss of services. ' +
+        'Some functionality may be reduced.'
```
atruskie:
```suggestion
        '{{ brandName }} is experiencing a temporary loss of services. ' +
        'Some parts of the website may not work.'
```

(not correct syntax)

src/app/components/shared/menu/website-status-indicator/website-status-indicator.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  protected isWebsiteHealthy = true;
+  protected websiteStatusRoute: StrongRoute = websiteStatusMenuItem.route;
+
+  public ngOnInit(): void {
+    this.api
+      .show()
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe((status: WebsiteStatus) => {
+        this.isWebsiteHealthy = status.isStatusHealthy;
+      });
```
atruskie: Does this component get reloaded on router navigation? I don't think it does right, since it is outside the router outlet.

Otherwise if the api enters a degraded state while they're on the page then this won't ever update.

Probably need an observeable in the service that requests a new status object every 30 seconds or so.

src/app/components/shared/website-capability-warning/website-status-warning.component.ts
```diff
@@ -0,0 +1,42 @@
...
+
+@Component({
+  selector: "baw-website-status-warning",
+  template: `
+    <div *ngIf="!isCapabilityHealthy" class="alert alert-danger">
+      The ability to {{ websiteCapabilityText }} is temporarily unavailable.
+      Please try again later.
+
+      <div>
+        <a [strongRoute]="websiteStatusRoute">More Information</a>
```
atruskie: probably should be a `<small>`

src/app/components/shared/website-capability-warning/website-status-warning.component.ts
```diff
@@ -0,0 +1,42 @@
...
+import { withUnsubscribe } from "@helpers/unsubscribe/unsubscribe";
+import { WebsiteStatus } from "@models/WebsiteStatus";
+import { takeUntil } from "rxjs";
+
+@Component({
+  selector: "baw-website-status-warning",
+  template: `
+    <div *ngIf="!isCapabilityHealthy" class="alert alert-danger">
+      The ability to {{ websiteCapabilityText }} is temporarily unavailable.
+      Please try again later.
```
atruskie: This should support a slot. e.g. usage should be 

```html
<baw-website-status-warning>
    The upload functionis currently unavailable.
</baw-website-status-warning>
```

You can keep the attribute too - much in the same way that ngbTooltip can accept either text in an attribute or a slot - but it's always smart to add the slot as an escape hatch in case we need to put some other content in.

src/app/components/shared/website-capability-warning/website-status-warning.component.ts
```diff
@@ -0,0 +1,42 @@
...
+  protected isCapabilityHealthy = true;
+  protected websiteStatusRoute = websiteStatusMenuItem.route;
+
+  public ngOnInit(): void {
+    this.api
+      .show()
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe((status: WebsiteStatus) => {
+        this.isCapabilityHealthy = status[this.feature] as boolean;
+      });
```
atruskie: potentially the same point about updates as with the indicator, but it's less important here since this component is reloaded on navigation. Still a user could be on a page for a while ...

src/app/components/shared/website-capability-warning/website-status-warning.component.ts
```diff
@@ -0,0 +1,42 @@
...
+})
+export class WebsiteStatusWarningComponent
+  extends withUnsubscribe()
+  implements OnInit
+{
+  public constructor(private api: WebsiteStatusService) {
+    super();
+  }
+
+  @Input({ required: true }) public feature: keyof WebsiteStatus;
```
atruskie: I'm not wholly convinced this is needed.

Anywhere where it matters enough to put this warning, any service not being available is enough for it not to work. 

If disk, or database are down then uploads, downloads, and everything is non-functional. 

Then again, testing for the status of the upload server is relevant to only the harvest pages. 

Ok I argued myself in favour of this as it is, as you were

atruskie (changes requested):
Looks great, a couple of tweaks and this should be good to go

src/app/components/shared/website-capability-warning/website-status-warning.component.ts
hudson-newey: For these warning banners, I can extend the interval to 2 minutes.

Additionally, because we are fetching the status every 30 seconds, we only need to invalidate the cache on the status icon request. The warning banner should be able to use the memoized value of the status icon request (meaning that we are not making multiple requests every 2 minutes).

src/app/components/audio-recordings/pages/download/download.component.html
hudson-newey: Done.

This was accomplished by adding an `options` property to the WidgetMenuItem interface.

src/app/components/audio-recordings/pages/details/details.component.ts
```diff
@@ -78,7 +80,17 @@ function getPageInfo(
   return {
     pageRoute: audioRecordingMenuItems.details[subRoute],
     category: audioRecordingsCategory,
-    menus: { actions: List([downloadAudioRecordingMenuItem, downloadAudioRecordingAnalysesMenuItem]) },
+    menus: {
+      actions: List([downloadAudioRecordingMenuItem, downloadAudioRecordingAnalysesMenuItem]),
+      actionWidgets: List([
+        new WidgetMenuItem(WebsiteStatusWarningComponent, undefined, {
+          message: `
+            The ability to download audio is temporarily unavailable.
```
atruskie:
```suggestion
            Downloading audio is temporarily unavailable.
```

src/app/components/audio-recordings/pages/download/download.component.ts
```diff
@@ -92,8 +100,18 @@ function getPageInfo(
...
 ): IPageInfo {
   return {
-    pageRoute: audioRecordingMenuItems.batch[subRoute],
     category: audioRecordingsCategory,
+    pageRoute: audioRecordingMenuItems.batch[subRoute],
+    menus: {
+      actionWidgets: List([
+        new WidgetMenuItem(WebsiteStatusWarningComponent, () => true, {
+          message: `
+            The ability to download audio is temporarily unavailable.
```
atruskie:
```suggestion
            Downloading audio is temporarily unavailable.
```

src/app/components/audio-recordings/pages/list/list.component.ts
```diff
@@ -176,7 +190,17 @@ function getPageInfo(subRoute: keyof typeof menuItems): IPageInfo {
   return {
     category: audioRecordingsCategory,
     pageRoute: menuItems[subRoute],
-    menus: { actions: List([downloadMenuItems[subRoute], visualizeMenuItem]) },
+    menus: {
+      actions: List([downloadMenuItems[subRoute], visualizeMenuItem]),
+      actionWidgets: List([
+        new WidgetMenuItem(WebsiteStatusWarningComponent, undefined, {
+          message: `
+            The ability to download audio is temporarily unavailable.
```
atruskie:
```suggestion
            Downloading and playing audio is temporarily unavailable.
```

src/app/components/harvest/pages/details/details.component.ts
```diff
@@ -56,7 +60,14 @@ DetailsComponent.linkToRoute({
     actions: List([...harvestsMenuItemActions, newSiteMenuItem]),
     actionWidgets: List([
       harvestValidationsWidgetMenuItem,
-      permissionsWidgetMenuItem
+      permissionsWidgetMenuItem,
+      new WidgetMenuItem(WebsiteStatusWarningComponent, undefined, {
+        feature: "isUploadingHealthy",
+        message: `
+          The ability to upload audio is temporarily unavailable.
```
atruskie:
```suggestion
         Uploading is temporarily unavailable.
```

src/app/components/harvest/pages/list/list.component.ts
```diff
@@ -119,6 +121,15 @@ ListComponent.linkToRoute({
   category: harvestsCategory,
   menus: {
     actions: List(harvestsMenuItemActions),
+    actionWidgets: List([
+      new WidgetMenuItem(WebsiteStatusWarningComponent, undefined, {
+        feature: "isUploadingHealthy",
+        message: `
+          The ability to upload audio is temporarily unavailable.
```
atruskie:
```suggestion
          Uploading is temporarily unavailable.
```

src/app/services/baw-api/website-status/website-status.service.ts
```diff
@@ -1,48 +1,74 @@
...
-    Accept: "application/json",
-  });
+  private updateStatus(): void {
+    this.show()
+      .pipe(first())
+      .subscribe((status) => {
+        if (this.status$.value !== status) {
+          this.status$.next(status);
+        }
+      });
```
atruskie: this feels very messy.

For a start, the reference check, is that useless? `status` from `show()` will always be a new object.
Also, what function does `first()` serve? The API only gives back one result

I wonder if this can be written more compactly.


```
 status$ = tick$.pipe(map(this.show), distinct(), shareReplay(1));
```

atruskie (approved):
Noticing lots of line break churn. Did our line length settings change?

hudson-newey:
> Noticing lots of line break churn. Did our line length settings change?

Our line length settings have not changed. My best guess is that the formatter hasn't run on these files before

src/app/services/baw-api/website-status/website-status.service.ts
hudson-newey: Good idea

I've modified it a bit (we use `switchMap` instead of `map` because `this.show` returns an observable)

and I need to set an initial value of `status$`, otherwise we won't know the status until the interval fires its first event (after 30 seconds).

image

atruskie: For initial trigger you could just do $tick.next?

Also is the wrapper lambda really needed for the show invocation?

hudson-newey: > For initial trigger you could just do $tick.next?

it's not a behaviour subject, so we can't directly invoke `.next()` on it

image

We could if we had access to the subscriber [https://rxjs.dev/guide/observable](https://rxjs.dev/guide/observable), but we won't have access to it without some hacky and sub-par code.

I beleive this is the most efficiant and concise way.

---

> Also is the wrapper lambda really needed for the show invocation?

Kind of:

By using `switchMap(() => this.show())` I'm actually invoking a callback that has the signiture `(interval_time_elapsed: number) => Observable<any>`.

`this.show` used to take `cachingOptions` as it's first parameter meaning it had the signiture `(cache_options: BawCacheOptions) => Observable<WebsiteStatus>` meaning that if I implicitly called it like `switchMap(this.show)`, the `cache_options` would be set to the amount of time elapsed in the interval eg. the number 5 (not good).

However, because I want us to use the `status$` observable for this service, and not the `show()` method, I think it's okay to remove cache options override from `this.show()` since `this.show` is a private method, and we never override our caching options in this service.

Additionally our caching options are already defined by our abstract api (these caching options were to override our defaults).

hudson-newey: > For initial trigger you could just do $tick.next?

**Correction:**

By doing the initial call by assigning `this.status$` to `this.show()`, it actually destroyed our binding to `this.tick$`. (obvious in hindsight when you look at it)

I did some looking around, it looks like we can pipe the interval through `startsWith(0)`. This means that when the interval is initially created, it emits an event.

No need for explicit initial event invokation!

> Also is the wrapper lambda really needed for the show invocation?

Yes. Because otherwise `this.show` will be called from the rxjs context, and `this` will not refer to the service context. (we need to use the lambda as a closure)

---

image
_Note: This is my testing code where it triggers every 3 seconds so I don't have to wait 30 seconds during testing_

PR #2110, opened 2024-01-30
Upgrade to Angular 17

### LINKED ISSUE #2109: Upgrade to Angular 17
Angular 17 has just come out. It boasts significant performance improvements, first class support for SSR, and signals.

We should upgrade to take advantage of these benefits by upgrading to angular 17.

hudson-newey:
# Upgrade to Angular 17

Angular 17 is out and supports signals, a significant performance and build time boost!

## Changes

- Upgrade angular and its dependencies
- Some bootstrap variables (`--baw-highlight`) are now defined in the `themes.service.ts` during runtime instead of during scss compile time
- Remove `npx ngcc` from GitHub workflow as it is no longer needed
- bootstrap `.card` class now defines a text `color`. Because some pages previously relied on bootstrap `.card` inheriting colour, I've moved some color definitions down the DOM style tree (so that text colour is explicitly set on `card` components if it was relying on colour inheritance)
- Remove angular-universal (as SSR is now natively supported)
- Added `REQUEST` and `RESPONSE` injector tokens so that SSR has access to cookies

**Important!**

- By updating our IANA timezone formatter (vvo/tzdb), it changed UTC timezone formatting from `UTC` to `Etc/UTC` [See the `annotation-download.component.spec.ts`](https://github.com/QutEcoacoustics/workbench-client/pull/2110/files#diff-97219b00cb9356a7474e19a6361028698f7037c144c988f56996f4f11cf0b46eR161) tests change

## Problems

None

## Issues

Fixes: #2109 

## Visual Changes

Bootstrap changed how card borders and input group text looks

image
_New look_

image
_Old Bootstrap card look_

## Final Checklist


scripts/ssr-tests.ps1
```diff
@@ -97,7 +97,7 @@ Assert-Contains "<h1>Not Found<\/h1>" $brokenFile
 # Tests for angular production build
 Format-Info "Retrieve angular bundle"
 $bundleFile = "main.js"
-Get-Page $((Select-String -Pattern "main\..+?\.js" -Path $indexFile).Matches.Value) $bundleFile
+Get-Page $((Select-String -Pattern "main\-.+?\.js" -Path $indexFile).Matches.Value) $bundleFile
```
atruskie:
```suggestion
Get-Page $((Select-String -Pattern "main-.+?\.js" -Path $indexFile).Matches.Value) $bundleFile
```

src/app/components/shared/annotation-download/annotation-download.component.spec.ts
```diff
@@ -158,7 +158,7 @@ describe("AnnotationDownloadComponent", () => {
     it("should default to UTC if no timezone exists for site", () => {
       setup([projectKey, siteKey], [defaultProject, siteWithoutTimezone]);
       spec.detectChanges();
-      expect(spec.component.model.timezone).toBe("UTC");
+      expect(spec.component.model.timezone).toBe("Etc/UTC");
```
atruskie, line 161: make sure there is a test elsewhere in this file that tests the presentation value of the timezone remains unchanged.

We don't use the slash style timezones in the presented text, so hopefully this change won't even show to a user.

also make sure that the network request to the server is well understood. Did you try downloading an annotation files with `Etc/UTC` is the HTTP request parameters?

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +104,37 @@ export class ThemeService {
     }
   }
 
+  private overrideBootstrapColors(color: string) {
+    const variables: [string, number][] = [
+      ["--bs-link-color", 0],
+      ["--bs-link-hover-color", -0.1],
+      ["--bs-link-disabled-color", 0.1],
```
atruskie: this was just a concept yeah? We didn't actually need to overwrite the `bs-link-disabled-color` property?

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +104,37 @@ export class ThemeService {
     }
   }
 
+  private overrideBootstrapColors(color: string) {
+    const variables: [string, number][] = [
```
atruskie: factor out to a class level constant

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +104,37 @@ export class ThemeService {
     }
   }
 
+  private overrideBootstrapColors(color: string) {
```
atruskie, line 120: document function and write some some notes about the various blockers we have to deal with as bootstrap incorporates css variables (and which color function we're waiting for broad support for!)

src/app/services/theme/theme.service.ts
```diff
@@ -62,6 +62,8 @@ export class ThemeService {
     for (const themeColor of Object.keys(this.theme)) {
       this.setColor(themeColor as ThemeColor, this.theme[themeColor]);
     }
+
+    this.overrideBootstrapColors(this.theme.highlight);
```
atruskie, line 73: do we need to special case the SSR stuff? 

Or are the visual differences too small to worry about. Leave a note if we're intentionall ignoring.

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +104,37 @@ export class ThemeService {
...
+    const variables: [string, number][] = [
+      ["--bs-link-color", 0],
+      ["--bs-link-hover-color", -0.1],
+      ["--bs-link-disabled-color", 0.1],
+    ];
+
+    const targetColor = hsl(color);
+
+    variables.forEach(([variable, lightnessAdjustment]) => {
+      const hslColor = targetColor.copy({ l: targetColor.l + lightnessAdjustment });
```
atruskie: double check but make sure the `l` value is clamped to `[0.0, 1.0]` - the d3 function is almost definitely doing this

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +104,37 @@ export class ThemeService {
...
+      const rgbColor = rgb(hslColor);
+
+      // set the color in the browser
+      this.root.style.setProperty(variable, rgbColor.toString());
+      this.root.style.setProperty(
```
atruskie:
```suggestion
      // bootstrap seems to double define css color variables, one with the RGB color (as per above)
      // and one as an array of r, g, and b, values which can be inserted into other contructs
      this.root.style.setProperty(
```

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +104,37 @@ export class ThemeService {
...
+      );
+
+      // set the color in SSR
+      const style = this.document.createElement("style");
+      style.innerHTML = `
+      :root {
+        ${variable}: ${rgbColor.toString()};
+        ${variable}-rgb: ${rgbColor.r}, ${rgbColor.g}, ${rgbColor.b};
+      }
+    `;
```
atruskie: ok ignore my special case comment from above

but we should split the SSR and the client versions into their own functions:

- for consistency (with the other pair of set color functions in this file)
- and to makes ure SSR is not attempting to access dom elements (this may not be an issue, but since I believe it results in no effect, we shouldn't just leave it in - it is confusing having code that does nothing and it is extra work)

src/app/styles/bootstrap/_variables.scss
atruskie, line 1: did you confirm the pagination color is overridden still with the new changes?

src/app/styles/bootstrap/components/_links.scss
atruskie, line 1: ditto, double check this isn't actually needed

atruskie (changes requested):
Looks great!

src/app/components/shared/annotation-download/annotation-download.component.spec.ts
hudson-newey: Presentation tests: Complete

`Etc/UTC` is sent in requests when downloading annotation files

---

_Edit: I fixed the request bug by modifying the annotation download component [here](https://github.com/QutEcoacoustics/workbench-client/pull/2110/files#diff-c8fda3fbb6bdcfa728d109ac1d329057077e940900fbb5c689b8521321f775c2R119-R122)_

src/app/styles/bootstrap/_variables.scss
hudson-newey: I've visually confirmed

Additionally, [looking at the Bootstrap source](https://github.com/twbs/bootstrap/blob/8cc86cc5c5bc4f4b9c3a928b8fb7a4648c5bb5bc/scss/_variables.scss#L1301), we can see that $pagination-color is `--bs-link-color` (which we are overriding in [`theme.service.ts`](https://github.com/QutEcoacoustics/workbench-client/pull/2110/files/4fd49f76947b52f52d610d8b26f2cc8ea8225a20#diff-f1023b76e7649fc57179aafc3bd2a7bc3e0e7a59bf9a1688258b0fea094bcf02R109))

src/app/services/theme/theme.service.ts
hudson-newey: hsl lightness is _kind of_ clamped to [0,1] [d3 tests](https://github.com/d3/d3-color/blob/71c7f100f9fa85a1c70fcbaeb5f803ee8db5620d/test/hsl-test.js#L159-L164).

By the looks of it, it creates an invalid colour if brightness is outside the acceptable range (I'd expect it to round down to 1 or round up to 0)

src/app/services/theme/theme.service.ts
```diff
@@ -23,7 +23,13 @@ const themeVariants = [
   "-darker",
   "-darkest",
 ] as const;
-export type ThemeVariant = typeof themeVariants[number];
+export type ThemeVariant = (typeof themeVariants)[number];
+
+const bootstrapVariables: [string, number][] = [
+  ["--bs-link-color", 0],
+  ["--bs-link-hover-color", -0.1],
```
atruskie: i'd define the values with constants or a hash. Darkest, darker, normal, lighter, lightest - as we do in our css. Link back to those other definitions

src/app/services/theme/theme.service.ts
```diff
@@ -102,6 +110,42 @@ export class ThemeService {
...
+    bootstrapVariables.forEach(([variable, lightnessAdjustment]) => {
+      const hslColor = targetColor
+        .copy({ l: targetColor.l + lightnessAdjustment })
+        .clamp();
+
+      const rgbColor = rgb(hslColor);
+
+      if (!this.isServer) {
+        this.setBootstrapColorInBrowser(variable, rgbColor);
+      }
```
atruskie: explain why we're not doing this in SSR?

atruskie (approved):

src/app/services/theme/theme.service.ts
hudson-newey: I was under the impression that we didn't have access to the DOM in SSR

However, I investigated a bit further, and it appears that SSR has access to the DOM object model. Therefore, we can set these values server side

PR #2115, opened 2024-02-06
Add signals to date/time components

### LINKED ISSUE #2114: Migrate date/time components to use ng17 signals
Given that the date/time components will be present multiple times on a single page, we should use signals instead of zone.js reactivity.

As part of this issue, we should:

- Modify uses of RxJS in the date/time component to use signals
- Modify uses of old angular `Input` and `Output` (that use decorators, I just didn't want to tag the people who have these usernames)

hudson-newey:
# Add signals to date/time components

With the addition of signals into Angular, we can improve the change detection of the date/time components to be more efficient.

This PR changes the date/time component to use signals over `zone.js`

## Changes

- Upgrade from angular `17.0.9` -> `17.1.2`
- Add signals to date/time components
- Change date/time component tests to use the Angular `TestBed` instead of the ng-neat testbed. This is because (at the moment of writing) ng-neat `setInput` doesn't correctly handle signal inputs
	- Some assertions are now changed from their ng-neat assertions to their angular equivalents (eg. `toHaveExtactTrimmedText` is now `expect(foo.textContent.trim()).toBe("xyz")`)
- `time-since.component.ts` now uses a variable `now` instead of `DateTime.now()` to prevent time drift within `update()` function
- the `datetime-example.component.ts` now uses a property for the `fakeDateTime` that sets the zone in advance

## Problems

We still have to implement the date/time components on relevant pages

## Issues

Fixes: #2114 

## Visual Changes

None

## Final Checklist


src/app/components/admin/datetime-example/datetime-example.component.ts
atruskie, line 1: this file changing defeats the point: we can't adopt signals if the abstraction leaks outside the component. Makes it a cancer, this file and the template must be reverted or we close this PR

atruskie (changes requested):

src/app/components/admin/datetime-example/datetime-example.component.ts
hudson-newey: As discussed offline, signals have been removed from this file.

However, there is a change (that we discussed) to change `fakeDate` to a property that has it's zone set when the implicit timezone or datetime changes. (removing the referencing equality bug)

src/app/components/shared/datetime-formats/datetime/abstract-datetime.component.ts
```diff
@@ -40,9 +46,9 @@ export abstract class AbstractDatetimeComponent extends AbstractTemplateComponen
 
   public update(): void {
     const zone = this.extractTimezone();
-    const valueInZone = this.value.setZone(zone);
+    const valueInZone = this.value().setZone(zone);
 
-    const timezoneName = AbstractDatetimeComponent.formatTimezone(zone, this.value);
+    const timezoneName = AbstractDatetimeComponent.formatTimezone(zone, this.value());
```
atruskie: we should pull `value()` out once and store in a variable

src/app/components/shared/datetime-formats/datetime/abstract-datetime.component.ts
```diff
@@ -65,11 +71,11 @@ export abstract class AbstractDatetimeComponent extends AbstractTemplateComponen
...
 
   private dateTimeFormat(): string {
-    if (this.date && this.time) {
+    if (this.date() && this.time()) {
       return AbstractDatetimeComponent.FULL_DATETIME;
-    } else if (this.date) {
+    } else if (this.date()) {
       return AbstractDatetimeComponent.DATE_ONLY;
-    } else if (this.time) {
+    } else if (this.time()) {
```
atruskie: ditto, don't execute the methods more than once

src/app/components/shared/datetime-formats/duration/duration.component.ts
```diff
@@ -19,33 +25,33 @@ export class DurationComponent extends AbstractTemplateComponent<InputType, Dura
...
-      return this.value.toISO();
-    } else if (this.sexagesimal) {
-      return this.value.toFormat(DurationComponent.DURATION_SEXAGESIMAL);
+    if (this.humanized()) {
+      const relativePrefix = this.value().valueOf() < 0 ? "-" : "";
+      return relativePrefix + toRelative(this.value(), { largest: 2, round: true });
+    } else if (this.iso8601()) {
+      return this.value().toISO();
+    } else if (this.sexagesimal()) {
+      return this.value().toFormat(DurationComponent.DURATION_SEXAGESIMAL);
```
atruskie: read values once

src/app/components/shared/datetime-formats/time-since/time-since.component.ts
```diff
@@ -15,66 +21,64 @@ type InputType = Duration | DateTime | Date;
...
+        this.lastChange = now;
       }
     });
   }
 
+  private lastChange = DateTime.now();
+
   public update(): void {
-    const tooltipDate = this.value.toFormat(
+    const tooltipDate = this.value().toFormat(
```
atruskie: read value once

atruskie (approved):
looks great. 

avoid more the one invocation of a signal read per method.

src/app/components/shared/datetime-formats/datetime/abstract-datetime.component.ts
hudson-newey: Chosen variable names: `showDate` and `showTime`

I chose this as to reduce the impact of typos eg. if we used `const date`, if we forgot to type `this` we would be using the wrong variable.

Additionaly, in the previous PR, we agreed that `showDate` and `showTime` were semantically correct, however, to improve dx, we shortened the attribute names to `date` and `time`

src/app/components/shared/datetime-formats/duration/duration.component.ts
hudson-newey: For this component, I am able to read the value once in `update` and pass `value` as a parameter to the `formattedValue` and `tooltipValue` methods.

image

The alternative is reading `this.value()` three times (in `update`, `formattedValue`, tooltipValue`)

src/app/components/shared/datetime-formats/datetime/abstract-datetime.component.ts
hudson-newey: chosen variable name: local variable `const value`

usage:

```ts
const timezoneName = AbstractDatetimeComponent.formatTimezone(zone, value);
```

PR #2117, opened 2024-02-08
Implement Date/Time components

### LINKED ISSUE #345: Systematic check of dates & times, ensuring timezone formatting

### LINKED ISSUE #346: Clear up ambiguities about the timezone field in the website

hudson-newey:
# Implement Date/Time components

With the functionality of the date/time components now complete, we should implement them on the relevant pages to standardize formatting and localization.

This fixes a common and re-occuring point of user friction.

## Changes

- Remove `toRelative` pipe and associated tests
- Migrates batch download table to use `baw-zoned-datetime` and `baw-duration`
- The audio recordings list page now uses `baw-duration` for the "duration" column
- The harvest list page now uses localized `baw-datetime` for the "created" column
- The completed Harvest page now uses `baw-duration` for recording duration in the table
	- Now uses `baw-zoned-datetime` (localized to site timezone) for "recorded" column
- Annotation import pages (list & details) now use localized `baw-datetime` for "Created At" column
- "My-annotations" page now uses the `baw-time-since` component for the "last updated" column
- Event summary report now uses localized date/time component (`baw-datetime`) for generated date/time
	- Now uses `baw-duration` for search range table
- user-badge (created by xyz, updated by xyz, x time ago widget in menu bar) now uses `baw-time-since`
- The site details page "This site contains recordings from abc to xyz" has been replaced by `baw-zoned-datetime`
- The recent annotations component (found on the statistics page) now uses `baw-time-since`
- Recent audio recording components (found on the statistics page) now use the `baw-duration` component for the "Duration" column
	- Now uses the `baw-time-since` component for the "Uploaded [at]" column
- The abstract model details page (audio recording details page now checks for Duration and DateTime types and formats them correctly using the `baw-datetime` and `baw-duration` components)
	- DateTime is formatted in a localized date/time using `baw-datetime` (since this is predominantly used for "created at" fields and it retains the existing behavior)
	- Durations are formatted in ISO8601 format **I’m unsure this is correct, but it is the existing behavior. Perhaps we should use sexagesimal??**
- Add date/time test helper to standardise relative durations & date/time formats in tests
- Updated unit & integration tests for the above
- Removed the test that asserted the download table date/time tooltip was correct (this is now handled by the date/time components themselves)

### Bug fixes

- The tooltips on the Harvest statistics component is currently broken (due to infinite re-rendering on hover). This has now been fixed
- The `abtract-template.component.ts` added whitespace before and after the datetime. This lead to incorrectly formatted documents and has been fixed

## Problems

None

## Issues

Fixes: #345
Fixes: #346

### The "created by" and "updated by" shields widget (in the menu bar) now uses the two largest units (to keep consistent with the rest of the relative duration formats in the website)

image
_Old Format_

image
_New Format_

---

### The audio recordings page now shows fractional seconds

image
_Old Format_

image
_New Format_

---

### Harvest ETA component now uses the `<baw-time-since>` and `<baw-duration>` components

This means that I have had to re-order some words above the progress bar

image
_Old Format_

image
_New Format_

This also removes an additional RxJS observable (potentially improving performance)

---

### The "My Sites" page now uses the two largest units in the "Last Modified" column

image
_Old Format_

image
_New Format_

---

### Recent annotations and recent audio recordings now show the two largest units

image
_Old Format_

image
_New Format_

---

### The audio recording details page has new changed date/time formats

image
_Old Format_

image
_New Format (amended)_

---

### Unshown visual changes

- All implementation of the date/time, duration, and time-since components have their underline styling indicating that there is a tooltip on hover. See previously approved #2096 for underline styling

## Final Checklist


atruskie:
image

For this case we can probably just add a `(<baw-datetime-since ...></baw-datetime-since>)` afterwards to achieve a similar effect to what was there.

src/app/components/shared/detail-view/detail-view.component.ts
```diff
@@ -29,6 +30,8 @@ export class DetailViewComponent {
   @Input() public model: AbstractModel;
 
   public getValue(field: FormlyFieldConfig) {
-    return this.model[field.key as string];
+    // because formly fields can be numbers as well as string
+    // we use toString() so that we always convert the
```
atruskie: incomplete comment?

src/app/components/shared/detail-view/render-field/render-field.component.ts
```diff
@@ -106,6 +111,17 @@ export class RenderFieldComponent
     return this.display as ImageUrl[];
   }
 
+  // I don't like using "as" for type casing because it just disables TS compiler checks
```
atruskie:
```suggestion
  // I don't like using "as" for type casting because it just disables TS compiler checks
```

atruskie (approved):
I notice that we've got a lot of the same code in the test, e.g. 

```
   const relativeTime = toRelative(durationSince, {
          largest: 2,
          round: true,
        });
```

maybe add a test helper?

Otherwise:
ShipitSquirrelGIF

hudson-newey:
> For this case we can probably just add a `(<baw-datetime-since ...></baw-datetime-since>)` afterwards to achieve a similar effect to what was there.

My hesitation about this format was that we would have two tooltips that shouw the exact same information (one for the `baw-datetime` and one for the `baw-time-since`).

However, by adding a `hasTooltip` input signal to the `abstract-template.component`, we are able to conditionally hide the `baw-time-since` tooltip (since it give the same information as the `baw-datetime` tooltip)

image

atruskie:
> > For this case we can probably just add a `(<baw-datetime-since ...></baw-datetime-since>)` afterwards to achieve a similar effect to what was there.
> 
> My hesitation about this format was that we would have two tooltips that shouw the exact same information (one for the `baw-datetime` and one for the `baw-time-since`).
> 
> However, by adding a `hasTooltip` input signal to the `abstract-template.component`, we are able to conditionally hide the `baw-time-since` tooltip (since it give the same information as the `baw-datetime` tooltip)
> 
> image

I don't think we need to worry honestly. The tooltips add context to the text they wrap. They're not on by default. 

If we really wanted to solve this we might add another datetime control or option to show date and time-since in one component. But as it stands we have exactly one example so that would be premature optimization - as would be adding an option to hide the tooltip

atruskie (approved):

PR #2125, opened 2024-03-12
Don't send auth token when calling /status

### LINKED ISSUE #2124: Requests to API `/status` endpoint with expired tokens results in "unhealthy" status
**Describe the bug**
When API calls to the `/status` endpoint are made with an expired auth token, the API request fails and the status is reported as "unhealthy".

This can result in the website status warning icon being shown when the API is fully functional

To fix this we should not send the authentication Bearer token in `/status` endpoint requests

hudson-newey:
# Don't send auth token when calling `/status`

At the moment, the website status service contacts the `/status` endpoint with a Bearer token in the `Authentication` request headers.

This can cause the request to fail and incorrectly report an "Unhealthy" status when the website is actually healthy.

**_This PR modifies some very critical code that impacts a vast majority of the website_**

## Changes

- Adds `AUTHENTICATION_CONTEXT` token to `api.interceptor.service.ts`. This allows us to conditionally append the `Authentication` token in the interceptor (defaults to sending the auth token)
- Adds `withCredentials` to `baw-api.service.ts` `httpGet`, `httpPost`, `httpPut`, and `httpDelete`. This option takes a property `{ withCredentials: boolean }` that can be used to conditionally add the auth token to requests
- Combines all the options for baw api services into one interface called `BawServiceOptions`
- Refactors all services that use the old options config style to the new config style
- Explicitly sets caching strategy in the `website-status.service.ts` (to prevent regressions in the future where we could change the caching strategy without considering this service)
- Adds tests for the above
- Renames `options` in `baw-api.service.ts` http functions to `headers` (because they are always used in the headers, and it can now cause confusion as these methods can take option objects)
- Re-enables a majority of api interceptor tests (these are currently commented out, outdated, and non-functional)

## Problems

None

## Issues

Fixes: #2124 

## Visual Changes

None

## Final Checklist

### DIFF
--- a/src/app/services/baw-api/baw-api.service.ts
+++ b/src/app/services/baw-api/baw-api.service.ts
@@ -1,4 +1,4 @@
-import { HttpClient, HttpHeaders } from "@angular/common/http";
+import { HttpClient, HttpContext, HttpHeaders } from "@angular/common/http";
 import { Inject, Injectable, Injector } from "@angular/core";
 import { KeysOfType, Writeable, XOR } from "@helpers/advancedTypes";
 import { toSnakeCase } from "@helpers/case-converter/case-converter";
@@ -13,7 +13,6 @@ import {
   AbstractModelWithoutId,
 } from "@models/AbstractModel";
 import { HttpCacheManager, withCache } from "@ngneat/cashew";
-import { ContextOptions } from "@ngneat/cashew/lib/cache-context";
 import { withCacheLogging } from "@services/cache/cache-logging.service";
 import { CacheSettings, CACHE_SETTINGS } from "@services/cache/cache-settings";
 import { API_ROOT } from "@services/config/config.tokens";
@@ -21,17 +20,22 @@ import { ToastrService } from "ngx-toastr";
 import { Observable, iif, of, throwError } from "rxjs";
 import { catchError, concatMap, map, switchMap, tap } from "rxjs/operators";
 import { IS_SERVER_PLATFORM } from "src/app/app.helper";
+import { ContextOptions } from "@ngneat/cashew/lib/cache-context";
 import { BawSessionService } from "./baw-session.service";
+import { CREDENTIALS_CONTEXT } from "./api.interceptor.service";
 
 export const defaultApiPageSize = 25;
 export const unknownErrorCode = -1;
 
-interface NotificationsOpt {
+export interface BawServiceOptions {
+  /** If set and a request fails, an error notification won't be shown to the user */
   disableNotification?: boolean;
-}
 
-interface CacheOpt {
-  disableCache?: boolean;
+  /** If set, requests will include the users authentication token and cookies */
+  withCredentials?: boolean;
+
+  /** Allows you to modify the cashew cache options per request */
+  cacheOptions?: ContextOptions;
 }
 
 /** Default headers for API requests */
@@ -48,6 +52,11 @@ export const multiPartApiHeaders = new HttpHeaders({
   // eslint-disable-next-line @typescript-eslint/naming-convention
   Accept: "application/json",
 });
+export const defaultBawServiceOptions: Required<BawServiceOptions> = {
+  disableNotification: false,
+  withCredentials: true,
+  cacheOptions: { cache: false },
+};
 
 /**
  * Interface with BAW Server Rest API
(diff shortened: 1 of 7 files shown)


atruskie (changes requested):
Ok: General feedback on naming. I think `emitSession` is generally misleading. Prefer `send` over `emit` because it is more common vocabulary. The real issue is `session`. Sessions are usually tied to cookies... and in this case `emitSession` has no effect on the cookies. `emitSession: false` even still sends the cookie!

I agree that sending the cookie does not cause harm but for parity with the fetch API it might be simpler if we provided an option similar too - if not the same underneath - as `withCredentials` that sends neither token or cookies. That has a more general purpose utility to it.

If you really don't want to do that, then the name should be _more specific_ to reduce ambiguity. Something like `sendAuthToken` - that is explicitly clear that we're only not sending our auth token.

I'm also not a fan of the options blow out on the service methods. I'd suggest we should just have a single `options` object.

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -167,16 +187,14 @@ export class BawApiService<
   public list(
     classBuilder: ClassBuilder,
     path: string,
-    opts?: CacheOpt & NotificationsOpt
+    options: BawServiceOptions = defaultBawServiceOptions
```
atruskie: there's not much reason to set this to the default if you're just merging it in with the default one line later.

Would ` options: BawServiceOptions = {}` do instead?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -441,19 +534,36 @@ export class BawApiService<
...
+    headers: HttpHeaders = defaultApiHeaders,
+    options: BawServiceOptions = defaultBawServiceOptions
   ): Observable<ApiResponse<Model>> {
+    const fullOptions = this.buildServiceOptions(options);
+
+    const context = new HttpContext().set(
+      CREDENTIALS_CONTEXT,
+      fullOptions.withCredentials
+    );
+
```
atruskie: this is repeated  in all these methods

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -441,19 +534,36 @@ export class BawApiService<
...
+   * eg. `{ withCredentials: false }` will not include the users authentication token and cookies in the request
    */
   public httpPatch(
     path: string,
     body?: any,
-    options: any = defaultApiHeaders
+    headers: HttpHeaders = defaultApiHeaders,
+    options: BawServiceOptions = defaultBawServiceOptions
   ): Observable<ApiResponse<Model>> {
+    const fullOptions = this.buildServiceOptions(options);
```
atruskie, line 533: looks like the options are all built twice, in the create, show, list... methods and in the lower level ones. Inefficient?

atruskie (approved):

src/app/services/baw-api/baw-api.service.ts
hudson-newey: Good point, I'll make the code change

src/app/services/baw-api/baw-api.service.ts
hudson-newey: I originally did this so that we wouldn't be relying on `undefined` as falsy values for `disableNotifications` eg. [in show](https://github.com/QutEcoacoustics/workbench-client/pull/2125/files/d4ce617e055ffa8add5b0a2c1e287104fbff2cd1#diff-2e5a0799ecfae901aaf8f95aa03ccce27de670e1faf96896372e2992493d13a3R263) if the user passes in `{ cacheOptions: { cache: false } }`, `disableNotifications` will be `undefined` (defaulting to `false` because `undefined` is a falsy value).

This is currently how it works [proof](https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/services/baw-api/baw-api.service.ts#L179), but it might lead to some troubles down the road (as I believe that using `undefined` as a default "falsy" value is bad practice and might result in a foot gun)

example of potential future pitfalls: if we change the `defaultOptions.disableNotifications` to `true` sometime in the future, this change won't take effect on some API calls where we set `withCredentials` or `cacheOptions`

---

In hindsight, we could do something like this

image

Where we make nullish `options?.disableNotifications` values default to the default options value

PR #2145, opened 2024-10-04
Add Annotation Search and Verification Pages

### LINKED ISSUE #2140: Create verification Interface
Users have expressed the desire to verify annotations created by various human annotators and machine learning models.

To accomplish this, we should utilize the new Open Ecoacoustics [oe-verification-grid](https://github.com/ecoacoustics/web-components/blob/verification-component/src/components/verification-grid/verification-grid.ts) web component.

[Figma designs](https://www.figma.com/design/PVRw96Rn3JlSGaYCGKGoYe/Validation-interface?node-id=0-1&t=F7CLzdPr8TFJ7g7m-1)

hudson-newey:
# Verification Interface

This pull request creates an annotations search page & accompanying verification grid to the workbench client.

### Features

- Add open-ecoacoustics web components as a dependency
- Add annotations search page
- Add verification interface page
- Add `@angular/elements` allowing us to create custom elements using Angular components (was needed for full Lit + Angular interoperability within slotted content)
- Add semantic selectors to wip component
- Add disabled attributes to `date-time-input.component` start/end date/time inputs
- Add `pageSize` attribute to pagination template
- You can now add response headers to the SSR server using the `angular.json` `serve.options.headers` object

### Bug Fixes

- Fix the previously non-functional `UnsavedInputGuard`
- Fix docker image publish workflow doubling the inputted tag name on manual dispatches
- Fix docker version build numbers now use an incrementing number instead of the number of commits in the tag
- Fixes a bug with route titles when viewing profiles
- Fixes a bug where typeahead input component pills would not be vertically centered
- Fix failing CI due to deprecated dependencies

### Code Quality

- Disable AOT during development builds and disable lib check (don't typecheck third party library code). These changes result in decreased dev build time by ~40-50%
- Add new test helpers `selectFromTypeahead` and `getElementByInnerText`

## Issues

Fixes: #2140 

## Visual Changes

image

_Annotations search page_

---

image

_Verification grid page_

---

image

_Verification grid with context card expanded_

## Final Checklist


src/app/app.component.ts
```diff
@@ -83,6 +100,8 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
       return;
     }
 
+    import("@ecoacoustics/web-components");
```
atruskie: probably need to explain why this import is dynamic

src/app/app.component.ts
```diff
@@ -64,6 +71,16 @@ export class AppComponent extends withUnsubscribe() implements OnInit {
      */
     this.router.initialNavigation();
     globals.initialize();
+
+    if (!this.isServer) {
```
atruskie, line 78: comment

```
// register all web components here
// we make some of our standalone angular components into standards based web components
// so that they can operate entirely independently - e.g. in shadow dom
```

src/app/components/annotations/annotation.menu.ts
```diff
@@ -0,0 +1,100 @@
...
+
+function makeAnnotationSearchMenuItem(
+  subRoute: AnnotationRoute,
+  parent?: MenuRoute,
+) {
+  return menuRoute({
+    icon: ["fas", "layer-group"],
+    label: "Search Annotations",
+    tooltip: () => "(BETA) Search Annotations",
+    predicate: isLoggedInPredicate,
```
atruskie: showing annotations should be available on public anonymous access for public access projects

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -0,0 +1,94 @@
...
+    id="recordings-input"
+    label="Recordings of interest"
+    inputPlaceholder="All"
+    [searchCallback]="createIdSearchCallback(recordingsApi)"
+    [resultTemplate]="idTypeaheadTemplate"
+    [model]="searchParameters.audioRecordingModels"
+    (modelChange)="
+      updateModel('audioRecordings', getIdsFromAbstractModelArray($event))
+    "
+  ></baw-typeahead-input>
```
atruskie: ok this is an id search. this is a pretty advanced filter and not that friendly to use. Honestly most of the time if this will be used it will be activated by an external cold load from a generated url.

- move it the bottom of the search parameters (least important)
- maybe hide it in an advanced section
- and the label need to be clearer - it's an ID filter but that's not in the name

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -0,0 +1,94 @@
...
+      class="form-check-input"
+      [checked]="searchParameters.onlyUnverified"
+      (change)="updateModel('onlyUnverified', $event.target['checked'])"
+      disabled
+    />
+    <label for="filter-verified" class="form-check-label">
+      Only show un-verified annotations
+    </label>
+  </div>
+
```
atruskie, line 104: wrap in a `wip` component so it is hidden

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -0,0 +1,134 @@
...
+      ];
+    }
+
+    if (dateTimeModel.timeStartedAfter || dateTimeModel.timeFinishedBefore) {
+      this.searchParameters.time = [
+        dateTimeModel.timeStartedAfter,
+        dateTimeModel.timeFinishedBefore,
+      ];
+    }
+  }
```
atruskie: it feels like this code must have existed elsewhere. Is this a copy? Can we reuse the code? A bunch of static helpers?

the same goes for everything else in this file

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  ): TypeaheadSearchCallback {
+    return (text: string, activeItems: T[]): Observable<T[]> =>
+      api.filter({
+        filter: filterAnd(
+          contains<T, keyof T>(
+            key as keyof T,
+            text as any,
+            includeDefaultFilters && this.defaultFilter()
+          ),
+          notIn<T>(key as keyof AbstractModel, activeItems)
```
atruskie: i'm going to need a comment for this one. 

oh i think i worked it out: for a new typeahead query, only suggest things that haven't already been chosen

src/app/components/annotations/components/broad-filters-warning/broad-filters-warning.component.ts
```diff
@@ -0,0 +1,50 @@
...
+        class="btn-close"
+        aria-label="Close"
+        (click)="closeModal(false)"
+      ></button>
+    </div>
+
+    <div class="modal-body">
+      <p>
+        You no applied filter conditions. The verification grid will display all
+        available tasks.
```
atruskie:
```suggestion
        You are seeing <i>all</i> annotations because no search filters were used.
        All annotations that will be shown for verification.
```

src/app/components/annotations/components/broad-filters-warning/broad-filters-warning.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      </p>
+
+      <p>
+        This can result in a verification grid that is difficult to navigate and
+        may take a long time to complete.
```
atruskie:
```suggestion
        You can verify all annotations in one go, but it may take a long time to complete
        and it will be difficult to navigate. We suggest breaking up the task into smaller
        chunks.
```

src/app/components/annotations/components/broad-filters-warning/broad-filters-warning.component.ts
atruskie: future feature request: show number of annotations that will be in the set for the all/no filter case. Mybe omit the warning if it is less than 1000 etc..

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.html
```diff
@@ -0,0 +1,44 @@
...
+        <oe-spectrogram
+          [id]="contextSpectrogramId"
+          [src]="contextSource"
+          color-map="audacity"
+          scaling="stretch"
+        ></oe-spectrogram>
+      </oe-indicator>
+    </oe-axes>
+
+    <!-- <oe-media-controls [for]="contextSpectrogramId"></oe-media-controls> -->
```
atruskie: why disabled?

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.html
atruskie, line 1: i'd like to establish a naming convention for web-components

since they're standalone, they're truly different from other components we have and i'd like to demarcate that clearly

suggestions:

`web-components` folder?
`web-component` in the extension?

one or both or something else?

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.html
```diff
@@ -0,0 +1,44 @@
...
+    </span>
+  </div>
+
+  <div class="context-card-elements">
+    <oe-axes>
+      <oe-indicator>
+        <oe-spectrogram
+          [id]="contextSpectrogramId"
+          [src]="contextSource"
+          color-map="audacity"
```
atruskie: we should use grayscale here to match our host site

todo later: spectrogram settings per user site wide

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    searchParams.set("start_offset", newStart.toString());
+    searchParams.set("end_offset", newEnd.toString());
+    searchParams.delete("audio_event_id");
+
+    const newUrl = new URL(existingUrl.origin + existingUrl.pathname);
+    newUrl.search = searchParams.toString();
+
+    return newUrl.toString();
+  }
```
atruskie: this should be moved out of this component (TODO) and into the model component that's passed into it so we're just calling `model.contextUrl(30)` or something like that

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    const webComponents = await import("@ecoacoustics/web-components")
+    const token = webComponents.gridTileContext;
+
+    this.wrapper.nativeElement.dispatchEvent(
+      new ContextRequestEvent(
+        token,
+        this.handleContextChange.bind(this),
+        true
+      )
+    );
```
atruskie: we should bind this up into a decorator on a field just like `@ViewElementRef` (TODO)

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    return newUrl.toString();
+  }
+
+  public ngAfterViewInit(): void {
+    this.requestContext();
+  }
+
+  public handleContextChange(subjectWrapper: SubjectWrapper): void {
+    this.model.set(new Annotation(subjectWrapper.subject));
```
atruskie: why are we `new`ing here? shouldn't this be an `as` - just a type cast?

src/app/components/annotations/components/reset-progress-warning/reset-progress-warning.component.ts
```diff
@@ -0,0 +1,53 @@
...
+        class="btn-close"
+        aria-label="Close"
+        (click)="closeModal(false)"
+      ></button>
+    </div>
+
+    <div class="modal-body">
+      <p>
+        Because you have changed the verification tasks search parameters, your
+        progress will be reset if you continue.
```
atruskie:
```suggestion
        Because you have changed the verification search parameters, your
        progress will be reset if you continue.
```

src/app/components/annotations/components/reset-progress-warning/reset-progress-warning.component.ts
```diff
@@ -0,0 +1,53 @@
...
+      </p>
+
+      <p>
+        Do you wish to reset your progress and continue with the new search
+        parameters?
```
atruskie:
```suggestion
        Do you wish to reset your progress and continue with the new search
        parameters?
        <br/>
        <!-- TODO: replace with don't worry everything have been saved when we implement server persistence -->
        If you do wish to continue, make sure you download the annotations you
        have already done first!
```

src/app/components/annotations/pages/search/search.component.html
```diff
@@ -0,0 +1,72 @@
...
+    (searchParametersChange)="updateSearchParameters($event)"
+    [project]="project"
+    [region]="region"
+    [site]="site"
+  ></baw-annotation-search-form>
+
+  <div class="mt-2">
+    <!--
+      TODO: Once the audio events download route supports url encoded filters
+      we should uncomment this buttons functionality
```
atruskie, line 12: file an api issue and link back

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -0,0 +1,156 @@
...
+    if (this.sites) {
+      return filterModelIds("sites", this.sites);
+    } else if (this.regions) {
+      return filterModelIds("regions", this.regions);
+    }
+
+    return filterModelIds("projects", this.projects);
+  }
+
+  //! FOR PR REVIEWER: THIS IS WHERE I AM CURRENTLY UP TO
```
atruskie: expand into todo with context

src/app/helpers/filters/audioEventFilters.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @param startTime (optional) The minimum `recordedDate` allowed
+ * @param endTime (optional) The maximum `recordedEndDate` allowed
+ */
+export function filterTime(
+  initialFilter: InnerFilter<AudioEvent>,
+  ignoreDayLightSavings: boolean,
+  startTime?: Duration,
+  endTime?: Duration
+): InnerFilter<AudioEvent> {
+}
```
atruskie: as we discussed, not available with current API. leave these stubs and raise not implemented. cross link with api issue

src/app/models/data/Annotation.ts
```diff
@@ -0,0 +1,49 @@
...
+import { ITag } from "@models/Tag";
+import { ITagging, Tagging } from "@models/Tagging";
+
+export interface IAnnotation extends Required<IAudioEvent> {
+  tags: ITag[];
+  audioRecording: AudioRecording;
+  audioLink: string;
+}
+
+export class Annotation extends AbstractModelWithoutId implements IAnnotation {
```
atruskie, line 21: explain context behind this class

viewmodel

multiple models included to make rendering easier

multiple services used in a meta service to retrieve and create one of these

src/app/services/models/annotation.service.ts
```diff
@@ -0,0 +1,74 @@
...
+@Injectable()
+export class AnnotationService {
+  public constructor(
+    private tagsApi: TagsService,
+    private audioRecordingsApi: AudioRecordingsService,
+    private session: BawSessionService,
+    @Inject(API_ROOT) private apiRoot: string
+  ) {}
+
+  public async buildAnnotation(audioEvent: AudioEvent): Promise<Annotation> {
```
atruskie: todo: i'd like to see this class get an API like our other services. 

`show`, `create`, and `update`

src/app/services/models/annotation.service.ts
```diff
@@ -0,0 +1,74 @@
...
+        filter: {
+          id: {
+            in: tagIds,
+          },
+        } as any,
+      })
+    );
+  }
+
+  private async createAudioRecording(
```
atruskie: `create` naming is problematic, especially in a service context

suggest `fetch` or `show` or `list`

src/app/services/models/annotation.service.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    // if the user is logged in, we want to add their auth token to the
+    // query string parameters
+    // we do not add the auth token if the user is not logged in because the
+    // auth token will be undefined, resulting in `&user_token=undefined`
+    if (this.session.authToken) {
+      urlParams += `&user_id=${this.session.authToken}`;
+    }
+
+    return this.apiRoot + basePath + urlParams;
```
atruskie: this belongs elsewhere - probably in our media service?

atruskie (changes requested):
great work, a lot more to do but if we're functional let's do these things as needed

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.html
hudson-newey: Angular `[for]` attribtue gets removed when used on this media controls (for some reason)

I've commented it out for the time being, so that demos don't have a broken media controls component

You'll see that I've got the media controls working in the search card component using `ViewChild`. That's a hack, and not ideal.

I couldn't use `ViewChild` here (for some reason). Perhaps because it's in the shadow dom?

hudson-newey:
### Known issues I want to fix before meging into main

- [ ] I currently have a hack so that audio event `tags` attribute is converted into a string to get around https://github.com/ecoacoustics/web-components/issues/213
- [ ] The decision buttons sometimes error on initalization
- [ ] The verification grid can sometimes collapse into a 1x1 shape
- [x] The tile context doesn't disapear when paging in the verification grid
- [ ] You should only be able to open one context dialog at a time
- [x] Grid tile context can go off the edge of the screen
- [ ] Search page can lock up causing a crashed tab
- [x] Audio files can sometimes not load
- [x] Round start/end times to get around range request bug
- [x] `AnnotationSearchParameters` model should be constructed inside a resolver
- [ ] Search page has a lot of cumulitive shift when loading grid tiles
- [ ] Use grayscale spectrograms (depends on https://github.com/ecoacoustics/web-components/issues/179)
- [ ] `?page=n` query string parameters get removed when reloading the search page
- [ ] Annotation tiles are variable size on search page

### Like to do

- Unit/e2e tests

### Out of scope

- Re-order search annotations menu link (out-of-scope for PR; completing in https://github.com/QutEcoacoustics/workbench-client/issues/2149)

src/app/components/annotations/components/broad-filters-warning/broad-filters-warning.component.ts
```diff
@@ -17,13 +17,19 @@ import { NgbActiveModal } from "@ng-bootstrap/ng-bootstrap";
 
     <div class="modal-body">
       <p>
-        You no applied filter conditions. The verification grid will display all
-        available tasks.
+        You are seeing <em>all</em> annotations because no search filters were
+        used. All annotations will be shown.
```
atruskie:
```suggestion
        used.
```

src/app/components/annotations/components/broad-filters-warning/broad-filters-warning.component.ts
```diff
@@ -17,13 +17,19 @@ import { NgbActiveModal } from "@ng-bootstrap/ng-bootstrap";
...
-        This can result in a verification grid that is difficult to navigate and
-        may take a long time to complete.
+        You can verify all the annotations in one go, but it may take a long
+        time to complete and it will be difficult to navigate.
+      </p>
+
+      <p>
+        We suggest breaking up the task into smaller chunks. This can result in
+        a verification grid that is difficult to navigate and may take a long
+        time to complete.
```
atruskie:
```suggestion
        We suggest breaking up the task into smaller chunks. 
```

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -45,13 +44,12 @@ export interface IAnnotationSearchParameters {
   time: MonoTuple<Duration, 2>;
 }
 
+// we exclude project, region, and site from the serialization table because
+// we do not want them emitted in the query string
```
atruskie, line 70: we don't?

seems like a bug waiting to happen

I'm inferring that we take these values from route context?

src/app/components/annotations/pages/search/search.component.html
```diff
@@ -18,25 +18,43 @@ <h1>Annotation Search</h1>
...
+    <!--
+      We add the tooltip to a wrapping span because if the button is disabled
+      ng-boostrap will not show the tooltip
+     -->
+    <span
+      [ngbTooltip]="
+        session.isLoggedIn
+          ? 'Verify the annotations in the list'
+          : 'You must be logged in to verify annotations'
+      "
```
atruskie: this seems like it should be a reusable component (disabled unless logged in and tooltip to explain it)

src/app/services/models/annotation.service.ts
```diff
@@ -41,34 +45,103 @@ export class AnnotationService {
...
+    if (proposedEndTime > recordingDuration) {
+      const difference = proposedEndTime - recordingDuration;
+      proposedStartTime -= difference;
+      proposedEndTime -= difference;
+    }
+
+    return {
+      startTimeSeconds: proposedStartTime,
+      endTimeSeconds: proposedEndTime,
+    };
```
atruskie: this function does a lot of the same calculations as the audio link function. I'd suggest inlining it.

src/app/services/models/annotation.service.ts
```diff
@@ -41,34 +45,103 @@ export class AnnotationService {
...
+      );
+
+      safeEventStartTime = safeEventTimes.startTimeSeconds;
+      safeEventEndTime = safeEventTimes.endTimeSeconds;
+    }
+
+    const basePath =
+      `/audio_recordings/${audioEvent.audioRecordingId}/media.flac`;
+    let urlParams = `?audio_event_id=${audioEvent.id}` +
+      `&start_offset=${safeEventStartTime}&end_offset=${safeEventEndTime}`;
```
atruskie: you've got most of the parts in place here but I'd suggest a few tweaks:

- accept padding as an argument (default 0)
- calculate midpoint first
- if duration < 1 && padding  == 0 
    - then set padding to (1 - duration)/2
- then pad
- then floor/ceil start/end
- then bound check
   - on over/under flow, take excess and add it to the other side

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: Mock up with "Advanced Filters" section

image

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.html
hudson-newey: I like the idea of a `web-components` folder

The Angular style guide is pretty opinionated on file naming conventions, and I don't want to stray to far away from them.

Surprisingly, Angular doesn't specify any file structure for web components, so I think a web-components folder breaks the least Angular assumptions

If we were to put `web-component` in the file extension I think the least objectable option would be something like `grid-tile-content.web-component.component.ts` and at the worst case `grid-tile-content.web-component.component.spec.ts`

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.ts
hudson-newey: Good catch, the constructor prototype will still be attached because the web components pass around the subject by reference (we did this change a while back so that we would download results more efficiently)

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.html
atruskie: we agree on the folder

definite no to a triple extension.

I'm going to executively say either:

```
./web-components/a.component.ts
```

or 

```
./web-components/a.web-component.ts
```

you can choose

src/app/components/annotations/pages/search/search.component.ts
```diff
@@ -176,6 +179,29 @@ class AnnotationSearchComponent
...
+      } as any);
+      return initialFilter;
+    }
+
+    initialFilter.filter = filterAnd(initialFilter.filter, {
+      "audioRecordings.siteId": { in: Array.from(this.project.siteIds) },
+    } as any);
+
+    return initialFilter;
+  }
```
atruskie: this exact same function is in `verification.component.ts`?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: Feedback from offline review: This should be an accordian not a collapsable

src/app/components/annotations/pages/search/search.component.ts
hudson-newey: Yeah, I need to move this to `AnnotationSearchParameters.ts`.

To get the (resolved) project models in the annotation search parameters data model, I will need to use a resolver, which will take some time to make.

I did this hack so that I could get a working demo deployed to Ecosounds sooner, instead of spending a lot of time getting route resolvers working for the parameter data model

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: Update advanced filters section

image

Remember that WIP will be replaced with a placeholder when deployed to production

src/app/components/annotations/pages/search/search.component.html
hudson-newey: Maybe a directive that we can add to buttons and similar elements (instead of a wrapper component).

e.g.

```html
<button [requireLoggedIn]></button>

<div [requireLoggedIn]></div>
```

_Edit: After trying it, it'll probably be easier with a reusable component because we have to wrap the content in a span. We could do that with a directive, but it'd be overcomplicated_

src/app/services/models/annotation.service.ts
hudson-newey: Fixed and moved to `MediaService`

src/app/components/annotations/pages/annotationSearchParameters.ts
hudson-newey: Will be fixed when I resolve

https://github.com/QutEcoacoustics/workbench-client/pull/2145#discussion_r1802117157

src/app/components/annotations/pages/search/search.component.html
hudson-newey: Blocked by https://github.com/QutEcoacoustics/baw-server/issues/688

src/app/components/annotations/components/grid-tile-content/grid-tile-content.component.ts
hudson-newey: Something like this seems to do the trick

image

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -99,38 +156,98 @@ export class AnnotationSearchParameters
...
+  public routeFilters(): InnerFilter<AudioEvent> {
+    let siteIds: number[] = [];
+
+    if (this.routeSiteId) {
+      siteIds = [this.routeSiteId];
+    } else if (this.routeRegionId) {
+      siteIds = Array.from(this.routeRegionModel.siteIds);
+    } else {
+      siteIds = Array.from(this.routeProjectModel.siteIds);
+    }
```
atruskie, line 226: it would be better if these actually just filtered by the entities that are restricted rather than by site ids.

- it avoids the additional request
- and it's more semantically correct.

audio recording supports filtering by site, region, and project

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -99,38 +156,98 @@ export class AnnotationSearchParameters
...
+      initialFilter,
+      this.recordingDateStartedAfter,
+      this.recordingDateFinishedBefore
     );
+    // const dateTimeFilter = filterTime(
+    //   dateFilter,
+    //   this.daylightSavings,
+    //   this.recordingTimeStartedAfter,
+    //   this.recordingTimeFinishedBefore
+    // );
```
atruskie, line 254: why commented out?

src/app/components/annotations/pages/search/search.component.ts
```diff
@@ -144,13 +139,28 @@ class AnnotationSearchComponent
...
+      });
+
+      const response = await firstValueFrom(request);
+
+      const itemWarningThreshold = 1_000 as const;
+      const responseMetadata = response[0].getMetadata();
+      const numberOfItems = responseMetadata.paging.total;
+
+      if (numberOfItems > itemWarningThreshold) {
+        const warningModal = this.modals.open(this.broadFilterWarningModal);
```
atruskie, line 158: could we pass in the number of search results into the modal template? show it to the user

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -129,7 +120,17 @@ class VerificationComponent
...
-    const timeoutDuration = 1000;
+    // because the verification grid keybindings are scoped at the component
+    // level, we automatically focus the verification grid component so that
+    // users don't have to manually focus the verification grid to start using
+    // shortcuts
+    //
+    // without this automatic focusing, the user would have to click on the
+    // verification grid (e.g. to make a sub-selection) before being able to
+    // use the shortcut keys
+    this.verificationGridElement.nativeElement.focus();
```
atruskie: i wonder if we need a focus prompt/indicator on the verification grid... it could be confusing if keyboard shortcuts suddenly don't work

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -129,7 +120,17 @@ class VerificationComponent
...
+
+    const timeoutDuration = 1_000 as const;
```
atruskie: milliseconds?

our conventions are to suffix a unit in the variable name unless the value is a SI unit.

src/app/components/shared/if-logged-in/if-logged-in.component.ts
atruskie, line 1: this component is more simple than i was envisioning. I was thinking it could take an input for a capability to test against. that would make it useable for say the `canUpload` or `canDownload` audio capabilities. maybe todo for future scope?

but i'd consider renaming it to allow for future expansion. Maybe `can`. e.g. `<can if-logged-in>` or `<can [capability]="canDownloadCapability">`

src/app/components/shared/if-logged-in/if-logged-in.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  private disableInteractiveContent(state: boolean): void {
+    // TODO: use a ContentChild decorator here
+    const content =
+      this.elementRef.nativeElement.querySelectorAll("button, input");
+
+    for (const element of content) {
+      // if the disabled attribute is in the elements prototype, we want to
+      // set it to true
+      if ("disabled" in element) {
```
atruskie: hmmm this is testing the property is present not the attribute?

src/app/components/shared/if-logged-in/if-logged-in.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public updateState(): void {
+    this.disableInteractiveContent(!this.session.isLoggedIn);
+  }
+
+  private disableInteractiveContent(state: boolean): void {
+    // TODO: use a ContentChild decorator here
+    const content =
+      this.elementRef.nativeElement.querySelectorAll("button, input");
```
atruskie: we have other elements that have a disabled attribute. I'm not convinced we should be scanning for sub-elements to set this on. we should just set the attribute on the first child.

src/app/helpers/context/context-decorators.ts
```diff
@@ -0,0 +1,39 @@
...
+  return (target: WithContext, propertyKey: string) => {
+    const originalNgAfterViewInit = target.ngAfterViewInit;
+
+    target.ngAfterViewInit = function () {
+      // invoke the original ngAfterViewInit method if it exists
+      if (originalNgAfterViewInit) {
+        originalNgAfterViewInit();
+      }
+
+      const handler = target[propertyKey].bind(this);
```
atruskie: does it trigger angular change detection trigger when handler in invoked from the context provider

src/app/models/AudioRecording.ts
```diff
@@ -131,6 +131,10 @@ export class AudioRecording
     return apiRoot + audioRecordingOriginalEndpoint(this.id);
   }
 
+  public getSplittableUrl(apiRoot: string): string {
```
atruskie: we use the term `media` to describe fragments of an audio recording file. 
```suggestion
  public getMediaUrl(apiRoot: string): string {
```

src/app/services/baw-api/audio-recording/audio-recordings.service.ts
```diff
@@ -30,6 +33,12 @@ const downloadEndpoint = stringTemplate`/audio_recordings/downloader`;
  */
 export const audioRecordingOriginalEndpoint = stringTemplate`/audio_recordings/${audioRecordingId}/original`;
 
+/**
+ * A path that can be used to download audio recordings that have been split
+ * with start_offset and end_offset url parameters
+ */
+export const audioRecordingSplitableEndpoint = stringTemplate`/audio_recordings/${audioRecordingId}/media.${fileExtension}`;
```
atruskie: naming s/splitable/media/

src/app/services/models/annotation.service.ts
```diff
@@ -62,3 +66,62 @@ export class AnnotationService {
     );
   }
 }
+
+interface ResolverNames {
```
atruskie, line 70: custom resolvers are pretty rare right? maybe document why this one is needed (virtual service that need multiple other services)

atruskie (approved):

src/app/components/annotations/pages/annotationSearchParameters.ts
hudson-newey:
```ts
// time filtering is currently disabled until we can filter on custom fields
// and association
// see https://github.com/QutEcoacoustics/baw-server/issues/689
// TODO: enable time filtering once the api adds support
```

image

src/app/components/annotations/pages/search/search.component.ts
hudson-newey: Done

image

src/app/components/shared/if-logged-in/if-logged-in.component.ts
hudson-newey: I've read the w3 spec, and all `HTMLElement`'s have a global `disabled` attribute, meaning that this check is not necessary

<https://www.w3.org/TR/2011/WD-html5-20110525/elements.html>

src/app/components/annotations/pages/annotationSearchParameters.ts
hudson-newey: Audio recordings support filtering by associated site, region, and project models.

Because this data model constructs filter conditions for the audio_events API, our filter condition would have to look something like

```jsonc
// POST /audio_events/filter
{
	"filters": {
		// notice the double "." on associated models
		"audio_recordings.projects.id": {
			"in": [1, 2, 3, 4]
		}
	}
}
```

However because this is a filter condition on an association of an association it seems to fail

image

---

Potentially resolved by

https://github.com/QutEcoacoustics/baw-server/issues/689

atruskie: file a separate issue and cross link

hudson-newey: Associated issue

https://github.com/QutEcoacoustics/baw-server/issues/690

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: image

I've added a warning indicator if the verification grid loses focus (bound to the `blur` event)

I originally tried some text "Shortcuts are disabled" with a tooltip and button to re-gain focus, but there was a lot of cumulative shift, and the elements would move around a lot, so I ended up going with a simple warning icon + tooltip

src/app/components/shared/if-logged-in/if-logged-in.component.ts
hudson-newey: image

I like the idea. I made an interface that we can implement. It returns a watcher and a predicate that determines if the content should be enabled

src/app/helpers/context/context-decorators.ts
hudson-newey: No, calling the property doesn't inherently trigger a change detection cycle

I have fixed it so that it requests a change detection cycle after a context has been provided

PR #2165, opened 2024-11-07
Association decorators should not show error notifications

### LINKED ISSUE #1967: Association decorators should not show error notifications
Currently, when guest users access pages which use the association decorators to load data, error notifications are shown to them. This is an issue when they view the permissions widgets which retrieve the current user, but fail with a 401. These requests should not produce any warning messages.

This may need to be solved by either identifying association decorator requests, and handling them differently (this may help us with handling caching as well), or changing how notifications are handled.

hudson-newey:
# Association decorators should not show error notifications

This pull request creates a new association injector that can be used to inject different options into the `baw-api.service.ts`

## Changes

- Adds an `association-injector.service`. This injector used in the abstract model constructor
- Adds injectable `BawAptOptions` to the `baw-api.service`
- Fixes a bug where the associations would throw errors (e.g. the audio recording list page if not logged in)
- Fixes some minor formatting issues
- Corrects TypeScript typing for `ServiceToken.kind`
- Improves the TypeScript typing of `ServiceProviders` and `AbstractDataModel`
- Add missing `PageInfo` tests to annotation search page

## Issues

Fixes: #1967

## Final Checklist


src/app/models/AssociationDecorators.ts
```diff
@@ -173,8 +181,12 @@ function createModelDecorator<
...
     }
 
-    // Get Angular Injector Service
-    const injector = parent["injector"];
+    // the injector should be an AssociationInjector that was provided by the
+    // baw-api.service
+    // we use a different injector for associations so that the the we can
+    // inject the correct options for the baw-api service without affecting
+    // the options for the rest of the application
+    const injector: Injector = parent["injector"];
```
atruskie: can we type check or guard (on name?) this somehow?

src/app/services/association-injector/association-injector.service.ts
```diff
@@ -0,0 +1,39 @@
...
+          provide: BAW_SERVICE_OPTIONS,
+          useValue: { disableNotification: true },
+        },
+        ...services,
+        ...serviceTokens,
+        ...serviceResolvers,
+      ],
+    });
+  }
+}
```
atruskie: why is this a service?

Can't it, itself just be a useFactory?

```
{
  provide: Injector // or AssociationInjector if we subclass injector
  useFactory: (injector) => { 
     // rest of the owl
     Injector.create( // ...
   },
   deps: [Injector]
}
```

src/app/services/baw-api/ServiceTokens.ts
```diff
@@ -103,116 +104,57 @@ export class ServiceToken<
...
-export const AUDIO_EVENT_PROVENANCE = new ServiceToken<AudioEventProvenanceService, AudioEventProvenance>("AUDIO_EVENT_PROVENANCE");
-export const AUDIO_EVENT_SUMMARY_REPORT = new ServiceToken<EventSummaryReportService, EventSummaryReport>("AUDIO_EVENT_SUMMARY_REPORT");
-export const AUDIO_EVENT_IMPORT = new ServiceToken<AudioEventImportService, AudioEventImport>("AUDIO_EVENT_IMPORT");
-export const WEBSITE_STATUS = new ServiceToken<WebsiteStatusService, WebsiteStatus>("WEBSITE_STATUS");
-export const ANNOTATION = new ServiceToken<AnnotationService, Annotation>("ANNOTATION");
+export const AUDIO_EVENT_PROVENANCE = new ServiceToken< AudioEventProvenanceService, AudioEventProvenance >("AUDIO_EVENT_PROVENANCE");
+export const AUDIO_EVENT_SUMMARY_REPORT = new ServiceToken< EventSummaryReportService, EventSummaryReport >("AUDIO_EVENT_SUMMARY_REPORT");
+export const AUDIO_EVENT_IMPORT = new ServiceToken< AudioEventImportService, AudioEventImport >("AUDIO_EVENT_IMPORT");
+export const WEBSITE_STATUS = new ServiceToken< WebsiteStatusService, WebsiteStatus >("WEBSITE_STATUS");
+export const ANNOTATION = new ServiceToken<AnnotationService, Annotation>( "ANNOTATION");
```
atruskie: spaces around angle brackets?

src/app/services/baw-api/baw-api.module.ts
```diff
@@ -19,6 +21,10 @@ import { serviceResolvers, services, serviceTokens } from "./ServiceProviders";
       useClass: BawApiInterceptor,
       multi: true,
     },
+    {
+      provide: ASSOCIATION_INJECTOR.token,
+      useClass: AssociationInjectorService,
```
atruskie: yeah doubling down here, delete the service, inline the definition

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -145,13 +150,14 @@ export class BawApiService<
     @Inject(IS_SERVER_PLATFORM) protected isServer: boolean,
     protected manager: HttpCacheManager,
     protected http: HttpClient,
-    protected injector: Injector,
     protected session: BawSessionService,
     protected notifications: ToastrService,
-    @Inject(CACHE_SETTINGS) private cacheSettings: CacheSettings
+    @Inject(CACHE_SETTINGS) private cacheSettings: CacheSettings,
+    @Inject(ASSOCIATION_INJECTOR.token) protected associationInjector: any,
+    @Optional() @Inject(BAW_SERVICE_OPTIONS) private options: BawServiceOptions
```
atruskie, line 157: ok, flatten options with default in the ctor and store the merged result

- default options: static and frozen
- instance options: Object.assign({},  defaultOptions, options)
- and ideally options that are set instance wide are not configurable in the methods

Now that we have configurable services, do we need the options configurable from individual methods? Where is disableNotificaitons used elsewhere?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -133,9 +137,10 @@ export class BawApiService<
   // because users can input a partial options object, it is possible for the disableNotification property to be undefined
   // this can be a problem because it will default to a falsy value, not the default option
   // therefore, this method should be used to check if we should show a toast notification if the request fails
-  private shouldNotify(options: Partial<BawServiceOptions>): boolean {
+  private suppressErrors(options: Partial<BawServiceOptions>): boolean {
     return (
       options?.disableNotification ??
+      this.options?.disableNotification ??
       defaultBawServiceOptions.disableNotification
```
atruskie: eww - merge default and instance options

src/app/test/matchers/injectables.ts
```diff
@@ -342,6 +340,27 @@ const toHaveUrlActive = (util: MatchersUtil): CustomMatcher => ({
   },
 });
 
+const toBeProvidedBy = (): CustomMatcher => ({
```
atruskie, line 348: document

src/app/test/matchers/injectables.ts
```diff
@@ -342,6 +340,27 @@ const toHaveUrlActive = (util: MatchersUtil): CustomMatcher => ({
...
+  negativeCompare: (target: Injectable): CustomMatcherResult =>
+    target
+      ? matcherSuccess()
+      : matcherFailure("Expected target to be provided by"),
+  compare: (target: Injectable, injector: Injector): CustomMatcherResult => {
```
atruskie:
```suggestion
  compare: (target: Injectable, expectedInjector: Injector): CustomMatcherResult => {
```

src/app/test/matchers/injectables.ts
```diff
@@ -342,6 +340,27 @@ const toHaveUrlActive = (util: MatchersUtil): CustomMatcher => ({
...
+    if (!target) {
+      return matcherFailure("Injectable was not defined");
+    }
+
+    const targetProvider = target["injector"];
```
atruskie:
```suggestion
    const actualInjector = target["injector"];
```

atruskie (changes requested):

src/app/services/association-injector/association-injector.service.ts
hudson-newey: I've implemented this change. It does make providing the `ASSOCIATION_INJECTOR` slightly more verbose

e.g.
I now have to repeat this code whenever I want to provide the `AssociationInjector`. It is exactly the same, but I now have to provide the `deps` array

image

src/app/services/baw-api/baw-api.service.ts
hudson-newey: `disableNotification` is used in the `security.service` when we try to get the currently logged in user through a GET request.
Obviously in this case, we don't want to raise an error to the user if the user is not logged in, causing fetching the current user to fail.

src/app/models/AssociationDecorators.ts
hudson-newey: We can acomplish compile time type-checking with some fancy TypeScript branded types

image

We can't directly check the `name` property because injector types are broadened during compile time to `Injector` (losing name information)

The `name` property is also removed during runtime

_Edit: This was definately the correct call. I found so many bugs when I changed the TS type_

hudson-newey: Related commit

https://github.com/QutEcoacoustics/workbench-client/pull/2165/commits/792091386b679c88c291f17517a153d208800178

src/app/services/baw-api/baw-api.service.ts
atruskie: if that's the only other case, could we provide a service configured in the correct way and remove the method overload?

src/app/components/annotations/pages/search/search.component.ts
```diff
@@ -51,7 +53,7 @@ class AnnotationSearchComponent
     protected config: NgbPaginationConfig,
     protected modals: NgbModal,
     protected annotationService: AnnotationService,
-    private injector: Injector
+    @Inject(ASSOCIATION_INJECTOR) private injector: AssociationInjector
```
atruskie, line 56: ok this is a bit of a smell. the only place the association injector should be needed is from services that create models.

we should also be injecting this injector in creation of whatever is using it - the search parameters

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -75,7 +77,7 @@ class VerificationComponent
     private route: ActivatedRoute,
     private router: Router,
     private location: Location,
-    private injector: Injector
+    @Inject(ASSOCIATION_INJECTOR) private injector: AssociationInjector
```
atruskie, line 80: ditto here

src/app/components/harvest/components/metadata-review/folder-row.component.ts
```diff
@@ -146,7 +148,9 @@ export class FolderRowComponent {
     return this.harvestItem.report;
   }
 
-  public constructor(private injector: Injector) {}
+  public constructor(
+    @Inject(ASSOCIATION_INJECTOR) protected injector: AssociationInjector,
```
atruskie, line 152: ditto

src/app/components/projects/pages/permissions/permissions.component.ts
```diff
@@ -95,7 +97,7 @@ class PermissionsComponent
     private permissionsApi: PermissionsService,
     private accountsApi: AccountsService,
     private route: ActivatedRoute,
-    private injector: Injector
+    @Inject(ASSOCIATION_INJECTOR) private injector: AssociationInjector,
```
atruskie, line 100: ditto

src/app/components/reports/pages/event-summary/EventSummaryReportParameters.ts
```diff
@@ -79,12 +78,12 @@ const serializationTable: IQueryStringParameterSpec = {
...
   implements
     IEventSummaryReportParameters,
-    ImplementsInjector,
+    HasAssociationInjector,
     IParameterModel<EventSummaryReport>
 {
   public constructor(
     queryStringParameters: Params = {},
-    public injector?: Injector
+    public injector?: AssociationInjector
```
atruskie, line 86: ditto

src/app/helpers/advancedTypes.ts
```diff
@@ -63,3 +63,11 @@ export type RecursivePartial<T> = {
 
 /** This value may be asynchronous */
 export type PotentiallyAsync<T> = T | Promise<T>;
+
+// this brand symbol should never be used at runtime
+// it is only used to have a type that cannot be instantiated by TypeScript
+// outside of the Brand function
```
atruskie, line 69: and the purpose of this is?

To create something like a type alias that is not-equivalent as far as TS is concerned?

src/app/models/ImplementsInjector.ts
```diff
@@ -1,8 +1,13 @@
 import { Injector } from "@angular/core";
+import { Brand } from "@helpers/advancedTypes";
+import { AbstractModel } from "./AbstractModel";
+
+export type AssociationInjector = Brand<Injector, "AssociationInjector">;
```
atruskie, line 5: this is very fancy, but I wonder if you couldn't have just subclassed injector?

I guess with brand there's no runtime difference

src/app/services/baw-api/baw-api.service.ts
hudson-newey: Associated issue: #2166

atruskie (approved):

src/app/helpers/advancedTypes.ts
hudson-newey: Discussed offline that this needed an inline code comment

image

src/app/models/ImplementsInjector.ts
hudson-newey: There are a couple of benefits to using a branded type

1. There is no runtime difference (reducing the risk of bugs)
2. Using a subclassed injector risks polymorphic type matching

e.g.

```ts
export class AssociationInjector extends Injector {}

console.log(AssociationInjector === Injector);
//			^ will not throw a compile time error unless we add a branding property at compile time or run time
// 			  resulting in the same solution, just with an additional class that we have to maintain
```

src/app/components/annotations/pages/search/search.component.ts
hudson-newey: Related issue: #2167

Issue #2170, opened 2024-11-18
Cache filter requests


hudson-newey:
At the moment we are caching GET and HEAD requests

However, we can improve the performance quite a lot if we start caching filter requests (which are not cached by default because we abuse http semantics to send queries with a POST method and the filter parameters in the body).

- either using immutable objects with well defined hash protocol for all filters
bonus wins for structural equality
- or by controlling serialization ourselves with an interceptor, and then using payload as cache key, or hashing payload. Bonus more control over our serialization protocol (toJSON has been problematic in the past)

Issue #2160, opened 2024-10-31
Use show requests for `hasMany` associations


hudson-newey:
Once we can debounce duplicate http requests in #2159, we should consider using multiple `show` requests for `hasMany` associations.

The hope is that we will be able to debounce duplicate associations to improve performance and reduce the "real" number of outgoing requests

Blocked by: #2159

PR #2171, opened 2024-11-18
Add ng-http-caching request debouncing and caching

### LINKED ISSUE #2159: Find http request debounce interceptor
We currently use cashew for our request caching

However, cashew only caches responses, meaning that if the same request is sent out twice before a response is received, cashew will fail to cache the request.

To fix this we should look for an interceptor service that will debounce duplicate requests and merge the observable response.

hudson-newey:
# Add ng-http-caching request debouncing and caching

## Changes

- Replaces [ng-cashew](https://github.com/ngneat/cashew) with [ng-http-caching](https://github.com/nigrosimone/ng-http-caching) for request batching and improve caching
	- Allows us to only emit one GET request (by merging observables) if multiple requests contact the same endpoint
- Adds unit tests to ensure the correct requests clear the http cache
- Adds tests to the admin cache settings page

## Issues

Fixes: #2159

## Final Checklist


src/app/components/admin/settings/settings.component.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  beforeEach(() => {
+    setup();
+  });
+
+  assertPageInfo(AdminSettingsComponent, "Client Settings");
+
+  it("should create", () => {
+    expect(spectator.component).toBeInstanceOf(AdminSettingsComponent);
+  });
+});
```
atruskie, line 53: there's no actual tests here

src/app/components/admin/settings/settings.component.ts
```diff
@@ -41,9 +41,14 @@ import { adminSettingsMenuItem } from "./settings.menus";
         (change)="cacheSettings.setLogging($any($event.target).checked)"
       />
       <label class="form-check-label" for="enable-cache-logging">
-        Enable/Disable cache logging in the console
+        Enable cache logging in the console
       </label>
     </div>
+
+    <label class="mt-2">
+      Cache Time to Live (ms)
```
atruskie: units

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -105,14 +118,18 @@ export class BawApiService<
...
-   * API requests which include QSP and may be an issue in the future.
-   *
-   * @param path API path
-   */
-  private clearCache = (path: string) => {
-    this.manager.delete(path);
+  /** Clears the entire http cache */
+  private clearCache = () => {
+    // TODO: we should do targeting cache invalidation and invalidate all cache
+    // keys that contain the item being deleted from cache
```
atruskie: i don't understand what this means

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -105,14 +118,18 @@ export class BawApiService<
...
+    //
+    // this is harder than expected because cached requests sometimes contain
+    // url parameters, we also need to invalidate all cache items that which
+    // reference the item being deleted as an associated model.
```
atruskie: i think you're overthinking this.

The method is documented as clearing the whole cache, and as long as usage is consistent with that, then we don't need all these extra notes.

The intent of this cache is reduce page load speeds for repeated items. It is intended to be very short lived - like 60 seconds. We shouldn't have to be invalidating it to often.

src/app/services/cache/cache-settings.ts
```diff
@@ -1,10 +1,16 @@
...
-  /** TTL for HTTP GET requests */
-  public httpGetTtlMs = 1000;
+  public constructor(_enabled: boolean, withLogging: boolean) {
+    this._enabled = _enabled;
+    this.withLogging = withLogging;
+  }
 
-  public constructor(private _enabled: boolean, private withLogging: boolean) {}
+  /** time to live for cached HTTP requests */
+  public cacheTtlMs = NgHttpCachingConfigDefault.lifetime;
```
atruskie: units

src/app/services/cache/ngHttpCachingConfig.ts
```diff
@@ -0,0 +1,97 @@
...
+    withCacheLogging(req.context);
+  }
+
+  return isCacheable;
+}
+
+// TODO: when we add support for caching filter requests we will need to use a
+// custom cache key function that includes the request body in the cache key
+// see: https://github.com/QutEcoacoustics/workbench-client/issues/2170
+//
```
atruskie, line 96: the current commented implementation of using object-hash is not ideal, and should be marked as such or removed

atruskie (changes requested):
Minor changes.

Most importantly, I can't see what the default TTL is. Remember this is a performance optimisation task, not a caching task. We don't need to reinvent the browser's cache, so we should be using a short TTL.

Measure the performance of this change:

- 2 pages
- 4 TTLs: 0, 200ms, 10 seconds, 60 seocnds
- 10 samples each

See if average load times change.

src/app/services/baw-api/baw-api.service.ts
hudson-newey: It was supposed to add a todo stating "instead of clearing the entire cache when a single item is updated, we should just delete that one item from the cache. However, when an model is updated/deleted, it may be used as an association to another model, meaning that we will have to invalidate the changed model, and any other models that reference the newly invalidated model".

However, since we have lowered the cache lifetime to one second and our caching is designed to reduce the impact of associations and N+1 queries, this is no longer a problem we have to worry about.

src/app/services/baw-api/baw-api.service.ts
hudson-newey: Discussed offline

I agree and have reduced the caching time

The purpose of this cache is not to re-create the browser cache but to consolidate duplicate association queries

hudson-newey:
### Methodology

Results were gathered using a docker container to replicate the production environment as close as possible.

The "Cooloola" "Annotation Search" page was used in all tests.

Before each testing round, the Cooloola annotation search page was loaded without recording the results so that all audio files would be recalled, and would minimise deviations in the tests due to server caching.

A new browser tab was used between each test to prevent potential v8 caching.

Network caching was disabled.

---

The starting page for the tests was the "Cooloola" project details page.

The "Cooloola" project details page would be hard-reloaded, and the "Search Annotations" menu item would be clicked.

Performance would be measured from the start of the menu's "click" event to the end of the annotation search page load.

### Performance test results

Results are recorded in seconds

| Cache lifetime        | 1     | 2     | 3       | 4     | 5     | 6     | 7     | 8     | 9     | 10    |     | Average | Std-Deviation |
| --------------------- | ----- | ----- | ------- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | --- | ------- | ------------- |
| Baseline (without caching)              | 3.852 | 3.285 | 3.844 | 3.886 | 3.968 | 3.698 | 3.755 | 3.819 | 3.509 | 4.102 |     | 3.772   | 0.232         |
| 0ms (with debouncing) | 3.370 | 3.134 | 4.048 | 3.225 | 3.664 | 3.935 | 3.370 | 2.731 | 3.496 | 3.410 |     | 3.438   | 0.383         |
| 200ms                 | 2.830 | 2.614 | 2.801 | 3.436 | 3.676 | 3.232 | 3.086 | 3.200 | 3.339 | 3.274 |     | 3.149   | 0.322         |
| 10s                   | 3.179 | 3.047 | 3.283 | 3.747 | 3.207 | 3.450 | 3.090 | 3.257 | 3.553 | 3.571 |     | 3.338   | 0.230         |
| 60s                   | 3.478 | 3.397 | 3.341 | 3.152 | 3.191 | 2.997 | 3.182 | 2.681 | 3.426 | 3.331 |     | 3.218   | 0.239         |

atruskie (approved):

atruskie:
From discussion of the results we can see that the actual cache part of this library has little to no effect on performance. TTL does not seem to matter either. The main win is the request de-duplication.

so the understanding here is:
 
- debounce is effective at deduplicating requests for the same data on a page that is loading
- TTL has no effect unless more data is requested later by a user

We're going to trial 10 second TTL with the hope that for anyone doing any kind of quickly successive actions, they'll see a benefit from the cache (like flicking through pages of results). Also at the same time, anyone doing anything on a page is likely to take more than 10 seconds, stale cache results should basically be a non-issue.

We'll do some more testing when #2160 is implemented. #2170 should also help.

PR #2184, opened 2024-12-09
Fix hard failure on google maps error

### LINKED ISSUE #2175: Adblockers blocking Google Maps will cause the client to hard-fail
I have just experienced a case where my adblocker caused Google maps to break / not load.

This caused the entire application to hard-fail.

This highlights an underlying issue with our architecture.
If Google maps fails to load, only Google maps should stop working and should not cause the entire application to crash.

We should instead allow Google maps to fail, and overlay the error message where the Google maps placeholder would be. Displaying an error message such as "Maps failed to load: Request was blocked by ad blocker" (or similar)

hudson-newey:
# Fix hard failure on google maps error

## Changes

- Fixes a bug where if Google Maps failed to embed, the entire site would hard-fail
- Move the `embedGoogleMaps.ts` helper scripts into a DI service
- Lazy load google maps to only embed when we need it for the current page (improving performance)
- Asyncronously embed google maps (embedding google maps is no longer a blocking operating, improving performance and reducing the scope for bugs)
- Adds loading spinner when Google maps is loading
- Adds failure state when Google maps fails to load
- Re-enables map component tests

## Problems

Any problems caused by the PR that will need to be address in another PR. For example, these issues may be caused by another branch which will change how this branch functions.

## Issues

Fixes: #2175

## Final Checklist


src/app/components/shared/map/map.component.ts
```diff
@@ -72,31 +56,25 @@ export class MapComponent
   // Setting to "hybrid" can increase load times and looks like the map is bugged
   public mapOptions: MapOptions = { mapTypeId: "satellite" };
   public bounds: google.maps.LatLngBounds;
-  private updateMap: boolean;
+  private updateMap: boolean = false;
```
atruskie: what is the purpose of this field, what does it do?

src/app/services/maps/maps.service.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      this.loadPromise = { promise, resolve: resolver, reject: rejector };
+
+      this.embedGoogleMaps();
+    } else {
+      console.warn("Google Maps Service already embedded.");
+    }
+  }
+
+  private static embeddedService = false;
```
atruskie: naming. Suggest `embedded` or `hasScriptBeenEmbedded`

src/app/services/maps/maps.service.ts
atruskie, line 1: ok it seems you've tried to use a promise like a signal.

If you actually want to emit all the various states, perhaps an observeable is the right choice here.

Otherwise, your promise can return values too you know. Like `Loaded` or `Failed`. 

I don't understand why you need `SharedPromise` and I don't understand why `loadAsync` isn't just

```
return this.loadPromise;
```

And `embedGoogleMaps` is an async function that returns a promise. Why is that not just assigned to `this.loadPromise` ?

This file needs a simplicity pass.

atruskie (changes requested):

src/app/components/shared/map/map.component.ts
hudson-newey: Google maps runs the `ngAfterViewChecked` method every 100ms, even with `DetectChanges.OnPush`

This `updateMap` exists so that we don't need to run this code every 100ms
image

I'm looking into it now because I suspect this is something that we're doing wrong; I wouldn't expect Google to author code that continuously runs `ngAfterViewChecked`

atruskie: oh i didn't see that function.

you're right though, seems weird we're hooked into ngAfterViewChecked. Perhaps another hook is more appropriate?

hudson-newey: Update: `ngAfterViewChecked` is due to the `ViewChild` decorators for the `GoogleMap`, `MapInfoWindow`, and `MapMarker` components

atruskie (approved):

PR #2206, opened 2025-01-30
Create read interface for analysis jobs

### LINKED ISSUE #2207: analysis jobs read only interface
MVP read only interface

hudson-newey:
# Create read interface for analysis jobs

This pull request adds support for reading analysis jobs and scripts.

Both at the admin (site wide) and project level.

## Changes

- Updates the analysis job model
- Updates the scripts model
- Updates the admin analysis job list/details view
- Updates the admin script details view
- Adds project level analysis job list/details view

## Outside of scope

This PR **does not** add a UI for provenance models

## Issues

Fixes https://github.com/QutEcoacoustics/workbench-client/issues/2207

## Visual Changes

image

_Admin analysis job list view_

image

_Analysis job project details view_

---

image

_Admin script list view_

image

_Script details view_

## Final Checklist


src/app/components/admin/admin.menus.ts
```diff
@@ -28,7 +28,7 @@ export const adminUserListMenuItem = menuRoute({
 
 export const adminAnalysisJobsMenuItem = menuRoute({
   icon: ["fas", "server"],
-  label: "Analysis Jobs",
+  label: "Audio Analysis",
```
atruskie: why the label change? _Analysis Jobs_ was accurate

src/app/components/audio-analysis/audio-analysis.menus.ts
```diff
@@ -1,41 +1,44 @@
...
   label: "New Analysis Job",
   tooltip: () => "Create a custom analysis job",
   route: audioAnalysesRoute.add("new"),
   parent: audioAnalysesMenuItem,
-});
 
-export const audioAnalysisRoute = audioAnalysesRoute.add(":analysisJobId");
+  // TODO: I have hidden the new analysis job menu item behind the isAdmin
+  // predicate because the old baw client interface doesn't support scoping
+  // to projects
```
atruskie: the old baw-client UI is 100% incompatible with the new API. Consider it deprecated... 

to which, I don't understand this comment

src/app/components/audio-analysis/audio-analysis.menus.ts
atruskie, line 1: can we simplify the names in this file?

Technical names should `analysisJobs`. We can use user friendly terminology in the labels.
e.g. `audioAnalysisMenuJobItem` should be `analysisJobsMenuItem`

src/app/components/audio-analysis/audio-analysis.module.ts
```diff
@@ -2,20 +2,22 @@ import { NgModule } from "@angular/core";
...
-  AudioAnalysisComponent,
-  AudioAnalysisResultsComponent,
+  NewAudioAnalysisJobComponent,
+  AudioAnalysisJobComponent,
+  AudioAnalysisJobResultsComponent,
 ];
 
-const routes = audioAnalysesRoute.compileRoutes(getRouteConfigForPage);
+const newRoutes = audioAnalysesRoute.compileRoutes(getRouteConfigForPage);
+const oldClientRoutes = oldClientAnalysesRoute.compileRoutes(getRouteConfigForPage);
```
atruskie: just ditch this, they won't work now

src/app/components/audio-analysis/audio-analysis.routes.ts
```diff
@@ -0,0 +1,13 @@
+import { projectRoute } from "@components/projects/projects.routes";
+import { StrongRoute } from "@interfaces/strongRoute";
+
+const analysesPath = "audio_analysis";
+const analysisJobPath = ":analysisJobId";
+
+export const oldClientAnalysesRoute = StrongRoute.newRoot().add(analysesPath);
+export const oldClientAnalysisJobRoute = oldClientAnalysesRoute.add(analysisJobPath);
+export const oldClientNewAnalysisJobRoute = oldClientAnalysesRoute.add("new");
+export const oldClientAnalysisJobResultsRoute = oldClientAnalysisJobRoute.add("results");
```
atruskie: scrap

src/app/components/audio-analysis/audio-analysis.routes.ts
```diff
@@ -0,0 +1,13 @@
+import { projectRoute } from "@components/projects/projects.routes";
+import { StrongRoute } from "@interfaces/strongRoute";
+
+const analysesPath = "audio_analysis";
```
atruskie:
```suggestion
const analysesPath = "analysis_jobs";
```

src/app/components/audio-analysis/audio-analysis.routes.ts
```diff
@@ -0,0 +1,13 @@
...
+const analysisJobPath = ":analysisJobId";
```
atruskie: weird naming - it's not a path, it's more like a route parameter

src/app/components/audio-analysis/pages/list/list.component.html
```diff
@@ -0,0 +1,55 @@
...
+        <baw-user-link [user]="value.creator"></baw-user-link>
+      </ng-template>
+    </ngx-datatable-column>
+    <ngx-datatable-column name="Started" [width]="110" [maxWidth]="110">
+      <ng-template let-value="value" ngx-datatable-cell-template>
+        @if (value) {
+        <baw-datetime [value]="value"></baw-datetime>
+        } @else {
+        <span>Unknown</span>
+        }
```
atruskie, line 36: i would have expected baw-datetime should be able to handle null/undefined

src/app/components/audio-analysis/pages/list/list.component.html
```diff
@@ -0,0 +1,55 @@
...
+      </ng-template>
+    </ngx-datatable-column>
+    <ngx-datatable-column name="Status"></ngx-datatable-column>
+    <ngx-datatable-column name="Status Updated" [width]="115" [maxWidth]="115">
```
atruskie, line 40: why not use `updated_at`

src/app/components/audio-analysis/pages/list/list.component.ts
```diff
@@ -1,23 +1,119 @@
...
+    protected route: ActivatedRoute
+  ) {
+    super(api, (analysisJobs) =>
+      analysisJobs.map((analysisJob) => ({
+        name: analysisJob.name,
+        scripts: analysisJob,
+        creator: analysisJob,
+        started: analysisJob.startedAt,
+        status: analysisJob.overallStatus,
+        statusUpdated: analysisJob.overallStatusModifiedAt,
```
atruskie: yeah just switch this to updated_at

src/app/components/audio-analysis/pages/list/list.component.ts
```diff
@@ -1,23 +1,119 @@
...
+    statusUpdated: "overallStatusModifiedAt",
+  };
+
+  public get project(): Project | null {
+    const routeProjectModel = this.route.snapshot.data[projectKey];
+    if (routeProjectModel) {
+      return routeProjectModel.model;
+    }
+
+    return null;
```
atruskie: why would it ever by null?

didn't I see on the admin component you overrode the project property?

src/app/components/audio-analysis/pages/list/list.component.ts
```diff
@@ -1,23 +1,119 @@
...
+  }
+
+  protected override apiAction(
+    filters: Filters<AnalysisJob>
+  ): Observable<AnalysisJob[]> {
+    const filterByProject: Filters<AnalysisJob> = {
+      filter: {
+        or: {
+          projectId: { eq: this.project?.id },
```
atruskie: probably shouldn't include this filter condition if the project id is null

src/app/components/audio-analysis/pages/new/new.component.ts
atruskie, line 1: delete

src/app/components/audio-analysis/pages/results/results.component.ts
atruskie, line 1: delete

src/app/components/scripts/scripts.menus.ts
```diff
@@ -4,58 +4,54 @@ import {
...
     const componentModel = routeData.root.firstChild.data;
     const scriptName = componentModel.script.model.name;
     return scriptName;
   }
 });
 
+export const newScriptMenuItem = menuRoute({
+  icon: defaultNewIcon,
+  label: "New Script",
+  route: scriptsMenuItem.route.add("new"),
```
atruskie, line 43: is this going anywhere?

src/app/models/AnalysisJob.ts
```diff
@@ -92,21 +141,33 @@ export class AnalysisJob extends AbstractModel implements IAnalysisJob {
...
+    }
+
+    return audioAnalysisMenuJobItem.route.format({
       analysisJobId: this.id,
+      projectId,
     });
   }
 
+  public get viewUrl(): string {
+    throw new Error("method not implemented");
```
atruskie: why not implemented? you have all the information needed....

atruskie (changes requested):
Did we get rid of all the old links to the old-client analysis jobs interface?

src/app/components/admin/admin.menus.ts
hudson-newey: Discussed offline, I will revert this change

hudson-newey: For consistency, chaning this `admin.menu` label also means that the `analysis-jobs.menu` label has also been changed to "Analysis Jobs"

src/app/components/audio-analysis/audio-analysis.routes.ts
hudson-newey: It's weird naming, but it's consistent with the rest of the client

see

- https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/audio-recordings/audio-recording.routes.ts#L7-L9
- https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/harvest/harvest.routes.ts#L3-L5
- https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/projects/projects.routes.ts#L3-L5
- https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/regions/regions.routes.ts#L4-L6

---

So there are three options here:

- Continue using the "Path" suffix for consistency (even though it's semantically incorrect)
- Use "Parameter" suffix in this case only (breaking consistency), but keeping to scope and making this route variable semantically correct
- Refactor all "Path" suffix's to use the "Parameter" suffix, expanding scope, but improving variable semantics across the board

src/app/components/audio-analysis/pages/list/list.component.html
hudson-newey: Didn't have a hard reason to use `status_updated_at` instead of `updated_at`

You know the system $$10\times$$ better than me, so if you think that using `updated_at` is better I'll agree

src/app/components/audio-analysis/pages/list/list.component.ts
hudson-newey: Yeah, it should never be null, but this is just defensive programming

If I didn't return `null` here, the `Project` model would be typed as `private project: Project | null | undefined;` (because no return is implicit `undefined`.

In hind sight, it might be good to throw an error here so we hard fail if the project (somehow) fails to initialize.

src/app/components/audio-analysis/pages/list/list.component.ts
hudson-newey: image

Line below handles this condition

However, I agree that it's not very clear and it's not very readable

src/app/components/scripts/scripts.menus.ts
hudson-newey: You can access this route if you are an admin by navigating through

Admin dashboard > Scripts > New Script (in side bar)

src/app/models/AnalysisJob.ts
hudson-newey: Kind of. I previously had a `createViewUrl` method that allowed system analysis jobs to be fake scoped to a project

image

The problem is that system analysis jobs do not have a `project_id` field, so I only have enough information to return the `adminViewUrl` for system analysis jobs

src/app/components/audio-analysis/pages/list/list.component.html
hudson-newey: At the moment it doesn't. It will display "Invalid Date Time"

However, I agree that `null` and `undefined` should display "Unknown"

hudson-newey: Actually, looking at the work that would be required to implement this change, I suggest that it should be a new dev concern issue.

It will require a substantial amount of changes that are out of scope for this PR

src/app/components/audio-analysis/audio-analysis.routes.ts
atruskie: ok, i don't need to die on this hill

src/app/components/audio-analysis/pages/list/list.component.html
atruskie: file an issue, keep a reference in the target issue to the usage in this file to clean up afterwards

hudson-newey: Issue reference: https://github.com/QutEcoacoustics/workbench-client/issues/2212

src/app/components/audio-analysis/pages/list/list.component.ts
atruskie: oh i see

i'd wrap this in an if statement. the mixed up order is  hard to read.


```
filter = { filter: {}};
if (this.project) {
   filter.filter = { or: ...}
}
```

src/app/components/scripts/scripts.menus.ts
atruskie: yeah but we don't have a view for it right?

hudson-newey: We do and it was automatically updated when the `scripts.schema` was updated

https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/admin/scripts/new/new.component.ts

It doesn't have any of the fancy syntax highlighting that we want to do in the future, but it exists right now

I'll double check that it works as expected

src/app/models/AnalysisJob.ts
atruskie: well in the common case most jobs will have project id

can view url take an optional parameter?

hudson-newey: The `viewUrl` is a getter, so it can't take in an optional parameter, hence the `createViewUrl` method. `createViewUrl` is a convention of our models (though not backed up by any TS typing).

atruskie: k, i still reckon a default implementation is better than none

atruskie (approved):

src/app/components/audio-analysis/pages/list/list.component.ts
hudson-newey: Good idea, I am thinking something like this (word "like" is being used here because of type errors) using the `filterOr` helper

image

PR #2214, opened 2025-02-12
Update annotation import

### LINKED ISSUE #2200: Upgrade annotation import api service to new baw-server version format
In a recent commit (https://github.com/QutEcoacoustics/baw-server/commit/9ba186c99066d8fa8f9f3babc4caaa50b6f4d313) the baw-server changed its request/response format for the baw-server.

This breaking change means that the annotation import page has broken.

This page is not currently user facing as it's hidden behind the admin guard

hudson-newey:
# Update annotation import

Following breaking changes to the annotation import API, I have restructured the annotation import API and UI to reflect these changes.

## Changes

- Adds `AnnotationImportFile` models & services
- Adds error card component
- Adds `bawVirtualDatatablePagination` directive so that we can use table pagination without an API backing
- Adds new `typeaheadCallback` method to the `tags.service`
- Adds handling of edge case where the typeahead input doesn't have a callback instantiated
- Updates `AnnotationImport` model
- Updates `ImportedAudioEvent` model
- `AbstractModel` `getJsonAttributes(opts: Xor<{ create: boolean} | { update: boolean }>)` has been updated to `getJsonAttributesForCreate` and `getJsonAttributesForUpdate`
- The `baw-api` service `create` and `update` methods now accept a `params` object for unscoped form data
- Updates `BawApiError`s to provide the responses `data`
- Removes `AnnotationImportFileRead` and `AnnotationImportFileWrite` models (models returned for reading / writing used to be different. This was fixed in the API refactor)

## Issues

Fixes: #2200

## Potential future changes

- When clicking on the "Add annotations" button, it should take you to a combined page to create a new annotation import and add files
- Users might want to edit their annotation before they commit them. (have input boxes so that they can change the values)

## Visual Changes

image

_Annotation import details page (events)_

image

_Annotation import details page (files)_

image

_Add new annotations page (with no uploaded files)_

image

_Add new annotations page (with errors)_

(note: with the new columns, the table is now horizontally scrollable)

## Final Checklist


src/app/components/import-annotations/add-annotations/add-annotations.component.html
```diff
@@ -0,0 +1,228 @@
...
+              else audioRecordingTemplate
+            "
+            size="sm"
+          ></baw-loading>
+
+          <ng-template #audioRecordingTemplate>
+            <a [bawUrl]="value.audioRecording.viewUrl">
+              {{ value.audioRecording.id }}
+            </a>
+          </ng-template> -->
```
atruskie: Useless?

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,360 @@
...
+  protected errorCardStyles = ErrorCardStyle;
+
+  /**
+   * A boolean value indicating if the currently buffered audio event import can
+   * be successfully committed
+   */
+  protected valid: boolean = false;
+  protected uploading: boolean = false;
+  protected importFiles: File[] = [];
+  protected additionalTagIds: Id[] = [];
```
atruskie: Smells: additional tags are per file. The relationship seems mismatched?

atruskie: oh I think I get it now. it is one to one, but multiple files can be imported from the file picker?

src/app/components/import-annotations/add-annotations/add-annotations.component.html
```diff
@@ -0,0 +1,237 @@
...
+    saved.
+  </p>
+
+  <p>
+    You can review these events before saving them with the "Import Annotations"
+    button.
+  </p>
+
+  <div>
+    <ngx-datatable
```
atruskie, line 136: can this data table show events from more than one file? If so, we need a file column.

Probably need an index column too.

Maybe this could be combined: 

| ID (tooltip: "file index and even index in each file") |
|----|
| 0:0 |
| 1:0 |
| 1:1234|

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  /**
+   * Errors that apply to the entire file that we cannot display in the events
+   * table.
+   * E.g. duplicate file uploads, names, etc...
+   */
+  protected importErrors: BawErrorData[] = [];
```
atruskie: i'm again wondering what the arity is here.

if these are structured how i'm assuming, perhaps it should be an array of objects, rather than three arrays

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (!this.hasFiles) {
+      return "Please select a file to upload";
+    } else if (!this.valid || !this.form.valid) {
+      return "Please fix all errors before submitting";
+    } else if (this.uploading) {
+      return "Please wait for the current upload to complete";
+    }
+
+    // we should never hit this condition, but if we do, I return null so that
+    // we don't end up in an unknown state
```
atruskie: nit: concise comments are less load

```suggestion
    // should never hit, but be safe
```

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+  protected searchTagsTypeaheadCallback = (
+    text: string,
+    activeItems: Tag[]
+  ): Observable<Tag[]> =>
+    this.tagsApi.filter({
+      filter: filterAnd(
+        contains<Tag, "text">("text", text),
+        notIn<Tag>("text", activeItems)
+      ),
+    });
```
atruskie: feels like we use this a lot, and thus if it is here, it isn't DRY.... factor out?

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  // since the typeahead input returns an array of models, but the api wants the associated tags as an array of id's
+  // we use this helper function to convert the array of models to an array of id's that can be sent in the api request
```
atruskie: Perhaps the filter should accept an array of tags or ids? We use that convention elsewhere

```suggestion
  // typeahead returns models, we need ids for filter
```

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    const files: FileList = target.files;
+    const bufferedFiles = Array.from(files);
+
+    // sometimes the operating system will report files such as csv files as
+    // excel file types.
+    // because the api cannot process excel files, it will reject it and return
+    // an error.
+    // to fix this, we will change the file type to the correct type using the
+    // file extension.
+    const extensionMappings = new Map<string, string>([["csv", "text/csv"]]);
```
atruskie: don't redeclare this map in the handler each time. Make it a real constant

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+      }
+
+      return file;
+    });
+
+    this.performDryRun();
+  }
+
+  protected hasRecordingErrors(model: ImportedAudioEvent): boolean {
+    return model.errors.some((error) => "audioRecordingId" in error);
```
atruskie, line 275: is this a constrained string checked by TS? otherwise magic string

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  // sends all import groups to the api if there are no errors
+  protected commitImports(): Promise<void> {
+    // importing invalid annotation imports results in an internal server error
+    // we should therefore not submit any upload groups if there are any errors
+    if (!this.canCommitUploads()) {
+      return;
+    }
+
+    // creates a lock so that no more files can be added to the upload queue while the upload is in progress
```
atruskie: this comment belongs on the field declaration

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    this.uploading = true;
```
atruskie: we use this pattern a lot. Future idea a `Flag` class. Methods like `this.uploading.lock()`, `this.uploading.unlock()`, `this.uploading.lockFor( () => { } )`. It can also have helpers for locking and unlocking with promises/observables.

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+        complete: () => {
+          this.notifications.success("Successfully imported annotations");
+
+          this.router.navigateByUrl(
+            annotationImportRoute.toRouterLink({
+              annotationId: this.audioEventImport.id,
+            })
+          );
+        },
+      });
```
atruskie: what happens on failure?

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+        this.importErrors.push(...this.extractFileErrors(file, error));
+
+        // the error data can either be an AudioEventImportFile or an array of
+        // AudioEventImportFile
+        // because we know that dryRun will only return a single model, we can
+        // safely cast the array type away
+        //
+        // TODO: this should be properly fixed by modifying the typing of the
+        // BawApiError class
+        return of(error.data as AudioEventImportFile);
```
atruskie: seems to be worth fixing

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+      // the file name
+      // this makes the error message more user friendly and easier to
+      // understand when you upload multiple files
+      if (fileNameKey in errorInfo) {
+        errorInfo[file.name] = errorInfo[fileNameKey];
+        delete errorInfo[fileNameKey];
+      }
+    }
+
+    return newErrors;
```
atruskie: seems like some of this should be part of the service/error models

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
```diff
@@ -0,0 +1,364 @@
...
+      ];
+    }
+
+    // because we are mutating the error object (for nicer error message), I
+    // create a new object so that I don't accidentally mutate the original
+    // error by reference
+    const newErrors: BawErrorData[] = Array.isArray(error.info)
+      ? Object.assign({}, error.info)
+      : [Object.assign({}, error.info)];
+
```
atruskie: why mutate at all?

src/app/models/AbstractModel.ts
```diff
@@ -78,7 +90,91 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+    return this[AbstractModel.keys.meta];
+  }
+
+  public addUnscopedAttribute(attribute: unknown) {
+    return this.getUnscopedAttributes().push(attribute);
+  }
+
+  public getUnscopedAttributes(): any[] {
+    return [];
+  }
```
atruskie: not keen.  We don't need such broad functionality.

src/app/models/AbstractModel.ts
```diff
@@ -78,7 +90,91 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+    return this.formDataOnlyAttributes({ update: true });
+  }
+
+  public getFormDataOnlyAttributes(): FormData {
+    return this.formDataOnlyAttributes();
+  }
+
+  public getJsonAttributes(): Partial<this> {
+    return this.jsonAttributes();
+  }
```
atruskie: are these actually used anywhere?

src/app/models/AbstractModel.ts
```diff
@@ -78,7 +90,91 @@ export abstract class AbstractModelWithoutId<Model = Record<string, any>> {
...
+    return this.hasFormDataOnlyAttributes({ update: true });
+  }
+
+  public getFormDataOnlyAttributesForUpdate(): FormData {
+    return this.formDataOnlyAttributes({ update: true });
+  }
+
+  public getFormDataOnlyAttributes(): FormData {
+    return this.formDataOnlyAttributes();
+  }
```
atruskie: concise docs needed for all of these

```
/* { @bawPersistAttr({create: true} x = 1, @bawPersistAttr({supportedFormats: ['form-data']}) y = 2
 *  getFormDataOnlyAttributesForCreate() => { y = 2 }
 */  
```

src/app/models/AttributeDecorators.ts
```diff
@@ -284,6 +284,10 @@ export function bawBytes<Model>(opts?: BawDecoratorOptions<Model>) {
   });
 }
 
+export function unscopedAttribute<Model>(opts?: BawDecoratorOptions<Model>) {
+  return createDecorator<Model>(opts, () => {});
+}
+
```
atruskie: 👎

src/app/models/AudioEvent.ts
```diff
@@ -66,7 +73,9 @@ export class AudioEvent
   public readonly updatedAt?: DateTimeTimezone;
   @bawDateTime()
   public readonly deletedAt?: DateTimeTimezone;
+  @bawPersistAttr()
```
atruskie: hmm. provenanceId isn't writable. Why have a persistAttr?

src/app/models/AudioEvent.ts
```diff
@@ -56,6 +61,8 @@ export class AudioEvent
   public readonly highFrequencyHertz?: number;
   @bawPersistAttr()
   public readonly isReference?: boolean;
+  @bawPersistAttr()
+  public readonly score?: number;
```
atruskie: score isn't writable

src/app/models/AudioEvent.ts
```diff
@@ -47,6 +50,8 @@ export class AudioEvent
   @bawPersistAttr()
   public readonly audioRecordingId?: Id;
   @bawPersistAttr()
+  public readonly channel?: number;
```
atruskie, line 56: channel isn't writable - but it should be. File a bug and comment out

src/app/models/AudioEventImportFile.ts
```diff
@@ -0,0 +1,68 @@
...
+export class AudioEventImportFile
+  extends AbstractModel<IAudioEventImportFile>
+  implements IAudioEventImportFile
+{
+  public readonly kind = "audio_event_import_file";
+  public readonly id?: Id;
+  public readonly name?: string;
+  @bawPersistAttr()
+  public readonly fileHash?: string;
+  @bawPersistAttr()
```
atruskie: baw persist attrs

src/app/models/AudioEventImportFile.ts
```diff
@@ -0,0 +1,68 @@
...
+  public readonly path?: FilePath;
+  @bawDateTime()
+  public readonly createdAt?: DateTimeTimezone;
+  public readonly committed?: boolean;
+  @bawSubModelCollection<AudioEventImportFile, ImportedAudioEvent>(ImportedAudioEvent)
+  public readonly importedEvents?: ImportedAudioEvent[];
+
+  // form data fields
+  @bawPersistAttr({ supportedFormats: ["formData"], create: true })
+  public readonly analysisJobsItemId?: Id;
```
atruskie: ajii is not writable in the API

src/app/models/AudioEventImportFile.ts
```diff
@@ -0,0 +1,68 @@
...
+  public readonly kind = "audio_event_import_file";
+  public readonly id?: Id;
+  public readonly name?: string;
+  @bawPersistAttr()
+  public readonly fileHash?: string;
+  @bawPersistAttr()
+  public readonly path?: FilePath;
+  @bawDateTime()
+  public readonly createdAt?: DateTimeTimezone;
+  public readonly committed?: boolean;
```
atruskie, line 47: maybe add docs: this field is only present immediately after creation (dry or not)

src/app/models/AudioEventImportFile.ts
```diff
@@ -0,0 +1,68 @@
...
+  @bawSubModelCollection<AudioEventImportFile, ImportedAudioEvent>(ImportedAudioEvent)
+  public readonly importedEvents?: ImportedAudioEvent[];
```
atruskie, line 49: maybe add docs: this field is only present immediately after creation (dry or not)

src/app/services/baw-api/api-common.ts
```diff
@@ -71,6 +71,7 @@ export type Filter = "filter";
 export const emptyParam: Empty = "";
 export const newParam: New = "new";
 export const filterParam: Filter = "filter";
+export const commitParam = "?commit";
```
atruskie: this is not a common pattern, this constant should probably not be in common. Move to AEIF service

src/app/services/baw-api/audio-event-import-file/audio-event-import-file.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    return this.api.create(
+      AudioEventImportFile,
+      endpoint(audioEventImport, emptyParam, emptyParam),
+      (event) => endpoint(audioEventImport, event, emptyParam),
+      model,
+      { params: { commit: true } }
+    );
+  }
+
+  public dryCreate(
```
atruskie, line 85: add doc: "like create but does not result in a saved record. Specifically for testing if a payload passes validation"

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -42,6 +44,9 @@ export interface BawServiceOptions {
   /** If set, requests will include the users authentication token and cookies */
   withCredentials?: boolean;
 
+  /** Unscoped form data to be sent with the request */
```
```suggestion
  /** Additional parameters to merged in with the model payload to be sent with the request */
```

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -386,12 +460,16 @@ export class BawApiService<
...
-    const formData = model.getFormDataOnlyAttributes({ update: true });
+    const jsonData = model.getJsonAttributesForUpdate();
     const body = model.kind
       ? { [model.kind]: jsonData ?? model }
       : jsonData ?? model;
 
+    let formData = model.getFormDataOnlyAttributesForUpdate();
+    if (options.params) {
+      formData = this.addUnscopedParams(formData, options.params);
+    }
```
atruskie, line 496: should probably also handle a json case (in which case just merge with body, or QSP is acceptable too)

src/app/test/fakes/AudioEventImport.ts
```diff
@@ -1,49 +1,19 @@
...
-            })
-        ),
-        ...data,
-    };
+  return {
+    id: modelData.id(),
+    name: modelData.param(),
+    importedEvents: modelData.randomArray(5, 20, () =>
+      generateImportedAudioEvent()
+    ),
```
atruskie: not a prop on this model

src/app/test/fakes/AudioEventImportFile.ts
```diff
@@ -0,0 +1,24 @@
...
+    fileHash: modelData.param(),
+    path: modelData.system.filePath(),
+    createdAt: modelData.dateTime(),
+    committed: modelData.bool(),
+    importedEvents: modelData.randomArray(5, 20, () =>
+      generateImportedAudioEvent()
+    ),
+    analysisJobsItemId: modelData.id(),
+    audioEventImportId: modelData.id(),
+    file: modelData.file(),
```
atruskie: https://github.com/QutEcoacoustics/baw-server/blob/f18d7bf248aae05fd22e62614db33d244ea930a6/app/models/audio_event_import_file.rb#L278-L287

atruskie (changes requested):

src/app/models/AudioEvent.ts
hudson-newey: Related issue: https://github.com/QutEcoacoustics/baw-server/issues/717

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: Failure during a wet run should never happen (unless the wet run fails and the dry run passes).

I can think of one time when this might happen:
If the users access to a project is revoked between the dry run finishing and committing the annotations.

In this case, the user would go through the normal error handling, with a toast message.

To handle this edge case, I can display the errors in the preview table (similar to a dry run)

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: This is 50% type checked (good catch, it can be improved)

If I change the function signature to something like (this code below doesn't actually work, but it just some lookalike pseudo code)

```ts
// (could alternatively use `infer`)
protected hasRecordingError<T extends ImportedAudioEvent>(model: T): "audioRecordingId" in T["errors"]
```

Then using this function on a model would actually restrict the model type.

---

I can also restrict the `error` properties type further by improving the typing of this interface

https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/models/AudioEventImport/ImportedAudioEvent.ts#L11-L13

src/app/components/import-annotations/add-annotations/add-annotations.component.html
hudson-newey: Yes it can, I like the idea

src/app/components/import-annotations/add-annotations/add-annotations.component.html
hudson-newey: Resolved by implementing router links for audio recording column

atruskie:
Also just realized, these are the fields that are parsed:

image


So they should all appear in the grid

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
atruskie: yeah you should add error handling code.

you don't know what has changed since the last dry run

src/app/models/AbstractModel.ts
hudson-newey: Only in tests... I will remove them

atruskie:
import annotations button should be flushed right

atruskie:
Shorten "no associated tags" (empty tags template) to just "none" or "empty". Use a style to indicate it's a special value (maybe italics, and grey color)

atruskie:
Since we support importing multiple files, we should

1. list out each file being imported
2. make it possible to add additional tags to each individual file.

Something like this:

`<index>. <remove button> <name> <error> <add tags> <additional_tags_to_be_added>`

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: Related issue

https://github.com/QutEcoacoustics/workbench-client/issues/2216

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: I've added a `typeaheadCallback` method to the tags.service which we can use in any component

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: image

After reviewing the typing of `error.data`, this comment is incorrect, the problem is that `error.data` can be an array of errors, or a single array.

Because we expect a single error object for this type of response I have added a condition (that should never hit) to remove the `array` type from details

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: I'm a bit hesitant to mutate the error message on the base response model, so I wouldn't want to modify/mutate the error message that is returned from the `dryCreate` method.

---

What I'm thinking from your proposal is:

A new method on the `audio-event-import-file.service` (e.g. `humanize-errors` that takes a `File` object, and a `BawApiError` model.

It will change all of the `file` keys on the `BawApiError` details to use the `File.name` instead

hudson-newey: Discussed offline, we shouldn't be modifying the BawApiError message at all, this should be a purely templating change.

hudson-newey: image

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: Following a later comment, and an offline discussion, we've decided that you should be able to add additional tags to each individual file, along with being able to add additional tags to every file in the current queue

src/app/components/import-annotations/add-annotations/add-annotations.component.ts
hudson-newey: image

The page component now uses an array of `QueuedFile[]` to remove these arrays

src/app/services/baw-api/audio-event-import-file/audio-event-import-file.service.spec.ts
atruskie, line 1: no tests?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -316,31 +383,56 @@ export class BawApiService<
...
+      if (formData) {
+        formData = this.addUnscopedParams(formData, options.params);
+      } else {
+        const url = new URL(path);
+        Object.entries(options.params).forEach(([key, value]) => {
+          url.searchParams.append(key, value);
+        });
+
+        path = url.href;
+      }
```
atruskie: why not just append unscoped vars to the jsonData body?

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -386,12 +480,32 @@ export class BawApiService<
...
+      // to send one.
+      if (formData) {
+        formData = this.addUnscopedParams(formData, options.params);
+      } else {
+        const url = new URL(path);
+        Object.entries(options.params).forEach(([key, value]) => {
+          url.searchParams.append(key, value);
+        });
+
+        path = url.href;
```
atruskie: as above

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -691,6 +807,17 @@ export class BawApiService<
...
     };
   }
+
+  private addUnscopedParams(
+    data: FormData,
+    params: BawServiceOptions["params"]
+  ): FormData {
+    for (const [key, value] of Object.entries(params)) {
+      data.append(key, value);
+    }
```
atruskie: seems we just special cased adding arrays to formdata, but we haven't done it here. is that an issue?

atruskie (approved):

src/app/services/baw-api/baw-api.service.ts
hudson-newey: Really good catch (thank you).

TypeScript should have caught this so I think an underlying issue here is that the `params` property on the options is a `Record<any, any>` so I didn't get any TypeScript errors to warn me about this.

image

PR #2235, opened 2025-03-20
Prompt users to opt-in to communications

### LINKED ISSUE #2172: Add an Opt-in For Communications
It would be good to have an opt-in to agree to email communications when registering as a user. This would allow the admin or platform manager to send the users enquiries about how they use the platform.

hudson-newey:
# Prompt users to opt-in to communications

## Changes

- Prompts users who have not been asked about communications to either opt-in or opt-out of communications (dismissing the toast will be recorded as an non-response decision)
- Adds custom ng-bootstrap toast service, provider, and `<baw-toast>` helper
- Removes ngx-toastr as a dependency
- Stop Dependabot being able to bump major angular versions
- Custom formly checkbox now has the checkbox inline with the label (the label was previously on its own separate line)

## Issues

Fixes: #2172

## Visual Changes

image

_New toast for communication opt-in and new success toast_

---

image

_Register page with communications opt-in checkbox_

---

image

_New error toasts_

## Final Checklist


src/app/components/security/pages/login/login.component.ts
```diff
@@ -38,21 +42,48 @@ export const loginMenuItemActions = [
       [recaptchaSeed]="recaptchaSeed"
       (onSubmit)="submit($event)"
     ></baw-form>
+
+    <baw-toast
+      #communicationsToast
+      [title]="'Subscribe to communications'"
```
atruskie:
```suggestion
      [title]="'Subscribe to email updates'"
```

src/app/components/security/pages/login/login.component.ts
```diff
@@ -38,21 +42,48 @@ export const loginMenuItemActions = [
...
+      [options]="{ autoHide: false }"
+    >
+      <ng-template>
+        <p>Would you like to subscribe to communications?</p>
```
atruskie:
```suggestion
        <p>Can we email you with updates to our platform?</p>
```

src/app/components/security/pages/login/login.component.ts
```diff
@@ -68,6 +99,20 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
           this.notifications.error("Unable to redirect back to previous page");
         }
       },
+      onSuccess: () => {
```
atruskie, line 102: ok i think i'd comment that `onSuccess` happens before `redirectUser` and the toast persists after the navigation event

but maybe the arguments for the `FormTemplate` constructor should be documented as well

src/app/components/security/pages/login/login.component.ts
```diff
@@ -68,6 +99,20 @@ class LoginComponent extends FormTemplate<LoginDetails> implements OnInit {
...
+        // When logging in we check to see if the user has been asked to opt in
+        // to communications.
+        // If they have never been asked, we show a toast asking if they would
+        // like to opt in.
+        //
+        // Although we ask the user to opt in/out of communications are
+        // registration, we have a backlog of users who have not been asked
+        // before.
+        const hasOptedIn = this.session.currentUser?.contactable;
+        if (hasOptedIn === undefined || hasOptedIn === UserConcent.unasked) {
```
atruskie: I'd attach this predicate to the model. Also nit: I typically reserve the `has` prefix for predicates (boolean only) values

src/app/components/shared/toast-provider/toast-provider.component.scss
```diff
@@ -0,0 +1,27 @@
+@import "bootstrap";
+
+:host {
+  position: fixed;
+  top: 0;
+  left: 50%;
+  transform: translateX(-50%);
```
atruskie: the transform is not ideal IMO (because it automatically overflows)

maybe:

```
{
    position: fixed;
    width: fit-content;
    top: 10px;
    left: 0px;
    right: 0px;
    height: fit-content;
    margin: auto;
}
```

src/app/components/shared/toast/toast.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  ToastVariant,
+} from "@services/toasts/toasts.service";
+
+@Component({
+  selector: "baw-toast",
+  template: "<ng-content></ng-content>",
+  standalone: true,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ToastComponent {
```
atruskie, line 30: i'm a little lost on the difference between this and toast-provider component.

I'm pretty sure toast provider is the component that hosts a toast at the top of the page.

And it can optionally accept as custom template.

And that template would be captured/created in another template, with this component?

So the docs for this component would be something like: _Creates a custom notification template which can be passed to the notification service_?

src/app/models/User.ts
```diff
@@ -93,6 +95,8 @@ export class User extends AbstractModel<IUser> implements IUser {
   public readonly isConfirmed?: boolean;
   public readonly rolesMask?: number;
   public readonly rolesMaskNames?: string[];
+  @bawPersistAttr({ update: true })
```
atruskie, line 100: if i remember our DSL, update: true negates create: true.

It should be writeable on creation, though it will probably cause an error. 

otherwise leave a note and a todo: when api gets proper users api, then we can fix this

src/app/components/security/pages/login/login.component.ts
atruskie: you already have a helper on the session service?

atruskie (approved):
We've talked offline about the churn this PR involves. There's no better way to do it with our current architecture, but I'm wondering if we can't improve and DRY out our service injection in another PR.

Otherwise, great work, especially removing that extra dependency.

src/app/components/shared/toast/toast.component.ts
hudson-newey: Yes, this component is just an ergonomic way of creating a toast/notification template that can be passed to the notification service or invoked using the components `open()` and `close()` methods.

The benefit of this component is that its easy to create toasts with custom templates.

atruskie: cool, add doc string

hudson-newey: image

src/app/models/User.ts
hudson-newey: image

src/app/components/security/pages/login/login.component.ts
hudson-newey: image

PR #2247, opened 2025-03-27
Add license information to projects

### LINKED ISSUE #1830: Add a license chooser and license badges to projects (and their derivatives)

We need to be able to clearly show the license which applies to data.

https://github.com/QutEcoacoustics/baw-server/issues/289 will implement the storage of the information. 

Then display the current license in a badge, and allow users to edit the current license in the project edit form.

hudson-newey:
# Add license information to projects

## Changes

- Adds a license picker to the new and edit projects page
- Adds a formly license input component
- Adds license badges to the project/regions list page
- Adds https://github.com/sindresorhus/spdx-license-list as a dependency
- Adds a license widget to the following pages
	- Project details
	- Region details
	- Site/Point details
	- Audio recording list & details
	- Download audio recording page
- Improves typing for `TypeaheadInputComponent`
- Bumps `lib.d.ts` version from ES2020 -> ES2022
- Fixes a bug where using the `withUnsubscribe()`, or `PageComponent` would cause type checking to be disabled. This fixed > 30 type related bugs that were present in production.
- Typeahead inputs now show suggestions when focused (can be disabled with the `queryOnFocus` attribute)
	- **Important!:** I have not disabled `queryOnFocus` for any typeaheads on the website, meaning that if the user clicks a typeahead on the annotation search page, it will show a list of 10 items
- Changes `ngTooltip="tooltip content"` references to `[ngbTooltip]="'tooltip content'"`. This is so that the Angular LSP can correctly recognise that `[ngbTooltip]` is a directive, and prompt the developer to import the required directive if it is not imported. Without the square brackets surrounding the attribute, the Angular LSP will not complain if the directive is not imported.
- Moves the "Update search filters" button in the annotation search modal to the right hand side, to be consistent with the rest of the website
- Migrates `styleUrls` to `styleUrl`
- Fixes broken `/statistics` route (I forgot to declare the `StatisticsPageComponent` in the `appPageComponents` array when I migrated to standalone)

## Issues

Fixes: #1830

## Visual Changes

image

_Project edit page with new "License" field_

---

image

_Project edit page when searching for a license_

Notice that if the user hasn't started typing,

---

image

_License viewer (via license input component)_

---

image

_New license widget on the site details page (showing a site with multiple projects)_

---

image

_New license badges next to projects on the project list page_

Dotted underline is to indicate there is a tooltip on hover

---

image

_Example of new typeahead that shows suggestions on focus (before the user starts typing) on the annotation search page. (this can be disabled per typeahead using the `queryOnFocus` option_

## Final Checklist


src/app/components/shared/formly/license-input.component.ts
```diff
@@ -0,0 +1,108 @@
...
+          [formControl]="asFormControl(formControl)"
+          [formlyAttributes]="field"
+          (ngModelChange)="updateSelectedLicense($event)"
+        >
+          @for (license of availableLicenses() | keyvalue; track license) {
+          <option [value]="license.key">
+            {{ license.value.name }}
+          </option>
+          }
+        </select>
```
atruskie: would it be easier to use a typeahead here?

atruskie (changes requested):
General UI changes:

- next to the license picker input, we should have a description.

```
 Picking a license applies to all data, metadata, and analysis artefacts in this project. It does not change your ownership.

Not choosing a license is equivalent to not saying anything. Your copyright is still in effect. See https://choosealicense.com/no-permission/ 

If you intend to make your project public, it is *strongly* recommended you choose a license.  You can see https://choosealicense.com/ for guidance on choosing.
```

Choose a license is a great website but inappropriate for data. I wish a data equivalent existed. This resource is good but it's too technical: https://opendefinition.org/licenses/


- We should only include licenses designed for data, rather than code
- we don't need the legal advice banner, we're not suggesting anything
- I don't know why we have the license copy in a `<pre>`. Just emit it.
- are the licenses clickable in the widget?
- we should make "Unknown" obviously special. Italicize, muted text colour.
- love the license tags on the project list page, but we can simplify
    - don't bother showing ones with "unknown"
    - remove the "license: " prefix on the tag (it can stay in a tooltip if someone is interested in what it means)

I started reviewing the code, but there's enough suggested changes here that I think we'll checkpoint here.

src/app/components/shared/formly/license-input.component.ts
hudson-newey: I did consider it. The reason why I chose against using a typeahead is because it requires you to know what license you are looking for.

And regular select boxes also support typing to go to options

Edit: The more I think about it, a typeahead works fine

hudson-newey:
Should we change "Unknown" license to "No license"?

And add a hyperlink to <https://choosealicense.com/no-permission/>?

hudson-newey:
> I don't know why we have the license copy in a \<pre\>. Just emit it.

The license copy is a single line that contains `\n` and `\t` for new lines and spaces.

If we wrapped the license content in a paragraph (`<p>`) element, the entire license would be emitted on one line.

So we can't emit the license directly in a `<p>` element without modification.

The only other solutions would be either:

- Do string replacement for `\n` characters and replace them with a `<br />` element, and then append it to the DOM as an unsafe (unsanitized) string
	- This is not semantically correct HTML
- Create semantically correct HTML by creating a parser to create paragraph elements by parsing the license content. Then (again) append the string to the DOM as an unsafe, unsanitized string.

(both of these would be vulnerable to supply chain attacks)

---

Which lead me to use a `<pre>` element. (it is the path of least resistance)

If you don't like the styling (e.g. the monospace text) I can easily change it to look like normal paragraph elements. I can even inherit the `p` scss styling.

hudson-newey:
> are the licenses clickable in the widget?

No. I originally had the license widget text clickable, but it required loading the 5MB license list to get the hyperlink.

src/app/components/shared/formly/license-input.component.ts
atruskie: I also thought about this: it should be less needed with just the data-oriented licenses in the list

atruskie:
> Should we change "Unknown" license to "No license"?
> 
> And add a hyperlink to https://choosealicense.com/no-permission/?

yes

atruskie:
> > are the licenses clickable in the widget?
> 
> No. I originally had the license widget text clickable, but it required loading the 5MB license list to get the hyperlink.

fair enough, leave it for now until we get feedback

hudson-newey:
> We should only include licenses designed for data, rather than code

Have any ideas?

I can't do this with the existing spdx license list (I've searched for others and none of them seem to have a license type field)

<https://www.npmjs.com/package/spdx-license-list>

In the mean time (until the API can provide a list of available licenses), I can maintain a small list of licenses on the client side (e.g. the top 50 licenses used by non-software projects)

I might even be able to do a clickable widget because I assume that the json file will be significantly smaller

atruskie:
> > We should only include licenses designed for data, rather than code
> 
> Have any ideas?
> 
> I can't do this with the existing spdx license list (I've searched for others and none of them seem to have a license type field)
> 
> https://www.npmjs.com/package/spdx-license-list
> 
> In the mean time (until the API can provide a list of available licenses), I can maintain a small list of licenses on the client side (e.g. the top 50 licenses used by non-software projects)
> 
> I might even be able to do a clickable widget because I assume that the json file will be significantly smaller

maybe we can just keep an array (of identifiers) of common ones, and show them a the top (or by default) for the typeahead?

hudson-newey:
I have selected CC licenses as the suggested licenses

image

I have also ordered them from least restrictive to most restrictive.

---

And a preview of what it looks like with all of the others (and the suggested licenses at the top)

image

src/app/components/shared/baw-client/baw-client.component.ts
```diff
@@ -51,6 +51,10 @@ export class BawClientComponent extends withUnsubscribe() implements OnInit {
    */
   @Input() public page: string;
 
+  // TODO: this loading property has never actually been implemented used and
```
atruskie: probably should just aim to deprecate the old client

src/app/components/shared/menu/widgets/license-widget/license-widget.component.ts
```diff
@@ -0,0 +1,132 @@
...
+          if (foundTarget) {
+            targetModel = foundTarget;
+            break;
+          }
+        }
+
+        if (!targetModel) {
+          console.warn("Could not find a supported model");
+          return;
+        }
```
atruskie: refactor to method: `pickModel`

src/app/components/shared/menu/widgets/license-widget/license-widget.component.ts
```diff
@@ -0,0 +1,132 @@
...
+              }
+
+              const licenses = projects.map((project) => project.license);
+              this.setLicenses(licenses);
+            })
+          );
+
+        return firstValueFrom(associatedProjectsRequest);
+      })
+    );
```
atruskie: seems like this should be factors out too

src/app/components/shared/menu/widgets/license-widget/license-widget.component.ts
```diff
@@ -0,0 +1,132 @@
...
+        for (const constructorType of supportedTypes) {
+          const foundTarget = modelValues.find(
+            (model) => model instanceof constructorType
+          );
+
+          if (foundTarget) {
+            targetModel = foundTarget;
+            break;
+          }
+        }
```
atruskie: Suggest:

```typescript
targetModel = models.entries.find(model => supportedTypes .includes(model.constructor));
```

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -91,4 +107,18 @@ export class CardComponent implements OnInit {
...
+    if (this.model.license === null) {
+      this.licenseText = null;
+      return;
+    }
+
+    const isSpdxLicense = await this.licenseService.isSpdxLicense(this.model.license);
+    if (isSpdxLicense) {
+      this.licenseText = this.model.license;
+    } else {
+      this.licenseText = "Custom License";
```
atruskie: second time i've seen this string constant. factor out?

src/app/services/baw-api/project/projects.service.ts
```diff
@@ -83,6 +87,36 @@ export class ProjectsService implements StandardApi<Project> {
...
+      siteIds = Array.from(model.siteIds);
+    } else if (model instanceof Site) {
+      siteIds = [model.id];
+    } else if (model instanceof AudioRecording) {
+      siteIds = [model.siteId];
+    }
+
+    if (!siteIds) {
+      console.warn("No siteIds found for model", model);
+      return;
```
atruskie: should this return an empty observable? one that is finished?

src/app/services/licenses/licenses.service.ts
```diff
@@ -0,0 +1,70 @@
...
+
+@Injectable({ providedIn: "root" })
+export class LicensesService {
+  public async availableLicenses(): Promise<Record<string, SpdxLicense>> {
+    // Because the spdx license list is quite large (5MB), importing the
+    // license list into the client bundle would almost double the clients
+    // bundle size for something that is only needed on one page.
+    // To minimize the bundle size, I dynamically import the spdx license
+    // list so that the 5MB is only loaded when the user needs.
+    const licenses = await import("node_modules/spdx-license-list/full");
```
atruskie: even 5MB delayed is still very bad.

We have plans to make this better. TODO with a link to sever license API?

atruskie (approved):
KK, looks like we don't support custom licenses with this change set?

Add another issue to do later?

Great work, can't wait to get this shipped

hudson-newey:
> KK, looks like we don't support custom licenses with this change set?
> Add another issue to do later?

Related issue: https://github.com/QutEcoacoustics/workbench-client/issues/2255

src/app/services/baw-api/project/projects.service.ts
hudson-newey: Yeah it should. I'm quite concerned that TypeScript didn't yell at me for this.

This signature of the function doesn't allow `undefined` to be returned

image

src/app/services/licenses/licenses.service.ts
hudson-newey: image

src/app/services/baw-api/project/projects.service.ts
hudson-newey: ~~image~~

Edit: I can actually clean this up a bit further by not having two different rxjs observables

Edit: Something like this

image

I don't think it's more optimized because the `siteIds` isn't an observable (it might actually be slightly slower because it has to create the filter object), but it simplifies it into one observable

hudson-newey:
CI is failing because the bundle size was increased by 2MB as a result of the license list being added.

However, we deemed that while bad, it is not horrible because the 2MB license JSON file is only lazy loaded when the user navigates to the project edit page. Meaning that users typically will not download this 2MB file.

I'll temporarily increase the budget by 2MB and then decrease it by 2MB once https://github.com/QutEcoacoustics/baw-server/issues/750 is complete.

We're planning on merging this PR with a sub-optimal implementation so that we can ship this much desired feature.

PR #2267, opened 2025-04-10
Upgrade to application builder

### LINKED ISSUE #2265: Migrate to application builder
We previously migrated to the `browser-esbuild` builder which brought us reduced compile times.

However, the `application` builder is now stable, and we can use it over the `browser-esbuild` builder.

https://angular.dev/tools/cli/build-system-migration#for-new-applications

This builder provides a lot of benefits such as

- HMR support
- Ability to suppress sass deprecation warnings (see #2260 "Problems" as to why this might be useful)
- SSR dev environment. Meaning that any SSR bugs are more likely to be caught during dev

hudson-newey:
# Upgrade to application builder

## Changes

- Runs SSR during dev so that dev environments are more aligned with production environments (should catch more bugs during development)
- Enables CSS HMR in dev environments (quicker development)
- Migrate to using `fetch` for `HttpClient` requests (needed for SSR fetching) (we don't support IE, so using XHR is just archaic)
- Explicitly specify the `development.ecosounds.org` CNAME in the `angular.json` so that Vite (and new application builder) can correctly perform automatic dev server reloads
- Fixes a bug where environment.json would always fail to fetch in SSR (environment.json is now dynamically imported)

_Notes: I was a bit worried about the compatibility with the web components and the `assets/` directory, but I have manually tested the compatibility and it all seems to work great!_

## Changes I want scrutiny on

The `config.service` now uses a dynamic esm import for the environment.json when in SSR.

https://github.com/QutEcoacoustics/workbench-client/blob/application-builder/src/app/services/config/config.service.ts#L56-L62

image

This fixes a bug where environment.json would fail to load in SSR.

## Issues

Fixes: #2265 

## Final Checklist


src/app/components/shared/toast/toast.component.ts
```diff
@@ -24,7 +24,6 @@ import {
 @Component({
   selector: "baw-toast",
   template: "<ng-content></ng-content>",
-  standalone: true,
```
atruskie, line 27: This seems strange - didn't you just make everything standalone?

atruskie (approved):

src/app/components/shared/toast/toast.component.ts
hudson-newey: Yes. This seems to be a straggler that got through the migration because this component doesn't have any imported dependencies.

---

Note: That `standalone: true` has been the default from Angular 19, and you now have to explicitly opt-out of standalone components with `standalone: false`.

This component is still standalone, but doesn't explicitly state that it is standalone.

---

We have an eslint rule enabled that will throw an error if a component is marked with `standalone: false`, advising us that we should remove the standalone override.

I was hoping to find a similar rule that would throw an error if standalone was explicitly defined (even if it had no effect such as `standalone: true`).

However, doing a quick search in the angular-eslint repo, I can't find any rules that we could enable to throw errors for components that explicitly state `standalone: true`

https://github.com/search?q=repo%3Aangular-eslint%2Fangular-eslint%20standalone&type=code

---

I did perform a search over our code base to see if I missed any others, and this seems to be the only one that I missed.

---

Tl;DR: This change has no effect, it's purely stylistic to be consistent with the rest of the code base.

`standalone: true` has been the default since Angular 19, so explicitly stating it here has no effect.

PR #2292, opened 2025-05-15
Migrate to server routing

### LINKED ISSUE #2290: Migrate to server routing
Angular 19 now has support for hybrid rendering / server routing.

<https://angular.dev/guide/hybrid-rendering>

---

Migrating to server routing also allows us to use different rendering strategies per page. At the moment all of our initial page loads are preformed by SSR.

Migrating the server routing makes this configurable either through the `PageInfo` or the `StrongRoute` declarations.

hudson-newey:
# Migrate to server routing

This enables [hybrid rendering](https://angular.dev/guide/hybrid-rendering) and allows us to specify what rendering strategy we should use on a per-page basis through the `PageInfo`.

---

We might want to expose this through the `StrongRoute` instead of the `PageInfo` because `ServerRoute`'s support a lot of other cool features such as specific headers for certain routes, status codes, etc....

https://angular.dev/guide/hybrid-rendering#setting-headers-and-status-codes

Additionally, putting all of this information on the `PageInfo` leaks server routing information to the client which we might not want.

## Changes

- Migrates to the app server router.
- Migrate the home page to use pre-rendering (ssg) and defer the loading of "example project" cards to the client.
	- this was done more or less to test that the server routing / `renderMode` was working correctly. Let me know if you want this reverted, but I think it's a nice improvement since it improves the load times, and the "example project" cards are typically off screen on first load, so by the time the user scrolls to them, they would have loaded in by csr.
- Replaces `REQUEST` and `RESPONSE` injection tokens with Angular native SSR tokens.
- Do not provide `APP_CONFIG` provider twice (we were providing it through the express provides and the `config.service`.
- Fixes console errors being thrown when page title route models could not be accessed due to un-authenticated sessions.
- Do not show "bad" website status warnings if the website status has not been fetched.

## Issues

Fixes: #2290

## Final Checklist


src/app/components/projects/pages/new/new.component.ts
```diff
@@ -62,6 +63,7 @@ ProjectNewComponent.linkToRoute({
   category: projectsCategory,
   pageRoute: newProjectMenuItem,
   menus: { actions: List(projectsMenuItemActions) },
+  renderMode: RenderMode.Client,
```
atruskie: Add a comment explaining why this route in particular is client rendering?

src/app/helpers/page/pageInfo.ts
```diff
@@ -28,6 +27,12 @@ export interface IPageInfo extends Data {
   fullscreen?: boolean;
   resolvers?: ResolverList;
   menus?: Menus;
+
+  /**
+   * Describes where the page will be rendered.
+   * @default RenderMode.Server
```
atruskie, line 39: add a comment about how our linkToRoute(?) function plumbs this into the route data for the page

src/server.new.ts
atruskie, line 1: i assume this is WIP and we'll just keep one server.ts?

src/app/components/projects/pages/new/new.component.ts
hudson-newey: Reverted

hudson-newey:
I removed `ng-mocks` from the `home.component.spec` tests and they work in isolation when using the `ng test --include` command.

However, even though the `home.component.spec` fixture doesn't use any mocked components, it still breaks when running the entire test suite.

I suspect (I have not checked) this is because one of our other tests use mocked components. Meaning that if we use `ng-mocks` anywhere in our test suite, we immediately lose the ability to test `@defer` blocks.

src/app/test/fakes/PageInfo.ts
```diff
@@ -1,14 +1,20 @@
...
     fullscreen: modelData.datatype.boolean(),
     category: homeCategory,
     pageRoute: homeMenuItem,
     resolvers: {},
     menus: {},
+    renderMode: modelData.helpers.arrayElement([
+      RenderMode.Server,
+      RenderMode.Client,
+      RenderMode.Prerender,
+    ]),
```
atruskie: probably don't want to randomize this, it could create flaky tests that render different ways? It depends how often we use this pageinfo generator/helper...

atruskie (approved):

src/app/test/fakes/PageInfo.ts
hudson-newey: We use the `PageInfo` generator/helper a lot.

Most of our Testbed's don't actually have a router, and those that do provide their own router, meaning that the mocked routes don't use `StrongRoute`'s or `PageInfo` at all.

e.g.
https://github.com/QutEcoacoustics/workbench-client/blob/bfee1a82c92a8f56d2b192dbb42cdb55191c1545/src/app/directives/strongRoute/strong-route-active.directive.spec.ts#L38-L48

Additionally, all of our tests are client-side only.\
We don't have any ssr or e2e tests.
Meaning that it's not possible to provide `ServerRoute`'s (which is the type of route that takes a `RenderMode`. Client side Route's cannot take a `RenderMode`).

hudson-newey: image

PR #2303, opened 2025-05-19
Improve annotation import UI

### LINKED ISSUE #2297: UI improvements to annotation imports
Page should navigate to file import screen after creating annotation import

File import screen should clearly show that the files have not been committed yet until the "Import" button is clicked (perhaps a warning banner / alert if files have not been comitted)

Add a "verify annotations" button to the batch import screen

Batch import annotations should be moved to the projects page

Simplify menu item / page title to Import Annotations

We should show a prompt to look at the table for errors

It is not evident that the table is horizontally scrollable

---

Parent issue: #2238

hudson-newey:
# Improve annotation import UI

## Changes

- Fixes directory structure of annotation import pages. They were previously located directly under the `components/import_annotations/` directory when our convention is to place them under the `components/import_annotations/pages` directory
- Add warning to the menu widgets if files have not been committed
- Add menu widgets explaining how to fix errors
	- Add card explaining how to fix event errors which contains a hint that more descriptive errors can be found in the table.
	- Card explaining duplicate file errors.
	- Unsupported format error card
	- Internal server error card
- Change branding of "Batch Import Annotations" to "Import Annotations"
	- Updates page title
	- Updates breadcrumbs
	- Updates menu item text
	- Updates URL slug from `batch_annotations` to `import_annotations`
- Add gradient if ngx-datatable's are horizontally scrollable
- After creating an annotation import, users are now navigated to the file upload page instead of the "details" page that shows the files and events associated with the import.
- Bumps the `@ecoacoustics/web-component` version
	- Adds support for MacOS
	- Fixes indicator line going offscreen
	- Fixes indicator line moving if audio does not play
	- Some performance improvements
- Adds a "compact" configuration for the ngx-datatable
- Moves the annotation import menu link to the project details page
- `baw-error-card` component now fills its `:host` container. Meaning that changing the size of the `baw-error-card` selector now changes the size of the errors contained.

### Bug fixes

- Fixes the "Download Annotations" menu item being visible even if the user is not logged in. (previously visible on the project, region, site, and point detail pages)

## Issues

Fixes: #2297

## Visual Changes

image

_New "Instructions" widget that is displayed before any action is taken_

---

image

_Information card explaining that a valid file has not been committed to the database until the "Import Annotations" button is clicked._

---

image

_Information cards explaining how to fix errors in the uploaded datasets. Note that errors now fill up the entire cell._

## Final Checklist


src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
+<h5>
+  <fa-icon class="pe-1" [icon]="['fas', 'circle-info']"></fa-icon>Instructions
+</h5>
+
+<p>
+  You can use this interface to upload annotations from human or machine
+  generated datasets.
```
atruskie:
```suggestion
  This page allows you to import annotations from human or machine
  generated datasets.
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+
+@if (hasUncommittedFiles()) {
+  <div class="uncommitted-file-warning alert alert-warning">
+    <h6>
+      <fa-icon class="pe-1" [icon]="['fas', 'triangle-exclamation']"></fa-icon>
+      Warning
+    </h6>
+
+    <strong>Your files have not been saved</strong>. Please review your uploaded
+    files, and click "Import Annotations" to save your annotations.
```
atruskie:
```suggestion
    <strong>Your files have not been imported yet</strong>. 
    Check the preview table and make sure we've
    correctly parsed your data.
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+  <fa-icon class="pe-1" [icon]="['fas', 'circle-info']"></fa-icon>Instructions
+</h5>
+
+<p>
+  You can use this interface to upload annotations from human or machine
+  generated datasets.
+</p>
+
+@if (hasUncommittedFiles()) {
+  <div class="uncommitted-file-warning alert alert-warning">
```
atruskie: i think this should be an info rather than a warning

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+}
+
+@if (annotationImport.importErrors().length > 0) {
+  <section class="error-section widget-section p-3 border-top">
+    <h5>How to handle errors</h5>
+
+    <div class="error-cards">
+      @if (hasEventErrors()) {
+        <div class="widget-card">
+          <h6 class="callout callout-danger">Event errors</h6>
```
atruskie:
```suggestion
          <h6 class="callout callout-danger">Individual errors</h6>
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+
+          <p>
+            Event errors are broken down per-row in the "Identified Events"
+            table.
```
atruskie:
```suggestion
            Each detected annotation is shown in the preview table along with any detected errors
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+          </p>
+
+          <p>
+            Errors can arise due to invalid values, incorrect data types, or
+            referencing audio recordings that you don't have permission to
+            access.
```
atruskie: This bit is a bit redundant.

```suggestion
            You may have to change your files to correct
            the errors so we can understand your data.
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+
+          <p>
+            Datasets that contain the same content cannot be uploaded twice.
+          </p>
+
+          <p>
+            This includes other accounts uploading the same dataset. Please
+            ensure that the conflicting dataset has not previously been uploaded
+            by another person.
+          </p>
```
atruskie, line 56: add contact us cta for help

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+        <hr>
+      }
+
+      @if (hasServerError()) {
+        <div class="widget-card">
+          <h6 class="callout callout-danger">Internal Server Error</h6>
+
+          <p>
+            We encountered an unrecoverable error while processing your file. We
+            have been notified and will investigate this failure.
```
atruskie:
```suggestion
            We encountered an error while processing your file. We
            have been notified and will investigate this failure.
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+          </p>
+
+          <p>
+            This error can sometime arise if the structure of your dataset is
+            invalid or corrupted.
```
atruskie:
```suggestion
            This error can occur if the structure of your dataset is
            invalid or corrupted.
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
```diff
@@ -0,0 +1,104 @@
...
+          </p>
+
+          <p>
+            If you have any further information, you can
+            <a [strongRoute]="contactUs.route">{{ contactUs.label }}</a> to help
+            resolve this bug faster.
+          </p>
```
atruskie: we have a contact us cta like three times here, common template?

src/app/directives/datatable/datatableConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+    protected datatableRef: ElementRef
+  ) {}
+
+  public abstract headerHeight: InputSignal<number>;
+  public abstract footerHeight: InputSignal<number>;
+  public abstract summaryHeight: InputSignal<number>;
+  public abstract rowHeight: InputSignal<DatatableComponent["rowHeight"]>;
+
+  public ngOnInit(): void {
+    this.datatableRef.nativeElement.classList.add("bootstrap");
```
atruskie: is this common to both implementations of the config?

src/app/directives/datatable/datatableConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    this.datatable.footerHeight = this.footerHeight();
+    this.datatable.headerHeight = this.headerHeight();
+    this.datatable.summaryHeight = this.summaryHeight();
+
+    this.datatable.rowHeight = this.rowHeight();
```
atruskie: bit useless  having a signal if we can only ever update the value once. Don't we want to listen to changes to these signals?

atruskie (changes requested):
add import annotations menu item to project menu page as well please (not withstanding scope) so users can find the item easily

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: Nit: Why "Individual" instead of "Event".

I remember thinking about this terminology and used "Event" because it indicates that this error helper applies to the errors found in the "Identified Events" table.

atruskie: Yeah so I was actually anti "Event" because we use annotation consistently. Individual sounded clearer, but you could sway me on s/event/annotation/

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: I like the idea of reducing the amount of text shown.

Although I'm a bit concerned about being vague.

---

In my mind, these error widgets are attempting to assist users in fixing their data.

e.g. Note that if you don't have access to an audio recording, the error that you will receive will be "audio recording not found", which doesn't provide much help with fixing the error.

---

After writing this out, I think the root cause is that some of the errors returned by the api are a bit vague.

E.g. we should return different error messages for unauthorized access and incorrect data types (e.g. strings instead of `int`'s)

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: That actually makes a lot of sense.

If I was to replace this with "Annotation Errors", I'm thinking that I should change the heading for the "Identified Events" table to "Identified Annotations"?

src/app/directives/datatable/datatableConfig.ts
hudson-newey: Yes, this should be consistent across all of our datatable configs because it applies the [bootstrap theming](https://swimlane.github.io/ngx-datatable/bootstrap) to the ngx-datatable.

Now that you bring it up, this should actually be refactored out of this directive and into the `_ngx-datatable.scss` styling.

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: image

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: image

Opinion?

Should it still say warning? Should it still have an exclamation mark?

---

Here's an alternative that I mocked up that has the "Information" heading, and a `circle-exclamation` icon.

image

src/app/directives/datatable/datatableConfig.ts
hudson-newey: I'm checking now, but I'm pretty sure that invoking the signal creates a subscription.

I _think_ if you follow this getter, it attaches itself to the active consumer (the component).

https://github.com/angular/angular/blob/main/packages/core/primitives/signals/src/signal.ts#L59

And setting the signal also triggers active consumers to update

https://github.com/angular/angular/blob/main/packages/core/primitives/signals/src/signal.ts#L134

hudson-newey: For a bit of fun, here's my (simplified) mental model of siganls.

This might be wrong, but I find it a bit interesting if I can re-create a concept like signals from scratch;

```ts
class Component {
	public async update() {}
}

const signalGet = function(subscriberClosure, value) {
	// "this" here refers to the "Component" or nearest prototype.
	subscriberClosure.add(this);
	return value;
}

function signal<T>() {
	const subscriberClosure = new Set([]);
	const value: T | undefined = undefined;

	const updateClosure = () => {
		for await (const subscriber of subscriberClosure) {
			subscriber.update();
		}
	};

	const signalSet = (newValue: T) => {
		value = newValue;
	};

	const signalObj = function() {
		return signalGet(subscriberClosure, value);
	};
	signalObj.set = signalSet;
	signalObj.update = updateClosure;

	return signalObj;
}
```

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: Relevant commit:

[`d652cee` (#2303)](https://github.com/QutEcoacoustics/workbench-client/pull/2303/commits/d652ceeed094d5df5ed924f9d74e436fe6182b97)

src/app/directives/datatable/datatableConfig.ts
hudson-newey: image

src/app/directives/datatable/datatableConfig.ts
hudson-newey: I have created a small test to see if the signals can be updated after creation.

https://github.com/user-attachments/assets/30252dc2-bfc0-4443-9236-12ac52ae39c8

image
image
image

---

I suspect one of two things is happening here:

- What I think is happening: Evaluating the signals value attaches it to the components `render3` change detection.
- or/and updating the `[footerHeight]` input is triggering the underlying `ngx-datatable` `input`.

My understanding of signals is that when they are invoked inside a `REACTIVE_NODE`, the reactive node's CD is automatically subscribed to the signal.

e.g.

```ts
class Component {
	public async update() {
		// do update
	}
}

const signalGetter = function(subscribers, value) {
	// "this" here refers to the "Component" or the nearest prototype
	subscribers.add(this.update);
	return value;
}

function signal<T>(initialValue: T) {
	const value: T = initialValue;
	const subscribers = new Set<(newValue: T) => void>([]);

	const notifySubscribers = (newValue) => {
		for (const subscriber of subscribers) {
			subscriber(newValue);
		}
	};

	const signalObj = function() {
		return signalGetter(subscribers, value);
	}
	signalObj.set = () => {
		notifySubscribers();
	};
	signalObj.subscribe = (callback: (value: T) => void)) {
		this.subscribers.add(callback);
	};

	return signalObj;
}
```

---

I'll obviously do some further investigation, but I think my mental model checks out.

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
atruskie: Don't we have stock components for these alerts? 

Maybe we don't need the _Information_/_Warning_ title at all. The first sentence of the paragraph can just be the title.

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
atruskie: Good point about also being consistent with the table heading - though i was leaning to renaming it something including _preview_.

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
atruskie: >e.g. Note that if you don't have access to an audio recording, the error that you will receive will be "audio recording not found", which doesn't provide much help with fixing the error.

It _should_ actually tell you don't have permission - i specifically made the error distinct.

> Although I'm a bit concerned about being vague.

Fair, but then we should provide better error messages, customised to the errors we're seeing.

A basket list of possible errors is just text people don't want to read

src/app/directives/datatable/datatableConfig.ts
atruskie: I don't have the mental bandwidth for this right now, but i'm keen for you to walk me through it later

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: We don't have stock components for these alerts.

But we do have bootstrap styling classes for different alert types.

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: > though i was leaning to renaming it something including preview

How about using "Annotation Preview" for the table's heading

src/app/directives/datatable/datatableConfig.ts
hudson-newey: Discussed offline.

In my demonstration above, I demonstrated that updating the attributes does update the datatable, so there's no need for a subscription.

However, this is not because subscriptions automatically subscribed to components upon invokation.

The demo worked because it was triggering the old `@Input` change detection from the ngx-datatable component.

https://github.com/swimlane/ngx-datatable/blob/fe69d1e96b93bfc93e8e8ca9a1dc179db6593d05/projects/swimlane/ngx-datatable/src/lib/components/datatable.component.ts#L450

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: image

You're correct, the permission failure is distinct per event returned.

However, the errors shown in the sidebar are using the errors returned from the response's `meta.error.info` object.

image

In this case, the error message for the file is "Validation failed".

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
hudson-newey: This looks quite good. I like your suggestion

image

src/app/components/import-annotations/widgets/annotation-import-issue.component.html
atruskie: Yeah and?

You have a function for searching through to get error types

atruskie (approved):
💯

PR #2314, opened 2025-05-30
Add verification grid/annotation search sorting

### LINKED ISSUE #2309: Allow sorting verification interface results
Users would like to specify a sorting / sampling method in the verification interface.

E.g. users would like to be able to sort audio events by `score`

hudson-newey:
# Add verification grid/annotation search sorting

## Changes

- Adds options to sort by
	- Created date (Ascending)
	- Created date (Descending)
	- Score (Ascending)
	- Score (Descending)
- Fixes annotation search form site and point filters
- Adds tests for annotation search parameters
- Paged table template now takes an optional callback that can be used to create sorting conditions for a filter request
- Annotation model in `annotation-event-card.component` now uses a signal
- Media controls `ViewChild` selector in `annotation-event-card.component` is now a static selector
- Fix bug in number qsp serializer where non-numbers would be serialized as `NaN`. Incorrect numbers in qsps are now omitted and log a warning.
- Remove focused test that somehow got onto the main branch (looking for eslint rule now)

## Issues

Fixes: #2309

## Visual Changes

image

_New annotation search with "Sort By" dropdown_

---

image

_Expanded "sort by" dropdown_

## Final Checklist

### DIFF
--- a/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
+++ b/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
@@ -108,7 +108,7 @@
     <select
       id="sort-input"
       class="form-select"
-      placeholder="Upload Date (Ascending)"
+      placeholder="Created Date (Oldest First)"
       [value]="searchParameters.sort ?? 'upload-date-asc'"
       (change)="updateSortBy($event)"
     >
--- a/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
+++ b/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
@@ -102,6 +102,19 @@
       ></baw-typeahead-input>
     </div>
   </div>
+
+  <div cclass="mt-4 mb-4">
+    <label for="sort-input">Sort by</label>
+    <select
+      id="sort-input"
+      class="form-select"
+      [value]="searchParameters.sortBy"
+      (modelChange)="updateSortBy($event)"
+    >
+      <option value="">Upload date/time</option>
+      <option value="score">Score</option>
+    </select>
+  </div>
 </form>
 
 <ng-template #idTypeaheadTemplate let-result="result" let-searchTerm="term">
--- a/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
+++ b/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
@@ -103,14 +103,18 @@
     </div>
   </div>
 
-  <div cclass="mt-4 mb-4">
+  <div class="mt-2 mb-2">
     <label for="sort-input">Sort by</label>
     <select
       id="sort-input"
       class="form-select"
+      placeholder="Upload Date (Ascending)"
       [value]="searchParameters.sortBy"
       (change)="updateSortBy($event)"
     >
+      <option value="upload-date-asc">Upload Date (Ascending)</option>
+      <option value="upload-date-desc">Upload Date (Descending)</option>
+
       <option value="score-asc">Score (Ascending)</option>
       <option value="score-desc">Score (Descending)</option>
     </select>
--- a/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
+++ b/src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
@@ -22,7 +22,7 @@
     label="Project(s)"
     [inputPlaceholder]="project ? project.name : 'All'"
     [searchCallback]="
-      createSearchCallback(projectsApi, 'name', searchParameters.modelFilters())
+      createSearchCallback(projectsApi, 'name', this.routeModelFilters())
     "
     [resultTemplate]="nameTypeaheadTemplate"
     [inputDisabled]="!!project"
@@ -36,7 +36,7 @@
       label="Site(s)"
       [inputPlaceholder]="region ? region.name : 'All'"
       [searchCallback]="
-        createSearchCallback(regionsApi, 'name', searchParameters.modelFilters())
+        createSearchCallback(regionsApi, 'name', this.routeModelFilters())
       "
       [resultTemplate]="nameTypeaheadTemplate"
       [inputDisabled]="!!region"
@@ -49,7 +49,7 @@
     [label]="site ? (site?.isPoint ? 'Point(s)' : 'Site(s)') : 'Point(s)'"
     [inputPlaceholder]="site ? site.name : 'All'"
     [searchCallback]="
-      createSearchCallback(sitesApi, 'name', searchParameters.modelFilters())
+      createSearchCallback(sitesApi, 'name', this.routeModelFilters())
     "
     [resultTemplate]="nameTypeaheadTemplate"
     [inputDisabled]="!!site"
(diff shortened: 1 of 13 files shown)


src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -102,6 +102,23 @@
...
+    <label for="sort-input">Sort by</label>
+    <select
+      id="sort-input"
+      class="form-select"
+      placeholder="Upload Date (Ascending)"
+      [value]="searchParameters.sort ?? 'upload-date-asc'"
+      (change)="updateSortBy($event)"
+    >
+      <option value="upload-date-asc">Upload Date (Ascending)</option>
+      <option value="upload-date-desc">Upload Date (Descending)</option>
```
atruskie: upload date? of what, audio recordings? audio events? audio events don't have an upload date??

perhaps this should be created date

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -115,6 +117,28 @@ export class AnnotationSearchFormComponent implements OnInit {
...
+    } else if (this.region) {
+      return filterModel("regions", this.region);
+    } else if (this.project){
+      return filterModel("projects", this.project);
+    }
+
+    // We should never get to this condition because each annotation search
+    // should at least be scoped under a project.
+    // However, I have included this condition so that if we ever get into this
+    // unexpected state, the client doesn't completely crash and can recover.
```
atruskie: maybe it's different, but the library page is essentially an unscoped annotation search. It would be good if we didn't have to reinvent the wheel for that page

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -183,4 +207,23 @@ export class AnnotationSearchFormComponent implements OnInit {
     this.searchParameters.onlyUnverified = value;
     this.searchParametersChange.emit(this.searchParameters);
   }
+
+  public updateSortBy(event: Event): void {
+    if (!(event.target instanceof HTMLSelectElement)) {
+      console.warn("Attempted to update sort key through non-select element");
+      return;
+    }
```
atruskie, line 226: this is a strangely specific guard. You can coded the select options and you want to guarantee the event is attached to a select that has a value field?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -183,4 +207,23 @@ export class AnnotationSearchFormComponent implements OnInit {
...
+
+    // So that we can minimize the number of query string parameters, we use
+    // upload-date-asc as the default if there is no "sort" query string
+    // parameter.
+    const newValue = event.target.value;
+    if (newValue === "upload-date-asc") {
+      this.searchParameters.sort = null;
+    } else {
+      this.searchParameters.sort = newValue;
+    }
```
atruskie: seems like this could be encapsulated into the searchParameters object?

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -26,13 +27,37 @@ import { hasMany } from "@models/AssociationDecorators";
...
+  },
+  "score-desc": {
+    orderBy: "score",
+    direction: "desc",
+  },
+  "upload-date-asc": {
+    orderBy: "createdAt",
+    direction: "asc",
+  },
+  "upload-date-desc": {
```
atruskie: previous comment mentioned upload date, fix here too

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -259,8 +347,32 @@ export class AnnotationSearchParameters
...
+   * A type guard that can be used to narrow the typing of a user provided
+   * "sort" query string parameter.
+   */
+  private isSortingKey(key: string): key is SortingKey {
+    // We use "hasOwn" instead of "in" here because we don't want to return
+    // "true" if the key is in the prototype chain.
+    // E.g. If we used the "in" operator here, a sorting key of hasOwnProperty
+    // would return true, and would attempt to serialize a function when
+    // creating the filter request body.
+    return Object.hasOwn(sortingOptions, key);
```
atruskie: i wonder if using a `Map` would make this simpler.

atruskie (approved):

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
hudson-newey: Yeah, this guard should never trigger (and it shouldn't), but it's the best way to narrow the `event.target` type.

In TypeScript's dom library, `event.target` is typed as `HTMLElement`, meaning that it doesn't have a `value` property.

By adding this type guard, it narrows the `target` property to `HTMLSelectElement` which does have a `value` property.

---

The hacky way to fix this would be to type cast the `event.target` using something like

```ts
const newValue = (event.target as HTMLSelectElement).value;
```

---

Note that because the type guard is at the top of the function the JIT performance _should_ (untested) be quite good.

src/app/components/annotations/pages/annotationSearchParameters.ts
hudson-newey: I 100% agree.

I originally had this as a Map but I thought that it hurt code readability.

image

---

Or if I explicitly override prettier I can do something like this (which looks a lot nicer IMO)

image

---

I think I'll go with the latter because it's more readable and I agree, there are some benefits from using a `Map` over an object.

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
hudson-newey: Yeah, I purposely left this component open for modification so that when we migrate the library page, we can re-use all of these components/data models by simply removing these comments & warning.

---

After typing this out, I agree and think it might be worthwhile to reword this comment in anticipation to minimize the risk of stale documentation.

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: I like the idea

Upon reflection I think ascending and descending dates makes sense to us as programmers, but I can imagine non-technical people getting confused with this terminoligy.

image

What do you think about "Oldest First" and "Newest First" terminology for dates?

atruskie: yeah oldest and newest are good

src/app/components/annotations/pages/annotationSearchParameters.ts
atruskie: it's unfortunate map literals aren't a thing. meh do what you think is  best

hudson-newey: I think I'll go with the latter map example because it's more readable and I agree, there are some benefits from using a Map over an object.

hudson-newey:
About to merge this PR once CI passes.

Only thing changed since your last review is the following commit

[`f5843ba` (#2314)](https://github.com/QutEcoacoustics/workbench-client/pull/2314/commits/f5843ba62516ce6a008628f0a918aeaa9b8cfb5f)

This commit adds support for getters in the query string parameter data model.

PR #2342, opened 2025-06-23
Admin site config

### LINKED ISSUE #2326: Expose `site_settings` through admin dashboard
We now have site settings recorded in the database.

We should expose these settings to users.

hudson-newey:
# Admin site config

## Changes

- Add site settings to admin dashboard including
	- Slider + number input for `batch_analysis_remote_enqueue_limit`
- `bawDebouncedInput` directive now works for non-text based inputs (e.g. sliders and up/down buttons on number inputs)
- Models + services + mocks for site settings
- Add `baw-range` component which is a number input with synced a synced slider / number input
- `@ng-neat/spectator` now supports signal inputs! I've updated some test beds to use spectator's `spec.setInput` instead of the verbose `spec.fixture.componentInstance.setInput()`

## Discussion

The range/slider input could be a pop-over that's shown when the input field is focused.

In my opinion this would make the UI look a lot nicer, but would also increase the number of actions needed to change the value (potentially bad ux for range input users)

## Issues

Fixes: #2326

## Visual Changes

image

_New `/admin` dashboard with site settings "Batch Analysis Remote Enqueue Limit" number input_

---

image

_Mobile view of admin dashboard_

## Final Checklist


src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  protected enqueueLimit = signal<SiteSetting | null>(null);
+  private enqueueLimitInput = viewChild<RangeComponent>("enqueueLimitInput");
+
+  public ngOnInit(): void {
+    // We use a "list" request here so that if this page is expanded to include
+    // multiple settings, we only have to make one request and then parse out
+    // the individual settings.
+    this.siteSettings
+      .show("batch_analysis_remote_enqueue_limit")
```
atruskie: but you don't actually? `show` is called

atruskie: the logic is sound though, i've got more site settings coming

src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
```diff
@@ -0,0 +1,78 @@
...
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe((initialValue: SiteSetting) => {
+        this.enqueueLimit.set(initialValue);
+        this.enqueueLimitInput().value.set(initialValue.value);
```
atruskie: this seems very specific. Why not just assign a list of models into the field? loop over them in the template?

src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
```diff
@@ -0,0 +1,78 @@
...
+    const value = toNumber(stringValue);
+    if (value === null) {
+      console.error("Failed to updated enqueue limit");
+      return;
+    }
+
+    const newModel = new SiteSetting({
+      ...this.enqueueLimit(),
+      value,
+    });
```
atruskie: ? does your api call not return a SiteSetting model?

src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.siteSettings
+      .show("batch_analysis_remote_enqueue_limit")
+      .pipe(takeUntil(this.unsubscribe))
+      .subscribe((initialValue: SiteSetting) => {
+        this.enqueueLimit.set(initialValue);
+        this.enqueueLimitInput().value.set(initialValue.value);
+      });
+  }
+
+  protected updateEnqueueLimit(stringValue: string) {
```
atruskie: this seems awfully specific.

src/app/components/admin/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,4 @@
+<section>
+  <h2>Instance Settings</h2>
```
atruskie:
```suggestion
  <h2>Site Settings</h2>
```

src/app/components/admin/dashboard/dashboard.component.ts
```diff
@@ -40,10 +41,8 @@ export const adminMenuItemActions = [
 
 @Component({
   selector: "baw-dashboard",
-  template: `
-    <h1>Admin Dashboard</h1>
-    <p>Welcome to the super-secret-site-settings!</p>
```
atruskie, line 45: why do you remove my fun?

src/app/components/shared/input/range/range.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  templateUrl: "./range.component.html",
+  styleUrl: "./range.component.scss",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [FormsModule],
+})
+export class RangeComponent implements ControlValueAccessor {
+  public label = input("");
+  public min = input(0);
+  public max = input(100);
+  public step = input(5);
```
atruskie, line 23: it just occurred to me: the angular compiler is doing magic here to infer the attribute names, right?

src/app/components/shared/input/range/range.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public setDisabledState?(isDisabled: boolean): void {
+    this.disabled.set(isDisabled);
+  }
+
+  protected updateValue(event: Event) {
+    if (!(event.target instanceof HTMLInputElement)) {
+      return;
+    }
```
atruskie: is this guard needed? are you binding the handler elsewhere?

src/app/models/SiteSetting.ts
```diff
@@ -0,0 +1,26 @@
...
+  implements ISiteSetting
+{
+  public readonly kind = "site_setting";
+  public readonly id?: Id;
+  @bawPersistAttr({ update: true })
+  public readonly name: string;
+  @bawPersistAttr({ update: true })
+  public readonly value: number;
+  public readonly description: string;
+
```
atruskie, line 31: not all the attributes on this model are exposed here? intentional?

https://github.com/QutEcoacoustics/baw-server/blob/39dc2871dac431c0cc096691f41a55389ca10b29/app/models/admin/site_setting.rb#L58-L65

src/app/models/SiteSetting.ts
```diff
@@ -0,0 +1,26 @@
...
+  public get viewUrl(): string {
+    throw new Error("Method not implemented.");
```
atruskie: this could just lead to admin dashboard?

src/app/services/baw-api/api-common.ts
```diff
@@ -23,16 +23,30 @@ export type IdParamOptional<T extends AbstractModel> = (
...
+  }
+}
+
+export function idOrName<T extends AbstractModel>(x: IdOrName<T> | Empty): Param {
   if (x === emptyParam || !isInstantiated(x)) {
     return x;
   } else if (isInstantiated(x?.["id"])) {
     return x?.["id"].toString();
+  } else if (isInstantiated(x?.["name"])) {
+    return x?.["name"];
```
atruskie, line 53: i'd probably preference name in this case

atruskie (changes requested):
> The range/slider input could be a pop-over that's shown when the input field is focused.

I'd actually go the other way for UX purposes (people want to see a number, but want to interact with a slider. Imagine a phone where you had to focus an input, then your keyboard pops up, the screen shifts, and then we show a handy dandy slider? That would frustrate me). Regardless, this is an admin UI at the moment, so it will do. If we have use for the component elsewhere, we can improve it.

Not sure why you landed on `Instance Settings`. The api and model are `SiteSettings`.

src/app/components/shared/input/range/range.component.ts
hudson-newey: No. This is just a type guard with more safety

perf is negligible because JIT can branch predict it quite easily

atruskie: perf argument: accepted

semantics were what lead me to query this though.

hudson-newey:
> Not sure why you landed on Instance Settings. The api and model are SiteSettings.

There is already a `Client Settings` page component

https://staging.ecosounds.org/admin/settings

"Instance settings" sounded less ambiguous to me

src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
hudson-newey: I like the idea. I'll probably also order by name so that the UI is semi-stable and doesn't change depending on the returned order

src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
hudson-newey: Resolved above by looping over returned value

src/app/components/admin/dashboard/dashboard.component.ts
hudson-newey: image

src/app/components/shared/input/range/range.component.ts
hudson-newey: I think you might be right

https://github.com/search?q=repo%3Aangular%2Fangular%20SignalBased&type=code

src/app/components/shared/input/range/range.component.ts
hudson-newey: image

I'm going to use type casts from now on. It's more descriptive and easier to read

atruskie:
> > Not sure why you landed on Instance Settings. The api and model are SiteSettings.
> 
> There is already a `Client Settings` page component
> 
> https://staging.ecosounds.org/admin/settings
> 
> "Instance settings" sounded less ambiguous to me

If I were to change the name, I'd use `ServerSettings`

src/app/services/baw-api/api-common.ts
hudson-newey: image

Note that I also changed the name of the function to `nameOrId` to better represent that name is preferred

src/app/components/admin/dashboard/components/site-settings/instance-settings.component.ts
hudson-newey: This is the updated model that is sent as the updated models PUT body.

My api service does return a `SiteSetting`'s model

src/app/components/admin/dashboard/components/site-settings/site-settings.component.html
```diff
@@ -0,0 +1,14 @@
+<form>
+  @for (setting of settings(); track setting) {
+    <baw-range
```
atruskie: not all of these settings will be range inputs. some will be booleans or strings. 

We can deal with that later though

src/app/components/admin/dashboard/components/site-settings/site-settings.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (nameA < nameB) {
+      return -1;
+    }
+
+    if (nameA > nameB) {
+      return 1;
+    }
+
+    return 0;
+  }
```
atruskie:
```suggestion
  private settingsSorter: SortFunction<SiteSetting> = (a, b) => a.name.localeCompare(b.name);
```

src/app/models/SiteSetting.ts
```diff
@@ -19,8 +24,26 @@ export class SiteSetting
...
+
+  public readonly createdAt?: DateTimeTimezone | string;
+  public readonly updatedAt?: DateTimeTimezone | string;
 
   public get viewUrl(): string {
-    throw new Error("Method not implemented.");
+    return adminRoute.format();
+  }
+
+  /** Humanizes the name of the setting to be disabled in the UI */
```
atruskie: disabled?

atruskie (approved):
Ok, good enough to pull in. 

Note: I don't think we've handled potential future fields of a type other than number in a good way. Leave some notes around for future you.

I also note there's no way to unset a value, and to that end we probably should show whether a setting has been customised or not as well (is id blank?)

src/app/components/admin/dashboard/components/site-settings/site-settings.component.ts
hudson-newey: Love it

src/app/models/SiteSetting.ts
hudson-newey: *rendered

Sorry, typo

PR #2346, opened 2025-06-27
Add base `/annotation` search/verification page

### LINKED ISSUE #2296: Add "search annotations" button to region list page
Add "search annotations" button to the region list page (so that users can search and verify annotations across the whole A2O)

hudson-newey:
# Add base `/annotation` search/verification page

## Changes

- Adds a base `/annotation` page that can be used to search annotation across the entire website. I have only exposed this route on the region list page so that clients with `projectsHidden` (E.g. A2O) can verify annotations across the entire project.
- Hides the "Projects" input in the annotation search form if `hideProjects` is set

I do not expect users to use this route for sites with > 1 project.

## Issues

Fixes: #2296

## Visual Changes

image

_Annotation search form without the "projects" input because `hideProjects` was set in the environment.json_

## Final Checklist

### DIFF
--- a/src/app/components/annotations/pages/search/search.component.ts
+++ b/src/app/components/annotations/pages/search/search.component.ts
@@ -37,6 +37,7 @@ import { AssociationInjector } from "@models/ImplementsInjector";
 import { IfLoggedInComponent } from "@shared/can/can.component";
 import { AnnotationEventCardComponent } from "@shared/audio-event-card/annotation-event-card.component";
 import { ErrorHandlerComponent } from "@shared/error-handler/error-handler.component";
+import { isInstantiated } from "@helpers/isInstantiated/isInstantiated";
 import { AnnotationSearchFormComponent } from "../../components/annotation-search-form/annotation-search-form.component";
 import { AnnotationSearchParameters } from "../annotationSearchParameters";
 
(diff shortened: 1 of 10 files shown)


src/app/components/annotations/pages/search/search.component.ts
```diff
@@ -200,15 +201,19 @@ class AnnotationSearchComponent
   }
 
   protected verifyAnnotationsRoute(): StrongRoute {
-    if (this.searchParameters.routeSiteId) {
+    // I use isInstantiated() here instead of a falsy assertion so that ids of
+    // zero will pass as a valid project route id.
```
atruskie, line 205: just  note: we should never have a `0` id. It always starts from 1.

I still like the type checking aspect of this however

atruskie (approved):

src/app/components/annotations/pages/search/search.component.ts
hudson-newey: Yeah, I thought about that when I made the change, but it's a minor change, and I _think_ it's more correct.

hudson-newey:
On hold until I can fix `reference` events causing an "Unauthorized" error for non-authenticated users

hudson-newey:
This should probably become the `/library` route

PR #2348, opened 2025-06-27
Use upsert requests for verification page

### LINKED ISSUE #2347: Use createOrUpdate `PUT` endpoint for verifications
At the moment we're making up to 3 requests per verification grid tile.

We recently shipped a new upsert route for verification state that can reduce the number of requests to 1 per grid tile.

hudson-newey:
# Use upsert requests for verification page

## Changes

- Add new `createOrUpdate` method to `baw-api.service`
- Use `createOrUpdate` api method to update verifications

## Issues

Fixes: #2347

## Final Checklist


src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
...
+
+    // As part of the multi part request, if there is only a JSON body, we want
+    // to return the output of the JSON POST request.
+    // If there is only a formData body, we want to return the output of the
+    // formData PUT request if there is both a JSON body and formData, we want
+    // to return the output of the formData PUT request because it is sent after
+    // the initial JSON request, so it will have the most up-to-date model.
+    return iif(
+      () => model.hasJsonOnlyAttributesForCreate(),
+      this.httpPut(path, body, undefined, options),
```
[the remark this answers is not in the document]
hudson-newey: Incorrect. The API uses PUT for upserts

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
...
+      // update route for the formData PUT request using the updatePath
+      // callback. We do this before the concatMap below because the updatePath
+      // callback is dependent on the instantiated class from the POST response
+      // object.
+      map(this.handleSingleResponse(classBuilder)),
+      // Using concatMap here ensures that the httpPost request completes before
+      // the httpPut (formdata) request is made.
+      concatMap((data) =>
+        iif(
+          () => model.hasFormDataOnlyAttributesForCreate(),
```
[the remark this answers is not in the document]
hudson-newey: @atruskie This is actually a really good point that I want your opinion on.

---

I originally chose the creation properties because the way I was thinking about this is:
A models updatable properties are (usually) a subset of the creation properties (with the exception `User.contactable`).

So when contacting this endpoint, you want to provide as much information as possible, because if you only send the `update` properties, the server might not have enough information to create a new model.

---

As a potential fix:

Should I add a third `createOrUpdate` property on our `bawPersistAttr` that makes the property send with upsert requests?

image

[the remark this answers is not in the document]
atruskie: so, the upsert behaviour is a new pattern and not super well defined and only used in place.

I don't want to lock us into a path just yet. 

I think we should send the union of create & update (everything) and when the server doesn't play well we'll adjust this as needed.

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
     );
   }
 
+  /**
+   * Uses the create endpoint to create a model if it doesn't already exist.
```
```suggestion
   * Uses the create endpoint (with PUT instead of POST)
   * to create a model if it doesn't already exist, or to
```

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
...
+   * If the model already exists, the model will be updated.
```
```suggestion
   * update the model if it exists.
```

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
...
+   *
+   * @param classBuilder Model to create
+   * @param path API path
+   * @param model Model to insert into API request
```
atruskie: Missing docs for other params.

Include a note for why we need 2 paths

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
...
+   */
+  public createOrUpdate(
+    classBuilder: ClassBuilder,
+    path: string,
```
```suggestion
    upsertPath: string,
```

src/app/services/baw-api/baw-api.service.ts
```diff
@@ -529,6 +529,83 @@ export class BawApiService<
...
+    updatePath: (model: Model) => string,
+    model: AbstractModel,
+    options: BawServiceOptions = {}
+  ): Observable<Model> {
+    const jsonData = model.getJsonAttributesForCreate();
```
atruskie: shouldn't this be the union of params of update or create?

src/app/services/baw-api/baw-api.service.ts
atruskie: it is right, it doesn't match the semantics from line 552:

```
let formData = model.getFormDataOnlyAttributesForUpdate();
```

atruskie (approved):

src/app/services/baw-api/baw-api.service.ts
hudson-newey: image

"API path to send multipart update requests"

I think this documents why we need two api paths, but let me know if it's still to vague

PR #2349, opened 2025-06-29
Fix project list layout shift

hudson-newey:
# Fix project list layout shift

## Changes

- Fixes images on the project list page not having space allocated before the image is rendered, causing layout shift
- Fixes typeahead input using an internal margin instead of being overrideable by the host

## Final Checklist

### DIFF
--- a/src/app/components/shared/model-cards/card/card.component.scss
+++ b/src/app/components/shared/model-cards/card/card.component.scss
@@ -16,7 +16,7 @@
     border-bottom: var(--card-border);
 
     a {
-      display: inline-block;
+      display: block;
       width: auto;
       height: auto;
     }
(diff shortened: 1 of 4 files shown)


src/app/components/shared/model-cards/card/card.component.scss
```diff
@@ -16,7 +16,7 @@
     border-bottom: var(--card-border);
 
     a {
-      display: inline-block;
+      display: block;
```
atruskie, line 24: so this is the magic that stops the card from collapsing?

maybe add a comment that this is load bearing

atruskie (approved):

src/app/components/shared/model-cards/card/card.component.scss
hudson-newey: image

atruskie: nit: s/disable/display/

PR #2359, opened 2025-07-14
Add sampling options to annotation search form

### LINKED ISSUE #2356: Behavior toggle in annotation search advanced filters
- Only show verification's unverified by me
- Only show verification's that have not been verified

This replaces the "only show unverified checkbox"

hudson-newey:
# Add sampling options to annotation search form

## Changes

- Add new "Sampling Behavior" advanced filter to annotation search form
- Add new `?annotationStatus=` query string parameter
- Replace annotation search "loading" text with loading spinner
- Bumps web components version
- Improves performance of `annotation-event-card.component`
	- Use element references when attaching media controls to spectrogram so there are no document queries
	- Remove the need to generate spectrogram ids from audio_event model ids
	- Use new `viewChild` signals instead of old decorators
- Removes download button from verification grid
- `baw-selectable-items` component now uses item values instead of index's
- Removes progress loss warning messages & modals

## Issues

Fixes: #2356

## Visual Changes

image

_New annotation search form_

---

image

_Verification grid with new styling changes_

## Final Checklist


src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -163,6 +143,21 @@
...
+          class="form-select"
+          placeholder="Only show annotations unverified by me"
+          [value]="searchParameters.select ?? defaultSelect()"
+          (change)="updateDiscreteOptions('select', $event)"
+        >
+          <option value="my-unverified" [disabled]="!isLoggedIn()">Only show annotations unverified by me</option>
+          <option value="only-unverified">Only show annotations that have not been verified</option>
+          <option value="show-all">Show annotations regardless of verification status</option>
+        </select>
+      </label>
```
atruskie: i wonder if we made this a radio option, it might be clearer because you could see all the options at once

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -163,6 +143,21 @@
       [(ngbCollapse)]="hideAdvancedFilters"
       class="card card-body border-top-0"
     >
+      <label class="form-label w-100">
+        Sample Selection Behavior
```
atruskie:
```suggestion
    Show annotations that:
```

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -163,6 +143,21 @@
...
+        <select
+          id="select-input"
+          class="form-select"
+          placeholder="Only show annotations unverified by me"
+          [value]="searchParameters.select ?? defaultSelect()"
+          (change)="updateDiscreteOptions('select', $event)"
+        >
+          <option value="my-unverified" [disabled]="!isLoggedIn()">Only show annotations unverified by me</option>
+          <option value="only-unverified">Only show annotations that have not been verified</option>
+          <option value="show-all">Show annotations regardless of verification status</option>
```
atruskie:
```suggestion
          <option value="my-unverified" [disabled]="!isLoggedIn()">I have not verified</option>
          <option value="only-unverified">no one has verified</option>
          <option value="show-all">even if they have been verified already</option>
```

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -163,6 +143,21 @@
       [(ngbCollapse)]="hideAdvancedFilters"
       class="card card-body border-top-0"
     >
+      <label class="form-label w-100">
```
atruskie: I don't think this should be in the advanced section

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -87,6 +95,10 @@ export class AnnotationSearchFormComponent implements OnInit {
   protected createIdSearchCallback = createIdSearchCallback;
   protected hideAdvancedFilters = true;
   protected scoreRangeBounds = ScoreRangeBounds;
+  protected isLoggedIn = signal(false);
```
atruskie: i wonder if our session service can't expose this signal, so we don't have to copy this local signal in every time to every component

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -269,6 +282,13 @@ export class AnnotationSearchFormComponent implements OnInit {
       return;
     }
 
+    const value = event.target.value;
+
+    this.searchParameters[key] = value;
+    this.emiUpdate();
+  }
+
+  private emiUpdate() {
```
atruskie: spelling, do you mean `emit`?

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -62,13 +63,18 @@ export const sortingOptions = new Map([
...
     direction: "desc",
   }],
-]) satisfies Map<string, Sorting<keyof AudioEvent>>;
+]) satisfies Map<SortingKey, Sorting<keyof AudioEvent>>;
+
+// The select options map can be found in the AnnotationSearchParameter's
+// getters.
+// I have to use a getter because some of the filter conditions depend on the
+// session state.
+export type SelectKey = "my-unverified" | "only-unverified" | "show-all";
```
atruskie: general comment for this file, since we don't like the terminology _select_ much: we can use `VerificationStatus` instead - because that's actually what it is. Not sure if that should leak to UI, i've proposed non-jargony changes for that one

src/app/components/annotations/pages/search/search.component.html
```diff
@@ -56,7 +54,7 @@ <h3>Loading...</h3>
 
     @if (searchResults.length === 0 && !loading) {
       <h2 class="no-results-error">
-        No {{ searchParameters.onlyUnverified ? "unverified" : "" }} annotations
+        No {{ searchParameters.select !== "show-all" ? "unverified" : "" }} annotations
```
atruskie: not sure we want to keep this generally. Should probably be more general like "no matching annotations found". This is  a general search page.

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
```diff
@@ -11,13 +11,30 @@
...
+  // don't have to paint until the user starts to scroll.
+  // This should improve the initial page load and responsiveness.
+  // However, it does have the possibility to create lag when scrolling because
+  // content will be painted on scroll.
+  //
+  // To see what this does, enable "show paints" in the Chrome dev tools
+  // "layers" tab.
+  //
+  // If this proves to be bad for UX, we should revert this change.
+  content-visibility: auto;
```
atruskie, line 25: does this also delay the network request for the audio? how much of spectrogram isn't rendered? is the element made and just not painted?

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
```diff
@@ -4,35 +4,41 @@ import {
...
-  private mediaControls: ElementRef<MediaControlsComponent>;
-
-  public ngOnInit(): void {
-    this.spectrogramId = `spectrogram-${this.annotation().id}`;
-  }
-
-  public ngAfterViewInit(): void {
-    this.mediaControls.nativeElement.for = this.spectrogramId;
+  public ngAfterViewInit() {
+    this.mediaControls().nativeElement.for = this.spectrogram().nativeElement;
```
atruskie: i don't suppose we could do this step in the template:

`<oe-media-controls [for]="this.spectrogram.nativeElement">`

src/app/services/models/annotation.resolver.ts
```diff
@@ -0,0 +1,50 @@
+import { Injectable } from "@angular/core";
+import { ActivatedRouteSnapshot, Resolve } from "@angular/router";
+import { BawSessionService } from "@baw-api/baw-session.service";
+import { SecurityService } from "@baw-api/security/security.service";
+import { AnnotationSearchParameters } from "@components/annotations/pages/annotationSearchParameters";
+import { lastValueFrom } from "rxjs";
+
+@Injectable({ providedIn: "root" })
+class AnnotationsResolver
```
atruskie: naming: this is specifically a _annotation search_ or filter resolver

src/app/test/fakes/data/AnnotationSearchParameters.ts
```diff
@@ -32,13 +36,14 @@ export function generateAnnotationSearchParameters(
       // string[], which incorrectly broadens the type.
       Object.keys(sortingOptions) as SortingKey[],
     ),
+    select: modelData.helpers.arrayElement(
```
atruskie: another place where we could use `verification_status` if we like

atruskie (changes requested):

src/app/components/annotations/pages/search/search.component.html
hudson-newey: I like the idea

What about the copy text "no ~~matching~~ annotations found"

"Matching" sounds confusing in this context (because what are you matching against)

---

Alternative:

"There are no annotations that match your search conditions"

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
hudson-newey: The spectrogram element is made and the fft is computed.

It's just not painted to the web browser

hudson-newey: Here's a small demo showing what the browser is painting when scrolling the page

https://github.com/user-attachments/assets/aead8928-b614-4468-bf2b-24c632c70925

Note that any lag seen is because I have the Chrome dev tools "paint" visualization enabled

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
hudson-newey: I wish we could do this (this was my original hope for allowing element reference's in `for`).

However, element reference's in the template are `undefined` until after the first render (`AfterViewInit`).

And when the element reference finally becomes defined, it doesn't trigger Lit's rendering cycle.

src/app/services/models/annotation.resolver.ts
hudson-newey: image

src/app/components/annotations/pages/annotationSearchParameters.ts
hudson-newey: I'll want to keep to consistent terminology:

E.g. If I was to make this change, the URL parameter should also be "verification-status"

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
hudson-newey: We have a `authTrigger` observable which serves the same purpose.

I'll just use the `authTrigger` in this component

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
atruskie: ok so it looks great, but i'd typically like to see an experiment done with some kind of quantifiable measurement of effect.

e.g. without this setting, page load and idle time was 3.5s, and with it was 3.3s, (10 runs averaged).

I'm not going to insist on this, but it's the only way to deal with perf improvements over a long period of time

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
atruskie: > it doesn't trigger Lit's rendering cycle.

really? that seems like something to unpack

Again, not going to insist on this, but it seems like something we should understand

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
atruskie: you're right it serves the same purpose, but having the signal reduces the dev friction with using it.

it's going to be a fact of life we see duplication like this until we use one rendering method or the other

src/app/components/annotations/pages/annotationSearchParameters.ts
atruskie: yeah see comment below

atruskie: more thinking:

export type VerificationStatusKey = "unverified-for-me" | "unverified" | "any";

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: Offline decided on inline button groups
image

atruskie (approved):

PR #2368, opened 2025-07-18
Fix "skip" decisions not working across users

hudson-newey:
# Fix "skip" decisions not working across users

This PR changes it so that `skip` decisions will **always** be shown.

Please check my logic. Do we want users to re-see their `skip` decisions? (This change will make it so that users will re-see their `skip` decisions if they reload the page).

## Changes

- Fixes `skip` decisions not working across users

## Issues

NA

## Final Checklist


hudson-newey:
The original condition is correct, but is dependent upon some potentially unknown filter behavior

https://github.com/QutEcoacoustics/baw-server/issues/803

PR #2380, opened 2025-08-04
Tag priority

### LINKED ISSUE #2340: Verification grid: tag selection
When showing a subject in the verification grid, we have to choose one tag to verify.

Currently we just show which ever tag is first in the taggings list.

We should instead show, in order of pirority:

0. A priority set by a user (e.g. PE wants to specify that a particular call variant tag should be verified)
1. the tag which matches any tag used to filter the current dataset
    - e.g. if we're searching for "Powerful Owl" then the tag "Powerful Owl" should be prioritized from the tag list "Powerful Owl Alarm Call, Ninox boobook, Powerful Owl"
2. The first common_name tag
3. The first species_name tag
4. and in the worst case: the first remaining tag that does not match any of those conditions

### LINKED ISSUE #2351: Audio event cards can overflow if tag/site names are long
We should set the `max-width: min-content` so that it fits to the spectrograms size.

hudson-newey:
# Tag priority

## Changes

- Adds tag priority algorithm
- Adds "verified tag" input to annotation-search-form.component
- Migrates annotation search form + verification page to signal-based reactivity
- `input.guard` now works with `hasUnsavedChanges` signals
- Adds "score" row to annotation annotation event cards

### Bug Fixes

- Fixes bug where removing all tags from annotation search form would not update search paramters
- Performance improvements to annotation search page due to correctly not using `await` when constructing Annotation model from AudioEvent's
- Fixes long tag & site names expanding annotation event cards
- Fixes [value] assignment into `typeahead[multipleInputs=false]`
- Fixes console error when there were no annotation search results due to attempt to index out of bounds

## Issues

Fixes: #2340
Fixes: #2351

## Visual Changes

image

_Annotation search form with tag filters selected. Notice that the placeholder in the "Tags to Verify" input has all of the tags in the order that they will be selected._

---

image

_Image showing that we only allow one "Tag to Verify" and do not use the pill style typeahead that allows multiple values_

---

image

_Annotation event cards with new "score" row + wrapping tags that would have previously expanded the card container_

## Final Checklist


src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -13,46 +13,57 @@
     inputPlaceholder="All"
     [searchCallback]="tagsApi.typeaheadCallback()"
     [resultTemplate]="tagsTypeaheadTemplate"
-    [value]="searchParameters.tagModels"
+    [value]="searchParameters().tagModels"
     (modelChange)="updateSubModel('tags', $event)"
   ></baw-typeahead-input>
 
+  <baw-typeahead-input
```
atruskie, line 20: we need an explainer for what this does.

> If an annotation has multiple tags, which should we prioritise? This is useful for verifying call type tags

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -180,17 +201,23 @@ export class AnnotationSearchFormComponent implements OnInit {
...
-    this.hideAdvancedFilters = !this.hideAdvancedFilters;
+    this.hideAdvancedFilters.update((current) => !current);
 
-    if (this.hideAdvancedFilters) {
-      this.searchParameters.audioRecordings = null;
+    if (this.hideAdvancedFilters()) {
+      this.searchParameters.update((current) => {
+        current.audioRecordings = null;
+        return current;
+      });
```
atruskie: so "hiding" filters actually clears the filter settings? That's confusing?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
atruskie, line 1: not a fan of the use of `.update` - it smells. I can't think of a better solution though. 

Should each search option itself be a signal?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -87,39 +86,54 @@ class VerificationComponent
...
+        current.routeRegionModel ??= models[regionKey] as Region;
+        return current;
+      });
     }
+
     if (models[siteKey]) {
-      this.searchParameters.routeSiteModel ??= models[siteKey] as Site;
+      this.searchParameters.update((current) => {
+        current.routeSiteModel ??= models[siteKey] as Site;
+        return current;
```
atruskie: ok we could should definitely batch all these changes (assuming we ignore the comment about each prop being turned into a signal)

src/app/components/annotations/pages/annotationSearchParameters.spec.ts
```diff
@@ -253,4 +257,51 @@ describe("annotationSearchParameters", () => {
...
+      // Meaning that relative order is maintained for the filtered tags.
+      const expectedResult = [
+        testedTags[2],
+        testedTags[0],
+        testedTags[1],
+        testedTags[3],
+        testedTags[5],
+        testedTags[7],
+        testedTags[6],
+        testedTags[4],
```
atruskie: this would be easier to read, IMO, if you asserted the order of the tag ids, not the indexes

src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -336,6 +343,50 @@ export class AnnotationSearchParameters
...
+      return 3;
+    }
+
+    if (tag.typeOfTag === "species_name") {
+      return 2;
+    }
+
+    return 1;
+  }
+
```
atruskie: I don't think this belongs here. As stated previously we have need for this function in other places. Sorting tags is not unique to search paramters

Move to annotation?

Also I think if you go with a `makeTagComparer(priorities: Tag[]): (Tag => number)` you won't need to rely on closure magic (which works but the parent object is mutable, right? so this is magic in a whole other way)

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
atruskie, line 1: this means cards will get taller with long/many tags?

src/app/models/data/Annotation.ts
```diff
@@ -37,12 +41,17 @@ export class Annotation extends AbstractModelWithoutId implements IAnnotation {
...
   public readonly audioRecording: AudioRecording;
   public readonly score: number;
   public readonly durationSeconds: number;
   public readonly channel: number;
   public readonly audioEventImportFileId: number;
+  public readonly tagComparer: TagComparer;
+
+  public get tags(): Tag[] {
+    return this.unsortedTags.sort(this.tagComparer);
+  }
```
atruskie: i'm guessing you went with the getter because if the search parameters change and this is read again due to the magic of the comparer closure, this returns a new sort?

Do we ever really need this though?

I'd have sorted the tags that were passed into the creation of this object, or sorted them on assignment with a setter

src/app/services/models/annotation.service.ts
```diff
@@ -14,13 +14,14 @@ export class AnnotationService {
...
-    const audioEventTags = await this.showTags(audioEvent);
+  public async show(audioEvent: AudioEvent, tagComparer: TagComparer): Promise<Annotation> {
     const audioRecording = await this.showAudioRecording(audioEvent);
+    const audioEventTags = await this.showTags(audioEvent);
 
-    const data = {
+    const data: Partial<IAnnotation> = {
       ...audioEvent,
-      tags: audioEventTags,
+      unsortedTags: audioEventTags,
```
atruskie: just sort them here?

atruskie (changes requested):
I'm really not confident on the search options UI.

I think:
- "tags to verify" is ambiguous (I can picture a user saying "I already did this by selecting tags to filter on"
- I think it's an option that's not going to be used very often so it shouldn't be displayed prominently (I hate advanced filters, but maybe it belongs in there for the time being)
- I think it's an option that customisations verification behaviour so it's questionable whether it should be in the search options UI at all (we don't have time to do a separate verification options UI section, but we'll need to do that soon for customising the task)
- Lastly, I think the tags to verify options should not be scoped to currently filtered tags. Phil's example: _filter by audio event import file, and prioritise kookaburra_laugh_. Why do we need filter by it as well? Discussion welcome.

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
hudson-newey: Yeah, instead of expanding the container which is the current behavior

### Before

image

### After

image

atruskie: Yeah looks great, well done

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
hudson-newey: Is it? You can't see it so why would the filters be applied.

I think having a filter that you can't see applied is more confusing

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
hudson-newey: I'll see if I can use some sort of computed `model` signal

src/app/models/data/Annotation.ts
hudson-newey: I think you're right about just sorting them upon assignment.

Changing/tracking in comment below

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: https://github.com/QutEcoacoustics/workbench-client/pull/2380/commits/4fafb39ba33d6cf28db6039f18c8cbb6dff3efcd

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
atruskie: Yeah, but hide/show do not indicate that values are cleared or not cleared.

I think we need a better way to the advanced filters. Let's leave it for another PR though

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
atruskie: talked offline: i think the solution is objects that have signal fields, and maybe a composite signal for the whole object as well.

but we need better best practice advise for this. not a change for this pr

atruskie (approved):
Looks great

PR #2382, opened 2025-08-05
Do not log out user before logging in

### LINKED ISSUE #2381: Do not log out user before logging in
We currently log the user out before logging them in.

This causes the authentication token to rotate, and interfere invalidate the auth token in any running scripts.

We have trialed doing authentication without this log out step in the microsites, and it seems to work without failure.

hudson-newey:
# Do not log out user before logging in

## Changes

- Do not call `signOut` before logging user in
- Do not send authentication information (e.g. `Cookies` or `Authentication` headers) with login requests
- Refactor out `HTTP_AUTHENTICATION_CONTEXT` and replace with native `withCredentials` Request option

## Issues

Fixes: #2381

## Final Checklist

### DIFF
--- a/src/app/services/baw-api/baw-form-api.service.ts
+++ b/src/app/services/baw-api/baw-form-api.service.ts
@@ -59,14 +59,14 @@ export class BawFormApiService<Model extends AbstractModelWithoutId> {
           throw new BawApiError(
             BAD_REQUEST,
             "Unable to retrieve authenticity token for form request.",
-            {}
+            {},
           );
         }
         return token;
       }),
       // Mimic a traditional form-based request
       mergeMap((token: string) =>
-        this.formRequest(submissionEndpoint, body(token), options)
+        this.formRequest(submissionEndpoint, body(token), options),
       ),
       tap((response: string) => {
         // Check for recaptcha error message in page body
@@ -75,14 +75,14 @@ export class BawFormApiService<Model extends AbstractModelWithoutId> {
           throw new BawApiError(
             BAD_REQUEST,
             "Captcha response was not correct.",
-            {}
+            {},
           );
         }
       }),
       // Complete observable
       first(),
       // Handle custom errors
-      catchError(this.handleError)
+      catchError(this.handleError),
     );
   }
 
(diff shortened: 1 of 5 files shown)


hudson-newey:
Discussed offline.

The root problem is that we are sending auth tokens + cookies with login requests.

If we don't send these auth methods when logging in we shouldn't have to dispatch a logout request

src/app/services/baw-api/security/security.service.spec.ts
```diff
@@ -213,18 +213,15 @@ describe("SecurityService", () => {
         );
       }
 
-      it("should call signOut", async () => {
+      // If the user is logged out before logging in, the authentication token
+      // will be rotated.
+      // We used to do this before we started rotating authentication tokens,
+      // so this test protects against regressions.
+      it("should not sign the user out", async () => {
```
atruskie, line 220:
```suggestion
      it("should NOT sign the user out", async () => {
```

This is a personal style thing: whenever I convert a test to a counter test (invert the behaviour), I capitalize the NOT to emphasize the new behaviour. 

Your comment is even better than my little style qurik

src/app/services/baw-api/baw-form-api.service.ts
```diff
@@ -47,25 +47,26 @@ export class BawFormApiService<Model extends AbstractModelWithoutId> {
...
-            {}
+            {},
           );
         }
         return token;
       }),
       // Mimic a traditional form-based request
       mergeMap((token: string) =>
-        this.formRequest(submissionEndpoint, body(token))
+        this.formRequest(submissionEndpoint, body(token), options),
```
atruskie, line 69: like we were talking about offline, this is the request where we want to use withCredentials  - to pass along the new baw session cookie (that's automatically assigned when you do withCredentials false request) and the csrf token cookie?

atruskie (approved):
I assume you saw no need for the http credentials context code any more?

Looks fine if it works. Manual testing needed.

src/app/services/baw-api/baw-form-api.service.ts
hudson-newey: If the above GET request is sent without credentials, neither the `_baw_session` or `XSRF_TOKEN` can be set from the response headers.

https://fetch.spec.whatwg.org/#concept-request-credentials-mode

hudson-newey:
Closed due to required API changes

see: https://github.com/QutEcoacoustics/workbench-client/issues/2381#issuecomment-3194818067

PR #2384, opened 2025-08-07
Verification Compound Tasks

### LINKED ISSUE #2383: "verify and correct" verification task
A stakeholder (DT) would like to provide a correct tag if a "false" decision is made in the verification grid.

With the new web-components [compound tasks](https://github.com/ecoacoustics/web-components/pull/435), this is now possible.

hudson-newey:
# Verification Compound Tasks

## Changes

- Adds `?task=verify-and-correct` url parameter that adds a `tag-prompt` component to the verification grid that is triggered when a `false` decision is made
- Fixes typing for `tagging` endpoint to take an `AudioRecording/AudioRecordingId` instead of `AnalysisJob/AnalysisJobId`
	- API controller: https://github.com/QutEcoacoustics/baw-server/blob/master/app/controllers/taggings_controller.rb#L37
	- `routes.rb`: https://github.com/QutEcoacoustics/baw-server/blob/master/config/routes.rb#L741

## Issues

Fixes: #2383

## Visual Changes

image

_New "verify + correct tag" task behavior in annotation search form_

---

image

_Verification grid with new "Correct Tag" button_

## Final Checklist

### DIFF
--- a/src/app/components/annotations/pages/annotationSearchParameters.ts
+++ b/src/app/components/annotations/pages/annotationSearchParameters.ts
@@ -71,7 +71,7 @@ export const sortingOptions = new Map([
 // session state.
 export type VerificationStatusKey = "unverified-for-me" | "unverified" | "any";
 
-export type TaskBehaviorKey = "verify-and-correct" | "verify";
+export type TaskBehaviorKey = "verify-and-correct-tag" | "verify";
 
 export interface IAnnotationSearchParameters {
   audioRecordings: CollectionIds;
@@ -550,7 +550,7 @@ export class AnnotationSearchParameters
   }
 
   private isTaskBehaviorKey(key: string): key is TaskBehaviorKey {
-    const validOptions: TaskBehaviorKey[] = ["verify-and-correct", "verify"];
+    const validOptions: TaskBehaviorKey[] = ["verify-and-correct-tag", "verify"];
     return validOptions.some((option) => option === key);
   }
 }
--- a/src/app/components/annotations/pages/annotationSearchParameters.ts
+++ b/src/app/components/annotations/pages/annotationSearchParameters.ts
@@ -251,7 +251,8 @@ export class AnnotationSearchParameters
   public set taskBehavior(value: string) {
     if (this.isTaskBehaviorKey(value) || !isInstantiated(value)) {
       // So that we can minimize the number of query string parameters, we use
-      // "unverified-for-me" as the default if there is no "sort" query string parameter.
+      // "unverified-for-me" as the default if there is no "taskBehavior" query
+      // string parameter.
       if (value === "verify") {
         this._taskBehavior = null;
       } else {
--- a/src/app/components/annotations/pages/annotationSearchParameters.ts
+++ b/src/app/components/annotations/pages/annotationSearchParameters.ts
@@ -71,6 +71,8 @@ export const sortingOptions = new Map([
 // session state.
 export type VerificationStatusKey = "unverified-for-me" | "unverified" | "any";
 
+export type TaskBehaviorKey = "verify-and-correct" | "verify";
+
 export interface IAnnotationSearchParameters {
   audioRecordings: CollectionIds;
   tags: CollectionIds;
@@ -105,6 +107,7 @@ export interface IAnnotationSearchParameters {
 
   sort: SortingKey;
   verificationStatus: VerificationStatusKey;
+  taskBehavior: TaskBehaviorKey;
 }
 
 // we exclude project, region, and site from the serialization table because
@@ -128,6 +131,7 @@ const serializationTable: IQueryStringParameterSpec<
 
   sort: jsString,
   verificationStatus: jsString,
+  taskBehavior: jsString,
 };
 
 const deserializationTable: IQueryStringParameterSpec<
@@ -194,6 +198,7 @@ export class AnnotationSearchParameters
 
   private _sort: SortingKey;
   private _verificationStatus: VerificationStatusKey;
+  private _taskBehavior: TaskBehaviorKey;
 
   public get sort(): SortingKey {
     return this._sort;
@@ -233,6 +238,24 @@ export class AnnotationSearchParameters
         this._verificationStatus = null;
       } else {
         this._verificationStatus = value;
+      }
+    } else {
+      console.error(`Invalid select key: "${value}"`);
+    }
+  }
+
+  public get taskBehavior(): TaskBehaviorKey {
+    return this._taskBehavior;
+  }
+
+  public set taskBehavior(value: string) {
+    if (this.isTaskBehaviorKey(value) || !isInstantiated(value)) {
+      // So that we can minimize the number of query string parameters, we use
+      // "unverified-for-me" as the default if there is no "sort" query string parameter.
+      if (value === "verify") {
+        this._taskBehavior = null;
+      } else {
+        this._taskBehavior = value;
       }
     } else {
       console.error(`Invalid select key: "${value}"`);
@@ -524,5 +547,10 @@ export class AnnotationSearchParameters
 
   private isVerificationStatusKey(key: string): key is VerificationStatusKey {
     return this.verificationStatusOptions.has(key as any);
+  }
+
+  private isTaskBehaviorKey(key: string): key is TaskBehaviorKey {
+    const validOptions: TaskBehaviorKey[] = ["verify-and-correct", "verify"];
+    return validOptions.some((option) => option === key);
   }
 }
--- a/src/app/components/annotations/pages/verification/verification.component.html
+++ b/src/app/components/annotations/pages/verification/verification.component.html
@@ -68,6 +68,23 @@
     (focus)="verificationGridFocused = true"
     (blur)="verificationGridFocused = false"
   ></oe-verification>
+
+  @if (hasCorrectionTask()) {
+    <oe-verification
+      verified="skip"
+      shortcut="S"
+      (focus)="verificationGridFocused = true"
+      (blur)="verificationGridFocused = false"
+    ></oe-verification>
+
+    <oe-tag-prompt
+      #tagPrompt
+      verified="unsure"
+      shortcut="C"
+      (focus)="verificationGridFocused = true"
+      (blur)="verificationGridFocused = false"
+    ></oe-tag-prompt>
+  }
 </oe-verification-grid>
 
 <ng-template #searchFiltersModal let-filtersModal>
--- a/src/app/components/annotations/pages/verification/verification.component.html
+++ b/src/app/components/annotations/pages/verification/verification.component.html
@@ -50,6 +50,16 @@
     <baw-grid-tile-content></baw-grid-tile-content>
   </template>
 
+  <oe-tag-prompt
+    #tagPrompt
+    verified="unsure"
+    shortcut="C"
+    (focus)="verificationGridFocused = true"
+    (blur)="verificationGridFocused = false"
+  >
+    Correct Tag
+  </oe-tag-prompt>
+
   <oe-verification
     verified="true"
     shortcut="Y"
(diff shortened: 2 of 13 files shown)


src/app/components/annotations/pages/annotationSearchParameters.ts
```diff
@@ -239,6 +244,24 @@ export class AnnotationSearchParameters
     }
   }
 
+  public get taskBehavior(): TaskBehaviorKey {
+    return this._taskBehavior;
+  }
+
+  public set taskBehavior(value: string) {
```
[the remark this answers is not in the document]
hudson-newey: Why? `TaskBehavior` is a subset of `string`, so this doesn't narrow the type any further.

And the setter is user facing where there is no type checking, so we can't guarantee type safety by narrowing to `TaskBehavior`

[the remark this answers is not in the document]
hudson-newey: (I wish copilot could respond to my query here)

src/app/components/annotations/pages/verification/verification.component.html
```diff
@@ -68,6 +68,23 @@ <h2>
...
+    ></oe-verification>
+
+    <oe-tag-prompt
+      #tagPrompt
+      verified="unsure"
+      shortcut="C"
+      (focus)="verificationGridFocused = true"
+      (blur)="verificationGridFocused = false"
+    ></oe-tag-prompt>
+  }
```
atruskie: ok discussed offline. there are some visual bugs, but:

skip should come last and should not be grouped with anything.  It also probably shouldn't be conditionally defined.

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -146,46 +172,8 @@ class VerificationComponent
...
-      );
-
-      // I use firstValueFrom so that the observable is evaluated
-      // but I don't have to subscribe or unsubscribe.
-      // Additionally, notice that the function is not awaited so that the
-      // render thread can continue to run while the request is being made
-      firstValueFrom(apiRequest);
-    }
+  protected tagTextFormatter(tag: Tag): string {
+    return tag.text;
```
atruskie, line 186: Can tag ever be undefined? Do we need a safe access operator?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+      tagId,
+    };
+
+    const verification = new Verification(verificationData, this.injector);
+
+    // I have to use "as any" here to remove the readonly typing
+    const apiRequest = this.verificationApi.createOrUpdate(
+      verification,
+      subject as AudioEvent,
+      this.session.currentUser,
```
atruskie: smell: shouldn't the service know what the current user is?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+      }
+
+      if (Object.prototype.hasOwnProperty.call(change, "newTag")) {
+        this.handleTagCorrectionDecision(subject);
+      }
+    }
+  }
+
+  private handleVerificationDecision(subjectWrapper: SubjectWrapper): void {
+    const subject = subjectWrapper.subject as Readonly<AudioEvent>;
```
atruskie, line 314: Aren't our subject's `Annotation`s?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
     this.hasUnsavedChanges = false;
   }
 
+  protected handleDecision(decisionEvent: Event): void {
+    if (!this.isDecisionEvent(decisionEvent)) {
+      console.error("Received invalid decision event", decisionEvent);
+      return;
+    }
+
+    this.hasUnsavedChanges = true;
```
atruskie, line 257: i don't know where this is unset, but the service requests are fire and forget... so?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+
+      if (Object.prototype.hasOwnProperty.call(change, "verification")) {
+        this.handleVerificationDecision(subject);
+      }
+
+      if (Object.prototype.hasOwnProperty.call(change, "newTag")) {
+        this.handleTagCorrectionDecision(subject);
+      }
+    }
+  }
```
atruskie, line 311: TODO: we should be updating our annotation model after these updates.

In particular it should now have a new tagging and a set of verifications. This can wait until we start showing verifications in history

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+
+    const decision = decisionEvent.detail;
+    for (const [subject, receipt] of decision) {
+      const change = receipt.change;
+
+      if (Object.prototype.hasOwnProperty.call(change, "verification")) {
+        this.handleVerificationDecision(subject);
+      }
+
+      if (Object.prototype.hasOwnProperty.call(change, "newTag")) {
```
atruskie: maybe leave a comment here along the lines of:

```
// if a new tag is selected then a positive verification is also emitted and saved
// if a new tag is removed, we should also see the old verification get deleted

File: src/app/components/annotations/pages/verification/verification.component.ts [hunk 9]
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+        this.handleTagCorrectionDecision(subject);
+      }
+    }
+  }
+
+  private handleVerificationDecision(subjectWrapper: SubjectWrapper): void {
```
atruskie, line 313: question, how are verifications removed?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+
+    // I use firstValueFrom so that the observable is evaluated
+    // but I don't have to subscribe or unsubscribe.
+    // Additionally, notice that the function is not awaited so that the
+    // render thread can continue to run while the request is being made
+    firstValueFrom(apiRequest);
+  }
+
+  private handleTagCorrectionDecision(subjectWrapper: SubjectWrapper): void {
+    const audioEvent = subjectWrapper.subject as any;
```
atruskie, line 364: this could be narrowed to `Annotation`

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+    const newTagId = subjectWrapper.newTag as any;
+
+    const apiRequest = this.correctTag(audioEvent, newTagId.tag.id);
```
atruskie: how do we handle the removal of `newTag`

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+
+    firstValueFrom(apiRequest);
+  }
+
+  // TODO: This logic should probably be moved to a service
```
atruskie, line 410: yes it should

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -246,18 +234,107 @@ class VerificationComponent
...
+            audioEventId: audioEvent.id,
+            confirmed: ConfirmedStatus.Correct,
+            tagId: newTagId,
+          });
+
+          return this.verificationApi.createOrUpdate(
+            correctVerification,
+            audioEvent,
+            this.session.currentUser,
+          );
```
atruskie: this should not be needed: the verification grid should have emitted the verification already and it should be handles by `handleVerificationDecision`

Is this just double work? or is there a bug somewhere?

src/app/components/annotations/pages/verification/verification.component.ts
atruskie: this is comment is wrong - but we should leave a comment explaining the behaviour

src/app/components/annotations/pages/verification/verification.component.ts
atruskie: this comment is wrong

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: TODO

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: Don't fully understand what you're meaning.

`hasUnsavedChanges` is a property used by our route guard that shows the "are you sure you want to navigate away, you may lose progress" alert.

Therefore, this should be set regardless of if the service request completes or not.

---

I think there's an argument for unsetting this once the request completes/fails (Is this what you're asking for)?

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: image

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: Technically no. And I think we should hard-fail here (although this is up for debate and it's not a hill I'm going to die on).

The `tagTextFormatter` is used to extract text from a tag model in the typeahead input and then cast it to a compatible web component `Tag` model which has a `text` property.

Once selected in the typeahead, the web components internal `Tag` model takes over.

---

I think we should hard fail instead of showing "undefined" or empty strings as a defensive programming practice so that if a user does somehow break the text transformer, the typeahead will throw an error and not show any results which caused an error. I think not showing a result that errored is better than trying to recover from an unexpected error and potentially silently producing bad/incorrect results.

src/app/components/annotations/pages/verification/verification.component.ts
atruskie: Yeah, isn't the result here that we'll always have unsaved changed and the page will always alert?

I think that's another bug we're tracking though...

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: Probably need to change something at the `Annotation` level model here.

`Annotation` implements the `IAudioEvent` interface, but does not extend the `AudioEvent` class.
So it has some poor type checking, and is technically incorrect.

src/app/components/annotations/pages/verification/verification.component.ts
atruskie: add: the verification grid does not emit a separate verification. The new tag model is the decision itself: we break it up into a tagging and a verification

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: The page doesn't currently alert if no decisions have been made. If any decisions have been made, the page will always alert.

This is a bit of a remanence from before users could resume verifications, so I don't think it's as needed these days.

src/app/components/annotations/pages/verification/verification.component.ts
atruskie: I think hard fail is the wrong choice because this will be really frustrating. 

But if we're confident it will never happen, that's fine with me

src/app/components/annotations/pages/verification/verification.component.html
hudson-newey: Skip button should be fixed by upstream

https://github.com/ecoacoustics/web-components/issues/446

Squared edges are being tracked upstream in

https://github.com/ecoacoustics/web-components/issues/447

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: TODO

This should have been done a while back...

hudson-newey: image

atruskie (approved):

hudson-newey:
Update:

I'm hoping with https://github.com/QutEcoacoustics/baw-server/issues/807, the logic of this PR should decrease in complexity.

I'm currently working with the following logic:

- If a false verification decision is made, the tag-prompt decision can enable, and the user can make a tag correction decision. Once a tag correction decision is made:
	- Check if currently have a tag correction applied. If we do:
		- Send a request to delete the AudioEvent/tagging verification
		- Send a request to remove the tagging from the audio event
	- Create a new tagging on the audio event for the corrected tag
	- Upsert new tagging association

- Sometimes a "tag correction" decision can be applied by a "correct" decision being applied to a subject, meaning that the tag correction is invalidated
	- Make initial true verification decision for old tag
	- Remove verification decision attached to tag correction
	- Remove tagging created by tag correction

Some of these requests can be made in parallel, but wanted to document the logic

---

This PR requires some upstream changes in https://github.com/ecoacoustics/web-components/pull/450 to be merged first + a web component version bump

---

Side note: This PR is really poor quality, and we **definitely** need to come back to it next week otherwise we will be left with a humongous tech debt burden

PR #2387, opened 2025-08-09
Fix order of verifications

atruskie:


hudson-newey:
This was a purposeful decision on my end

[`e924a19` (#2384)](https://github.com/QutEcoacoustics/workbench-client/pull/2384/commits/e924a195798fc061e28150edf0dbdaded1bd19ae)

The problem is that:

- Having a `oe-tag-prompt` button in-between the verification and skip button makes the inner verification buttons have squared inner edges
- And conditionally adding the skip verification button as I did before would create a bug where if you created a non-correction verification task, then changed it with the annotation search form, there would be a double skip button

If we want to do something like this, I'd like to fix upstream issues with the button rounding

PR #2402, opened 2025-09-03
Add compound tests

### LINKED ISSUE #2236: Verification grid should not show audio events without tags
At the moment if an audio event does not have a tag, verifying the event in the verification grid will result in a `422 Unprocessable Content` response from the server.

Since it is possible for audio events not to have a tag, this results in some audio events failing to be verified in the verification grid.

### LINKED ISSUE #2394: Add tests for verification grid page API calls
We have an excel spreadsheet of user actions and expected API calls

hudson-newey:
# Add compound tests

## Changes

- Spectrograms & verification grid now load using `SharedArrayBuffer` instead of using mocking
- Add verification transition tests
- Add tests for deleting a users verification for an audio event + tag combination
- Add compound task tests (for both UI buttons and service API calls)
- Add `tagging-correction.service.ts`
- Bump web components by "patch" version (note that this is not the recently released verification resumption version, further work is needed to adjust this version which is out of scope)
	- Fixes contrast for progress bar's undecided color

### Bug Fixes

- Fix error toast when skipping tag correction
- Fix bug where transitioning to a "skip" tag correction from a populated tag correction would not remove the previous tag correction
- Fix "when" predicates throwing a console error once the dataset had been completed
- Fix error when trying to create duplicate tagging during correction event. This can occur if multiple people are correcting the same data source and they both correct to the same tag.
- Only delete tagging corrections if there are no more verifications attached to the tagging. This prevents deleting other peoples tagging correction if two people correct to the same tag

## Issues

Fixes: #2394 (Add tests for verification grid page API calls)
Fixes: #2236 (Verification grid should not show audio events without tags)

## Final Checklist


src/app/services/models/tagging-corrections/tagging-corrections.service.ts
```diff
@@ -0,0 +1,112 @@
...
+        const correctVerification = new Verification({
+          audioEventId: model.audioEvent.id,
+          confirmed: ConfirmedStatus.Correct,
+          tagId: model.correctedTag,
+        });
+
+        const verificationRequest =
+          this.verificationApi.createOrUpdate(correctVerification);
+
+        firstValueFrom(verificationRequest);
```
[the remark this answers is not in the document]
hudson-newey: I purposely didn't await because I didn't want to block the caller

And I didn't make it as part of the observable chain because I still want to return the value from `taggingsApi.create`, I just want to use the return value in another request where I don't care about the response from the verification request.

Edit: Actually maybe I do care about the verification response for error handling...

src/app/components/annotations/pages/verification/verification.component.spec.ts
```diff
@@ -229,32 +294,52 @@ describe("VerificationComponent", () => {
...
+    await detectChanges(spec);
   }
 
   beforeEach(async () => {
-    patchSharedArrayBuffer();
+    // The verification grid will automatically scroll into view once it has
+    // loaded. However, I disable this behavior for testing because it can be
+    // annoying to deal with when trying to read test stack traces and the page
+    // automatically scrolls away from what you were reading.
+    Element.prototype.scrollIntoView = () => {};
```
atruskie: It'd be better if we could disable this on the grid rather than patching the browser

src/app/components/annotations/pages/verification/verification.component.spec.ts
```diff
@@ -338,18 +431,61 @@ describe("VerificationComponent", () => {
...
 
-  async function makeDecision(index: number) {
-    const decisionComponent = decisionComponents()[index];
-    decisionComponent.disabled = false;
+  /**
+   * Prevents the bootstrap modal from opening by setting a local storage key
+   * to automatically dismiss the modal.
+   */
+  function setNoBootstrap(): void {
+    localStorage.setItem("oe-auto-dismiss-bootstrap", "true");
```
atruskie, line 442: this isn't as bad as the other patch, but another place where it would be good if the grid supported it

src/app/components/annotations/pages/verification/verification.component.spec.ts
```diff
@@ -398,232 +534,562 @@ describe("VerificationComponent", () => {
...
+          initialDecision: null,
+          newDecision: DecisionOptions.SKIP,
+          expectedApiCall: verificationSkipApiCall,
+        },
+        {
+          testName: "unsure about a tag",
+          initialDecision: null,
+          newDecision: DecisionOptions.UNSURE,
+          expectedApiCall: verificationUnsureApiCall,
+        },
```
atruskie, line 708: yeah this is nice

src/app/components/annotations/pages/verification/verification.component.spec.ts
```diff
@@ -398,232 +534,562 @@ describe("VerificationComponent", () => {
...
+          // decision is toggled and a DELETE api call is made.
+          expectedApiCall: verificationDeleteApiCall,
+        },
+        // Overwriting to a "skip" decision is not currently supported
+        // TODO: https://github.com/ecoacoustics/web-components/issues/487
+        // {
+        //   initialDecision: DecisionOptions.FALSE,
+        //   newDecision: DecisionOptions.SKIP,
+        //   expectedApiCall: verificationSkipApiCall,
+        // },
```
atruskie: i would leave this test active. Maybe add an expect failure or ignore attribute. 

We want the test to break if the behaviour of the underlying component breaks

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
```diff
@@ -0,0 +1,112 @@
...
+   */
+  public create(model: TaggingCorrection): Observable<Tagging> {
+    const correctTagging = new Tagging({
+      audioEventId: model.audioEvent.id,
+      tagId: model.correctedTag,
+    });
+
+    const isNewTag = model.audioEvent.taggings.every(
+      (tagging) => tagging.tagId !== model.correctedTag,
+    );
```
atruskie: seems like a method that could be useful to have on the model. e.g. `model.audioEvent.hasTag(...)`

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
```diff
@@ -0,0 +1,112 @@
...
+   * immediately verifying it as correct.
+   */
+  public create(model: TaggingCorrection): Observable<Tagging> {
+    const correctTagging = new Tagging({
+      audioEventId: model.audioEvent.id,
+      tagId: model.correctedTag,
+    });
+
+    const isNewTag = model.audioEvent.taggings.every(
+      (tagging) => tagging.tagId !== model.correctedTag,
```
atruskie: Naming: `correctTag` is that an ID or a reference to an object of Tag? Because this comparison looks wrong from the naming

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
```diff
@@ -0,0 +1,112 @@
...
+@Injectable()
+export class TaggingCorrectionsService {
+  private verificationApi = inject(ShallowVerificationService);
+  private taggingApi = inject(TaggingsService);
+
+  /**
+   * Adds a tag correction to an audio event by creating a new tagging and
+   * immediately verifying it as correct.
+   */
+  public create(model: TaggingCorrection): Observable<Tagging> {
```
atruskie: I'm not fond of calling the tag correction a model yet.

I assume you're doing this in anticipation we'll have a tag correction service later on?

src/app/models/data/TaggingCorrection.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface ITaggingCorrection {
+  audioEvent: AudioEvent;
+  correctedTag: Tag["id"];
+}
+
+export class TaggingCorrection extends AbstractData<ITaggingCorrection> implements ITaggingCorrection {
+  public readonly audioEvent: AudioEvent;
+  public readonly correctedTag: Tag["id"];
+}
```
atruskie: Why not just store the correction on the `Annotation` model and pass that through to the service. 

having another model here seems like a lot?

src/app/models/data/TaggingCorrection.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Tag } from "@models/Tag";
+
+export interface ITaggingCorrection {
+  audioEvent: AudioEvent;
+  correctedTag: Tag["id"];
+}
+
+export class TaggingCorrection extends AbstractData<ITaggingCorrection> implements ITaggingCorrection {
+  public readonly audioEvent: AudioEvent;
+  public readonly correctedTag: Tag["id"];
```
atruskie: naming: 
```suggestion
  public readonly correctedTagId: Tag["id"];
```

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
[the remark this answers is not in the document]
atruskie: it's point about race conditions might be valid though

> I purposely didn't await because I didn't want to block the caller

Leave that up to the caller, yeah?

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        // If there are no more verifications for this tag on this audio event
+        // then we can delete the tagging.
+        //
+        // TODO: Once we have support for database backed tag corrections, we
+        // should only delete the tagging if it was created as part of this
+        // correction.
+        // see: https://github.com/QutEcoacoustics/baw-server/issues/807
+        return this.taggingApi.destroy(
```
atruskie: this seems dangerous.

Will this ever fire for taggings that are not corrections?

atruskie (approved):

src/app/components/annotations/pages/verification/verification.component.spec.ts
hudson-newey: This functionality is on the ecosounds verification grid side.

https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/annotations/pages/verification/verification.component.ts#L169-L186

Are you thinking that auto-scroll should be merged upstream into the web components with an `autoscroll` attribute (similar to `autofocus`). (Note I'm using a non-hyphenated `autoscroll` because it's more compliant with existing html attributes like `autofocus`, `autocapitalize`, `autocorrect`, etc...)

src/app/components/annotations/pages/verification/verification.component.spec.ts
hudson-newey: Do you have any suggestions how else this could be implemented?

I'm open to wild/idealistic future ideas that we can merge into the upstream web components

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: The `TaggingCorrection` model is a data model. Similar to the `Annotation` model, it doesn't require any API backing, and is purely for communicating between the `tagging-corrections.service` and its consumers.

In my head, it's like a view model that is bound to a non-api backed service

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: Yes, if the user is the only one performing a verification task and corrects to an existing tagging. If they then delete this tag correction, one of the original tagging that is not a correction will be deleted.

---

Until we have api-backed corrections so that we know if a tagging is a correction, I'm open to never deleting a tagging and just verifying the now defunct tagging as "false" instead of deleting it.

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: You're right in a later comment. The correct name should be `correctedTagId`

hudson-newey:
I'll be finishing off my microsite event cards PR then finishing off the review feedback for this PR.

Thank you for the review, I really appreciate it

src/app/components/annotations/pages/verification/verification.component.spec.ts
atruskie: I was mistaken in where I thought this was implemented.

But the sentiment remains the same. Some kind of angular property on the component that disables scrolling for test scenarios. Or maybe just a inTest service that allows us to conditionally define behaviours.

Either way, it's not the end of the world

src/app/components/annotations/pages/verification/verification.component.spec.ts
atruskie: `bootstrap="false"` on the web component?

hudson-newey: Sounds good. I'll open an issue on the web components

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
atruskie: >  I'm open to never deleting a tagging and just verifying the now defunct tagging as "false" instead of deleting it.


Hmm, that seems reasonable. 

Not sure what the best call is.

I think deleting leads to less confusing data.

Can we delete only if we know we created this tagging in the current session?

It's not a perfect metric (i.e. it will fail on multi session usage ) but it's probably correct 90% of the time?

Or maybe it's best to just leave it as is... every patch we add on adds complexity

src/app/components/annotations/pages/verification/verification.component.spec.ts
hudson-newey: https://github.com/ecoacoustics/web-components/issues/504

src/app/components/annotations/pages/verification/verification.component.spec.ts
hudson-newey: How about a `scrollService` so that we can mock the provider during testing?

That way during testing our `scrollService` can have no behavior, but during dev/production builds, it does scroll the viewport?

atruskie: sounds perfect

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: > Can we delete only if we know we created this tagging in the current session

I do currently track this information, so I think only deleting the tag if we're 100% sure it was created as part of a tag correction is the correct move.

This might cause some tags to be added in some rare cases, but I think having some additional tags that are verified as "false" is better than deleting tags

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: Just realized that `tagIds` is a set, so I don't need this `every`, I can just do `audioEvent.tagIds.has(correctedTagId)`

image

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: Correction: This is actually how it currently works. We only issue `tagCorrections.destroy` for taggings that were added as part of the `sessionTagCorrections`

image

src/app/models/data/TaggingCorrection.ts
hudson-newey: `Annotation` extends the `AudioEvent` model, so `Annotation`'s can have multiple tags.

There's no way to know what tagging a correction is referring to without a similar interface on the `Annotation` model.

I think this data model is a bit confusing because it doesn't have an `incorrectTagId`

image

hudson-newey: I don't mind the sound of this, but not sure how it'd work, can you elaborate a bit further? How would we add correction information to the Annotation model?

And how would we decide what tag corrections were added/removed?

atruskie: The annotation model is our source truth for an annotation. It's a metamodel. It's a bit under specced at the moment - it should be able to be loaded with all taggings, and verifications, ... and corrections.

It could simply store a Map of corrections, e,g. `Map<Tag.id, Tag.id>`.

Or we could augment the Tagging model to behave how we expect it to when the server implements corrections.

src/app/services/models/tagging-corrections/tagging-corrections.service.ts
hudson-newey: Moved to a `corrections` annotation property

PR #2443, opened 2025-10-01
Region list page map tab

### LINKED ISSUE #2436: Add site visualization to project/region list page
We should add tabs to the project/region list pages that allow you to search for a site by geographic location.

Note: Until we have a new optimized "fast" service to get site models, we should make this a configurable option, so that large website instances (e.g. Ecosounds) can disable this functionality.

hudson-newey:
# Project & Region list page map tab

## Changes

- Adds "tiles" and "map" tab to the project & region list pages
- Adds `filter` property to the `baw-site-map` component
- Refactors project & region list page logic into common template
- Adds clustering to Google maps
- Add `isFetching` state to site map component so that the loading spinner can show during data fetching
- Add `loading` class to `baw-site-map` component host that can be used to conditionally apply styles depending on if the site map is loaded or not
	- I use this on the project/region list pages to add a shadow once the site map has loaded. I don't add the shadow while the map is loading because it's too notable
- No longer fetch sites in a site-map component. I did this for two reasons
	- Some website-level site filter requests take a long time, and I don't want to hang SSR for just these requests
	- Site maps are heavily reliant on user authentication state, and return an empty array a lot of the time (because the unauthenticated SSR request wouldn't have access to any of the sites that would be shown). Meaning the map would flash a "No sites" message.
	- I recognize that we might be able to enable this once SSR does authenticated requests in an associated TODO comment
- site-map.component will now enter a `isFetching` state after 400ms of filter conditions changing. This means that we don't flash a loading spinner for fast requests. This makes map updates feel more instant & produces less jank
- Reduces `defaultDebounceTime` to 200ms (from `500ms`). Once we merge this PR, let me know if this results in a lot of server load, and I can increase/revert this change.
- Adds 5px to the annotation import preview table's "Start Time" and "End Time" columns. (I did this because when I imported Phil's annotations, I saw that long start/end times overflowed the allocated cells
- Adds `groupBy` attribute to `site-map.component`

### Development concerns

- Adds `readonly` annotation to signals that were added before I knew about this [recommendation](https://angular.dev/style-guide#use-readonly-for-properties-that-shouldnt-change)
- Fixes `templateUrl` and `styleUrl`s to conform to the Angular style guide [recommendation](https://v19.angular.dev/style-guide#style-05-04)
- Migrate from deprecated `MapMarker` to `AdvancedMapMarker` ([more info](https://mapsplatform.google.com/resources/blog/announcing-highly-customizable-higher-performance-markers/))

## Issues

Fixes: #2436

## Visual Changes

image

_Region "tiles" tab (showing A2O)_

Note: Images arn't loading because of CORP. Images do correctly load when on the correct domain

---

image

_Region "map" tab (showing A2O)_

---

image

_Project list page with `homePageAndProjectListMaps: false`_

This is how Ecosounds will look until we add the site clustering endpoint

---

image

_New marker colors depending on what region the marker belongs to_

## Final Checklist


src/app/components/shared/map/map.component.scss
```diff
@@ -4,7 +4,7 @@
   justify-content: center;
   align-items: center;
 
-  border: 1px solid black;
+  box-shadow: inset 0 0 0.2rem rgba(0, 0, 0, 0.2);
```
hudson-newey, line 7: I did this so that rounding could be applied to the map component.

This obviously isn't the correct solution. I'll find a correct solution before merging this PR.

src/app/services/maps/maps.service.ts
```diff
@@ -118,6 +126,11 @@ export class MapsService {
     return googleMapsUrl;
   }
 
+  private googleMapClusterUrl(): string {
+    // https://github.com/angular/components/blob/974d42f04/src/google-maps/map-marker-clusterer/README.md?plain=1#L5
+    return "https://unpkg.com/@googlemaps/markerclusterer/dist/index.min.js";
```
hudson-newey: Before merging, I'll add this as a dependency & static asset that will be served through our Angular server.

This would have two benefits.

1. The version would be tracked in our package.json
2. Hopefully we can re-use the same `Keep-Alive` connection so that we don't have to do another TLS + TCP handshake

src/app/services/maps/maps.service.ts
```diff
@@ -11,7 +11,7 @@ export enum GoogleMapsState {
   Failed,
 }
 
-export type MapMarkerOptions = Partial<google.maps.marker.AdvancedMarkerElement>;
+export type MapMarkerOptions = google.maps.marker.AdvancedMarkerElement;
```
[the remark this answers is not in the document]
hudson-newey: The Partial model was incorrect.

I now correctly initialize `MapMarkerOptions` in the `Site` model using the `AdvancedMarkerElement` constructor

src/app/components/projects/components/site-map/site-map.component.ts
```diff
@@ -72,19 +87,35 @@ export class SiteMapComponent extends withUnsubscribe() implements OnChanges {
...
-          this.markers.set(List());
-          throw new Error("Failed to load sites for map", { cause: err });
-        },
-      });
+      .pipe(first(), takeUntil(this.unsubscribe));
+
+    // For fast requests, we don't want to flash a loading state.
+    // However, if fetching takes a long time, we do want to show a loading
+    // state to inform the user that something is happening.
+    const timerSub = timer(400)
```
atruskie: name the magic number. Maybe also put it next to our default debounce const?

src/app/components/projects/components/site-map/site-map.component.ts
```diff
@@ -28,23 +30,29 @@ import { first, takeUntil } from "rxjs/operators";
...
 })
 export class SiteMapComponent extends withUnsubscribe() implements OnChanges {
   private readonly sitesApi = inject(ShallowSitesService);
+  private readonly isServer = inject(IS_SERVER_PLATFORM);
 
   public readonly projects = input<IdOr<Project>[]>();
   public readonly regions = input<IdOr<Region>[]>();
   public readonly sites = input<IdOr<Site>[]>();
   public readonly selected = input<List<IdOr<Site>>>();
+  public readonly filters = input<InnerFilter<Site>>();
```
atruskie, line 57: I'm not overly keen on passing filters through - see our previous discussion on just providing an observable

atruskie: also how do we handle filtering by regoin name?

on the region list page the filter would be `name eq "abc"`, but to pass it into the map, it would need to be be `regions.name eq "abc"`. I'll probably work out where this translation happens elsewhere....

atruskie: Maybe this is just a lack of documentation for the input? I'm making assumptions about it's behaviour...

src/app/components/shared/model-list/model-list.component.ts
```diff
@@ -0,0 +1,117 @@
...
+    });
+  }
+
+  private updateMapFilters(): void {
+    const textFilter = this.filter;
+    if (textFilter) {
+      // I need to type cast the filter here because our filter types don't
+      // currently reflect that we can filter by associated models.
+      this.mapFilter.set({
+        [`${this.modelKey()}.name`]: { contains: textFilter },
```
atruskie: this is a bit icky - it will quickly become unmaintainable if we add any more filter criteria.

I wonder if this is better setup as a two-stage filter. Filter for model with search criteria, and then use found ids to return sites (or just feed model ids into baw-site-map directly).

It will also allow us reuse a model filter request (e.g. filter regions) directly in the map without rerunning a request (e.g. filter sites where regions.name = region name) as soon as we turn on the map

src/app/helpers/app-initializer/app-initializer.ts
```diff
@@ -49,6 +49,7 @@ export interface Settings {
     harvestFilenameGuide: string;
   };
   hideProjects: boolean;
+  websiteMaps: boolean;
```
atruskie: nit naming: suggest `globalMaps`

src/app/services/config/configMock.service.ts
```diff
@@ -60,6 +60,7 @@ export const testApiConfig = new Configuration({
       harvestFilenameGuide: "http://broken_link",
     },
     hideProjects: false,
+    websiteMaps: true,
```
atruskie: if name change, then here too

src/app/app.helper.ts
```diff
@@ -3,8 +3,12 @@ import { inject, InjectionToken, PLATFORM_ID } from "@angular/core";
 
 /**
  * Default number of milliseconds to wait when de-bouncing an input
+ *
+ * I chose a 200ms debounce delay because it means that worst case scenario,
+ * we are sending 5 requests per second to the server, which _should_ not
+ * overload the server.
```
atruskie, line 9: worth mentioning this was originally chosen to reduce jank for a user based on typing speed so it's a balance to tune

atruskie (changes requested):
Great work!

The clustering is going to be necessary, but I'm not fond of the default styles. That dark blue colour has low contrast against the ocean, and it's a real jarring shift up to the bright red colour for larger clusters.


I'm kinda realising that always showing sites in a map for regions/projgects pages is going to be confusing. We didn't get round to coloring markers by group (region/project), did we?

src/app/components/projects/components/site-map/site-map.component.ts
hudson-newey: > also how do we handle filtering by regoin name?

You're correct, you just pass a `{ "regions.name": { "contains": "abc" } }` object into this input

hudson-newey: > see our previous discussion on just providing an observable

I'm quite opposed to this, why would we want an observable here?

We don't need the functionality of observable's (history/queue of items), and the pagination template doesn't omit anything like an `Observable<Filter>`

src/app/helpers/app-initializer/app-initializer.ts
hudson-newey: Nit: "global" sounds geographical

if we change this name to `globalMaps`, it might be confusing because you can still have global geographical map with this environment setting turned off.

By a similar thought train, I also originally rejected `siteMaps` because the term "site" was overloaded with "region/site/point" terminology (which is why I originally landed on "websiteMaps").

I'm happy to keep `globalMaps`, just want to make a note that it might be confusing

src/app/components/projects/components/site-map/site-map.component.ts
atruskie: I think the best idea I had for rearchitecting was in https://github.com/QutEcoacoustics/workbench-client/pull/2443#discussion_r2396408344 and thus my actual suggestion is to remove the filter completely

src/app/components/projects/components/site-map/site-map.component.ts
hudson-newey: I like the idea

src/app/helpers/app-initializer/app-initializer.ts
atruskie: Fair. Happy with a different name. Even long or obtuse - hopefully this is just a temporary config.

`disableHomePageAndProjectListMaps`

hudson-newey: `disableHomePageAndProjectListMaps` This works for me

I'll probably remove the negation and make it something like `homePageAndProjectListMaps`

src/app/components/shared/model-list/model-list.component.ts
hudson-newey: I like the idea of re-using region ids in the site map component, but I think we can only do this when there is a single page of results.

The "tile" view only returns 25 items at a time, meaning that if we pass the region ids returned by the pagination filter request, we'll only show the sites from the top 25 regions.

A possible fix would be to query all projects/regions on the list page. Potentially by still doing paging requests so that we can render the tile view after the first response. But this seems like a lot of overhead / really inefficient to (essentially) keep the entire projects/regions table in the clients memory.

hudson-newey: Discussed offline.

The main problem with this solution is that we'll have to maintain two filters, one for the project/region list page, and one for the site filters.

We came up with a middle ground solution where we add a `associationModelFilter` helper that can merge an association filter (project/region filter) into an existing base filter (sites).

That way we can re-use the same project/region filters and just call this helper to create a site filter.

PR #2462, opened 2025-10-09
Audio event map

### LINKED ISSUE #2403: Create geographic audio event map
> As a citizen scientist, I want to see a map of species locations and confidence estimates, so that I can see where the species have been detected and the reliability of the data.

Similar to how the A2O search shows a map where similar audio recordings exist, users want to see a list of events on a map that also includes confidence information.

Image

We have considered making these visualizations web components that so that they can be embedded in microsites as part of: https://github.com/ecoacoustics/microsite-template/issues/69

### LINKED ISSUE #2466: Map info box content should have inline "close" button
At the moment, the close button is on its own line

Image

However, when possible, the close button should be on the same line as the site name

### LINKED ISSUE #2468: map marker clusters should use theme color

### LINKED ISSUE #2477: Seperate verification parameters from annotation search parameters
At the moment the annotation search page and the verification page share the same url parameters.

This means that the search page has some unusual functionality like by default only showing events that you have not verified and having search filters for the "task tag" and a toggle for the "verification + correct" behavior.

The fact that annotation searching and verification is so tightly coupled has caused issues for the new event map where we want to show all annotations regardless of verification status.

---

We should bifurcate these search parameters so that these search parameters can be used individually.

We probably also need to change the annotation search form to not include verification information.
We should probably only show the verification behavior options once the "verify" button is clicked.

### LINKED ISSUE #2490: Associations should support backing field mutation
The association cache e.g. for `@hasMany(TAG, "tagIds") tags` are is invalidated if the backing field e.g. `tagIds` is updated.

This means that if you query `model.tags`, update `model.tagIds = []`, then query `model.tags` again, it will return the same value as if it was never updated.

This is one of the reasons why our models are readonly

hudson-newey:
# Add audio event map

## Changes

- Adds audio event map page + component
- Adds `showVerificationFilters` input to the annotation search form (and a TODO to bifurcate these components)
- `map.component` can now take an optional ng-template's for `markerTemplate` and `markerInfoTemplate` (for hover)
- Event map clustering now uses the `--baw-primary` color for clusters
- Bifurcate search parameters & verification parameters
- Adds support for mutating `AbstractModel` association identifiers

### Bug Fixes

- Fixes a lot of internally pointing links being `href` instead of `bawUrl` (using `href` does not use Angular router navigation)
- Fixes audio event cards being rendered while fetching search page
- Fixes layout shift in `audio-event-card.component` when audio file changes from an unfetched -> fetched state
- Fixes verification grid height not subtracting toolbar height on mobile devices
- Fixes verification grid overflowing x-axis on larger grid sizes

### Dev concerns

- `hasOne`, `hasMany`, `updater`, `creator`, and `deleter` decorators now infer types instead of requiring explicit type parameters
- Improved typing for `Id` type
- The `map.component` now creates map markers. This means that consumers can use the map component without having to wait for the `google` namespace to be initialized.

## Issues

Fixes: #2403 (Create geographic audio event map)
Fixes: #2466 (Map info box content should have inline "close" button)
Fixes: #2468 (map marker clusters should use theme color)
Fixes: #2477 (Seperate verification parameters from annotation search parameters)
Fixes: #2490 (Associations should support backing field mutation)

## Visual Changes

[Screencast_20251020_143241.webm](https://github.com/user-attachments/assets/c95d645e-d7e6-4843-a88d-5a7833b164c6)

_An event map with mock data_

---

image

_Mobile view while panning the event map_

---

image

_Mobile view after focusing an event_

---

image

_A map using the `--baw-primary` color for clustering

I chose not to use the `--baw-highlight` color because a lot of instances (A2O & dev) use dark green which does not contrast very well on maps

image

If you wish to override this decision, I am more than happy because it was a close decision in my mind.

## Final Checklist


src/app/components/events/pages/events/eventMapSearchParameters.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { EventMapSearchParameters } from "./eventMapSearchParameters";
+
+describe("EventMapSearchParameters", () => {
```
hudson-newey: I'll finish off these tests before merging

src/app/components/events/pages/events/events.resolver.spec.ts
hudson-newey, line 1: Will write tests before merging

src/app/components/shared/hidden-copy/hidden-copy.component.ts
hudson-newey, line 1: I originally refactored this component because I was going to include a "copy" button in the annotation search form, but decided against it to when I realized I would be bifurcating the component in the future.

There are no functional changes in these refactors, just uplifting it to the modern Angular style guide.

I didn't revert my refactors because I assume that they were useful, even if it didn't directly contribute to this PR

src/app/services/baw-api/grouped-audio-events/grouped-audio-events.service.spec.ts
hudson-newey, line 1: Will re-enable these tests before merging

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
```diff
@@ -37,8 +49,22 @@ oe-axes {
...
+  flex-direction: column;
+
+  flex: 1 1 auto;
+
+  min-height: 0;
+}
+
 .card-container {
-  max-width: min-content;
+  display: flex block;
```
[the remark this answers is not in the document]
hudson-newey: This is a new CSS feature

`flex` is for child elements, `block` is for the current element

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -85,20 +93,36 @@ export class AnnotationSearchFormComponent implements OnInit {
...
   }
 
-  public readonly searchParameters = model.required<AnnotationSearchParameters>();
+  // Having a boolean input to show/hide verification filters is a smell that
+  // the component architecture could be improved.
+  // However, this is a minor issue and can be addressed later if needed.
+  // TODO: We should refactor this component into two separate components.
+  // see: https://github.com/QutEcoacoustics/workbench-client/issues/2477
+  public readonly showVerificationFilters = input(true);
+  public readonly showSortingFilters = input(true);
```
atruskie:
```suggestion
  public readonly showSorting = input(true);
```

or 

```suggestion
  public readonly allowSorting = input(true);
```

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -204,6 +182,24 @@ export class AnnotationSearchParameters
...
+
+    // This is a hack so that we can use the annotation search parameters on
+    // the event map page without having to worry about verification parameters
+    // being included in the filter request.
+    //
+    // TODO: Remove this once verification parameters are seperated from the
+    // search parameters.
+    // see: https://github.com/QutEcoacoustics/workbench-client/issues/2477
+    public includeVerificationParams = true,
+    public includeSortingParams = true,
```
atruskie: is this really needed?

if a user can't choose the hidden options in the UI, why do we need to change QSP parser?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
atruskie: Also why hide the filters?

The sort has no meaning for the cluster API but we can reuse the sort preference for the show events list part.

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -363,7 +356,26 @@ export class AnnotationSearchParameters
...
+    }
+
+    // On the event map page, we don't want to use a sorting parameter.
+    // I have used an early return here instead of returning undefined or null
+    // from the sortingFilters() method so that both the
+    // includeVerificationParams and includeSortingParams conditions located
+    // next to each other for clarity.
+    if (!this.includeSortingParams) {
+      return { filter };
+    }
```
atruskie: i don't think we need this in this class, see above comment

atruskie (commented):
Did we change the default verification status filter for the search page to _are verified or unverified_? I don't see a customisation point for that.

hudson-newey:
> Did we change the default verification status filter for the search page to are verified or unverified? I don't see a customisation point for that.

We did not change this. But I can (and will)

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
hudson-newey: The default verification filtering is `unverified-for-me` which would emit this really complex filter body

image

If there are no verification param UI, the user could not change off of this default.
Additionally, manually setting the `verificationStatus` param to `any` would change the qsps.

Adding an `includeVerificationParams` makes the most sense to me

hudson-newey: Although if we are going to move forwards with making the default `any`, this will probably change to `defaultVerificationStatus` so that the verification grid can have a default of `unverified-by-me` and the search page can have `any`

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
hudson-newey: Do you mean "conditions"?

I don't see any reason to remove this class

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -151,41 +151,21 @@
...
-  <div class="mt-2">
-    <label class="form-label w-100">Task Behavior</label>
-    <baw-selectable-items
-      [options]="taskBehaviorOptions"
-      [selection]="searchParameters().taskBehavior ?? 'verify'"
-      [inline]="true"
-      (selectionChange)="updateParameterProperty('taskBehavior', $event)"
-    ></baw-selectable-items>
-  </div>
+  @if (showVerificationOptions()) {
```
atruskie, line 158: my understanding is that we shouldn't need this conditional anymore - why is it included?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
```diff
@@ -85,30 +90,35 @@ export class AnnotationSearchFormComponent implements OnInit {
     });
   }
 
-  public readonly searchParameters = model.required<AnnotationSearchParameters>();
+  public readonly searchParameters =
+    model.required<AnnotationSearchParameters>();
   public readonly searchParametersChange = output<AnnotationSearchParameters>();
 
+  public readonly showVerificationOptions = input<boolean>(true);
```
atruskie, line 97: following the chain of that conditional... why?

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -528,21 +422,33 @@ export class AnnotationSearchParameters
...
 
-    const filters = this.verificationStatusOptions.get(statusKey);
+    const isValidKey = this.isVerificationStatusKey(this.verificationStatus);
+    const statusKey = isValidKey ? this.verificationStatus : defaultVerificationStatus;
+    if (!isValidKey) {
+      console.warn(
+        `Invalid verification status key provided: '${this.verificationStatus}'. Falling back to '${statusKey}'.`,
+      );
+    }
+
```
atruskie: nit readability: i'd put the fallback assignment into the condition after the warning

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -528,21 +422,33 @@ export class AnnotationSearchParameters
...
     const defaultSortKey = "created-asc" satisfies SortingKey;
-    const sortingKey = this.isSortingKey(this.sort)
-      ? this.sort
-      : defaultSortKey;
+
+    const isValidKey = this.isSortingKey(this.sort);
+    const sortingKey = isValidKey ? this.sort : defaultSortKey;
+    if (!isValidKey) {
+      console.warn(`Invalid sorting key provided: '${this.sort}'. Falling back to '${sortingKey}'.`);
+    }
```
atruskie: ditto

src/app/components/annotations/components/modals/search-filters/search-filters.component.html
atruskie, line 1: ~i wonder if we should suffix the file name with `-modal`.~ Never mind I finally noticed it in the path.

I assume it is placed in this directory under the assumption it could be used in multiple places. That makes sense, but it threw me to begin with because it lacked context.

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
atruskie: yeah i can only find one place where this is used and it is set to true, and the default is true... so conditional pointless?

src/app/components/annotations/components/verification-form/verification-parameters.resolver.ts
```diff
@@ -0,0 +1,46 @@
...
+class VerificationParametersResolver
+  implements Resolve<VerificationResolverData>
+{
+  private readonly session = inject(BawSessionService);
+  private readonly security = inject(SecurityService);
+
+  public async resolve(
+    route: ActivatedRouteSnapshot,
+  ): Promise<VerificationResolverData> {
+    if (!this.security.firstAuthAwait.value) {
```
atruskie, line 16: i know this is here for a reason but i can't remember why.... why do we need auth to deserialize query string parameters?

atruskie: i followed the chain.. it's only needed for the toFilter stage,

ditto with annotation search parameters. Unless there's a reason we need this in the resolver, we could move the user requirement into the `toFilter` method of both.

src/app/components/annotations/pages/annotation-map/annotation-map-parameters.resolver.ts
```diff
@@ -0,0 +1,27 @@
...
+export const annotationMapParameterResolvers = {
+  showOptional: "annotationMapParametersResolver",
+  providers: [
+    {
+      provide: "annotationMapParametersResolver",
+      useClass: AnnotationMapParametersResolver,
+      deps: [],
+    },
+  ],
+};
```
atruskie, line 27: future note this pattern seems really common, mayber we can make a resolver function that works for all parameters

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
```diff
@@ -25,6 +17,26 @@ oe-axes {
...
+}
+
+.axes, oe-indicator, oe-spectrogram {
+  display: block;
+
+  position: relative;
+  width: 100%;
+  height: 100%;
+}
+
```
atruskie, line 39: these seem like hacks - should these be fixes applied to the web-components?

src/app/components/shared/event-map/event-map.component.html
```diff
@@ -0,0 +1,36 @@
+<baw-map
+  [markers]="markers()"
+  [markerOptions]="{ gmpClickable: true }"
+  (markerClicked)="handleMarkerClicked($event)"
+>
+  <ng-template #markerTemplate let-marker="marker">
```
atruskie, line 6: this is cool

src/app/components/shared/event-map/event-map.component.html
```diff
@@ -0,0 +1,36 @@
...
+      <circle cx="20" cy="20" r="16" fill="var(--baw-highlight)" opacity="0.8" />
+      <text
+        class="marker-count"
+        x="20"
+        y="25"
+        text-anchor="middle"
+        font-size="12"
+        fill="var(--baw-highlight-contrast)"
+        letter-spacing="0.75px"
+        font-weight="bold"
```
atruskie, line 30: though i'd suggest the style based attributes should be omitted and replace with css styles

src/app/components/shared/map/clusterRenderer.ts
```diff
@@ -0,0 +1,59 @@
...
+        <circle cx="120" cy="120" opacity=".3" r="90" />
+        <circle cx="120" cy="120" opacity=".2" r="110" />
+        <text
+          x="50%"
+          y="50%"
+          fill="var(--baw-primary-contrast)"
+          text-anchor="middle"
+          font-size="50"
+          dominant-baseline="middle"
+          font-family="roboto,arial,sans-serif"
```
atruskie, line 46: pull out to CSS?

src/app/components/shared/map/clusterRenderer.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const title = `Cluster of ${clusterWeight} markers`;
+    const zIndex: number = Number(google.maps.Marker.MAX_ZINDEX) + clusterWeight;
+
+    const parser = new DOMParser();
+    const markerContent = parser.parseFromString(
+      svg,
+      "image/svg+xml"
+    ).documentElement;
+    markerContent.setAttribute("transform", "translate(0 25)");
```
atruskie, line 59: ditto css - i assume it's mildly difficult or else you wouldn't have bothered with the javascript. What is the blocker?

src/app/components/shared/map/map.component.ts
```diff
@@ -44,27 +63,48 @@ type MarkerGroup = unknown;
...
+    input<google.maps.marker.AdvancedMarkerElementOptions>();
 
   // Setting to "hybrid" can increase load times and looks like the map is bugged
   public readonly mapOptions = input<MapOptions>({ mapTypeId: "satellite" });
+  public readonly fetchingData = input(false);
 
   public readonly newLocation = output<google.maps.MapMouseEvent>();
 
+  /**
+   * An event that can is emitted when the marker is clicked with the mouse or
```
atruskie:
```suggestion
   * An event that is emitted when the marker is clicked with the mouse or
```

src/app/components/shared/map/weightedSuperClusterAlgorithm.ts
```diff
@@ -0,0 +1,42 @@
...
+
+class WeightedCluster extends Cluster {
+  public override get count(): number {
+    return this.markers.reduce((total, marker) => {
+      return total + getMarkerWeight(marker);
+    }, 0);
+  }
+}
+
+export class WeightedSuperClusterAlgorithm extends SuperClusterAlgorithm {
```
atruskie, line 23: could do with a couple of doc strings.

// gets a marker's custom clusterWeight or defaults to one

// creates a cluster that sums all the weights of the markers within it

// ....
(i don't know what the `WeightedSuperClusterAlgorithm` class does)

src/app/components/shared/map/weightedSuperClusterAlgorithm.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type MarkerWeight = number;
+
+export type WeightedClusterMarker = Marker & {
+  clusterWeight?: MarkerWeight;
+};
+
+function getMarkerWeight(marker: Marker): MarkerWeight {
+  const weightedMarker = marker as WeightedClusterMarker;
+  return weightedMarker.clusterWeight ?? 1;
```
atruskie, line 12: didn't I see this already? Yeah in ClusterRenderer

src/app/helpers/parameters/merge.ts
```diff
@@ -0,0 +1,8 @@
+import { Params } from "@angular/router";
+
+export function mergeParameters<
+  const Base extends Params,
+  const Override extends Params,
+>(base: Base, ...overrides: Override[]) {
+  return Object.assign({}, base, ...overrides);
```
atruskie, line 7: i assume there's no nesting in Params so we don't have to worry about deep merging?

src/app/helpers/parameters/merge.spec.ts
atruskie, line 1: misisng tests?

src/app/services/baw-api/api-common.ts
```diff
@@ -125,18 +126,33 @@ export interface ApiFilter<M extends AbstractModel, P extends any[] = []> {
...
    * @param urlParameters URL parameter values
    */
-  show(model: M | I, ...urlParameters: P): Observable<M>;
+  show(model: M | IdOr<M>, ...urlParameters: P): Observable<M>;
+}
+
+/**
+ * An endpoint that groups models given a filter condition.
+ * These endpoints do not return database-backed models. Therefore they cannot
+ * have an id.
```
atruskie, line 142: this is only semi true, might have to revisit this comment later when API complete

atruskie (approved):
Great work - this was an especially big changeset but hopefully worth it due to resolving some tech debt

src/app/components/shared/map/weightedSuperClusterAlgorithm.ts
hudson-newey: This `weighterSuperClusterAlgorithm` attaches the `clusterWeight` to the `MarkerElement`.

I need to do this because the `ClusterRenderer` only works with `HTMLElement` markers, meaning that the clusterWeight, and custom marker properties are not serialized onto the `HTMLElement` by the default super cluster algorithm.

---

The `?? 1` default should probably not belong in both places and I should remove the `?? 1` in the clusterer so that the default is onlt set in one place

src/app/components/shared/map/weightedSuperClusterAlgorithm.ts
hudson-newey: TL;DR: It serializes the custom `clusterWeight` property from out custom MarkerOptions onto the element.

I need to serialize the `clusterWeight` onto the element because the `ClusterRenderer` works on elements.

https://github.com/QutEcoacoustics/workbench-client/pull/2462#discussion_r2467545249

---

This `count` is the total weight of markers inside of the cluster

src/app/components/shared/audio-event-card/annotation-event-card.component.scss
hudson-newey: image

Agreed, I don't like this functionality. I think that we should merge some patches upstream toinclude the axes labels in the box model content.

Related to: https://github.com/ecoacoustics/web-components/issues/414

---

By contrast, I don' think _all_ of this should be merged upstream.

image

Most of these styles exist so that the components expand to 100% height of their parent.

By default I like to keep the web components conforming to the `block` display/layout mental model, meaning that they will fill their parents width and the height will be determined by the content (typically the minimum size of the spectrogram in this case).

If I merged this upstream, the display/layout mental model would become "custom" because (from my knowledge) no other (native) HTML elements behave like this. Conforming to standard-like behavior makes the mental model simpler imo.

---

That being said, these styles do highlight a few shortcomings of our current css behavior.

1. The axes box model content should include axes labels + title

2. Nested web components should inherit their parents display model so that we don't have to manually set the display + height on each of the axes, indicator, and spectrogram components

E.g. This code should really be

```css
.axes {
  display: block;

  position: relative;
  width: 100%;
  height: 100%;
}
```

src/app/components/shared/map/clusterRenderer.ts
hudson-newey: For context: most of this `ClusterRenderer` comes from default `ClusterRenderer` provided by the Google Maps team.

https://github.com/googlemaps/js-markerclusterer/blob/8acb046d9b5/src/renderer.ts#L107

I'll be happy to pull this out to a CSS file, but just want to document why it's like this.

---

See the notice at the top of the class/file

image

src/app/components/annotations/pages/annotation-map/annotation-map-parameters.resolver.ts
hudson-newey: I'm quite keen of keeping the resolver as an interface / using the angular `ResolverFn` type because it aligns closer with Angular conventions.
Note we already have a `BawResolver` helper for resolving route parameters -> models, but it's not used here because it's impl is too specific to the route parameter -> model functionality.

IMO instead of creating another abstraction, I'd like to investigate (just spit balling here) removing our custom resolver format (the `showOptional` + `providers` object, etc...) and instead just provide the resolver directly to our `StrongRoute`s.

That way we can reduce the number of custom parts and conform more to the Angular data resolver patterns.

https://angular.dev/guide/routing/data-resolvers

---

As an example, I'd like this resolver to look **something** like this

```ts
// annotation-map-parameters.resolver.ts

@Injectable({ providedIn: "root" })
export class AnnotationMapParametersResolver implements Resolve<AnnotationMapParameters> {
  public resolve() {
	const route = inject(ActivatedRouteSnapshot);
    return new AnnotationMapParameters(route.queryParams);
  }
}
```

```ts
// annotation.routes.ts

export const annotationMapRoute = annotationsRoute.add("/map", {
	resolve: {
		annotationMapParameters: AnnotationMapParameterResolver,
	},
})
```

---

Note: The above example still uses the old interface-based resolvers, but the same statement could be used for function-based resolvers

hudson-newey: For an example of why I'd like to simplify before abstracting, take how I mock these revolvers as an example

https://github.com/QutEcoacoustics/workbench-client/blob/event-maps/src/app/components/annotations/pages/search/search.component.spec.ts#L99-L112

I could probably create a helper here to simplify mocking these resolvers, but I think it'd be much simpler (and more Angular-like) just to do something like this

```ts
data: {
    project: routeProject,
    region: routeRegion,
    site: routeSite,
    searchParameters: mockSearchParameters,
    verificationParameters: new VerificationParameters(),
},
```

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: Discussed offline, this was a typo. `true` was supposed to be `false`

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.ts
hudson-newey: Discussed offline, this is because I don't want to include the verification filters when we join the search form with the verification form. Otherwise there'd be two verification status filter inputs

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
hudson-newey: This was actually purposeful, I'm open to discuss

I typically like to assign first, then throw the warning with the new updated value.

I do this so that if the assignment (to `statusKey`) fails (for some unknown reason), it'll log out the realized value of what it really falled back to instead of what it intended to fall back to.

Admittedly, it's not very important in this case because assigning to a constant is very clearly not going to fail.
However, it's a pattern that I use especially because setters/object property assignment is a bit more convoluted/likely to fail in unexpected ways if you didn't track down every code path.

I'm happy to change, but just letting you know this beaks my coding style and creates inconsistencies in my coding style.

---

You've seen a lot more code than me, so if you have any insights, I'd love to hear it :rocket:

src/app/components/shared/event-map/event-map.component.html
hudson-newey: I agree, and tried to implement this suggestion

However, to get this working I had to break the CSS encapsulation of the `google-maps` element (note `::ng-deep`)

image

IMO I think attribute-based styles is better than breaking encapsulation

---

I'll add a code comment as to why these aren't in css files

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
atruskie: Logging out before and after values is good.

Two conditions is bad. (just because the inline conditional is not an if statement, does not mean it is not a branching instruction). I typically consider it good practice - though probably moot from an efficiency perspective - to reduce branching statements. Helps predictive compilers optimise.

Mostly though: when reading it quickly, I misread it. It took me two reads to understand. My first comment was "missing assignment for key validity).

src/app/components/shared/map/weightedSuperClusterAlgorithm.ts
atruskie: image

image


It's the same method with the same name in two spots...

src/app/components/annotations/components/verification-form/verification-parameters.resolver.ts
hudson-newey: Related issue: https://github.com/QutEcoacoustics/workbench-client/issues/2489

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
hudson-newey: Here's the updated code, I'll commit this code to main

image

---

I don't 100% agree, but that's ok

I do agree on the double conditional being harder to read

src/app/components/shared/map/clusterRenderer.ts
hudson-newey: See comment above: https://github.com/QutEcoacoustics/workbench-client/pull/2462#discussion_r2467642800

src/app/components/shared/map/weightedSuperClusterAlgorithm.ts
hudson-newey: > It's the same method with the same name in two spots...

Good point, I probably should refactor this out to a common function

PR #2494, opened 2025-10-30
Refine verification grid navigation confirmation

### LINKED ISSUE #2487: Rescope how the verification grids navigation confirmation should work
We currently have a warning prompt when the user tries to navigate away from the verification grid after a decision has been made.

We originally did this because we did not have the ability to resume verification tasks.

However, with the introduction of the "unverified-by-me" search filter, it is possible to pick up where you left a verification task.

This brings into question if we still need this confirmation.

I think it's still useful because:

1. It's a a shortcut heavy page where it's easy to accidentally navigate away from.
2. We should still show this "unsaved progress" confirmation if the verification/correction API request hasn't completed before closing the web page. Alternatively, we could use a service worker to ensure that incomplete requests complete.

hudson-newey:
# Refine verification grid navigation confirmation

## Changes

- When all verifications are saved, you now get a more generic "Are you sure you want to leave this page?" instead of the scary "Changes to this page will be lost! Are you sure you want to leave?" message
- When not all verification requests have completed, you now get the more descriptive "Some changes are still being saved. Please wait one moment." message
- Adds new `confirmation.guard` that can be used to show a confirmation/alert dialog on pages with a custom message

### Dev Concerns

- Adds new `SignalOr<T>` and `unwrapPotentialSignal` helpers that can be used to and unwrap inputs/values that can either be signals or regular variables. I suspect this will be useful in the future for directives as well

## Issues

Fixes: #2487

## Visual Changes

image

_New navigation warning that is shown if there are no pending requests_

---

image

_New warning that is shown if there are pending requests_

---

## Final Checklist

### DIFF
--- a/src/app/components/annotations/pages/verification/verification.component.ts
+++ b/src/app/components/annotations/pages/verification/verification.component.ts
@@ -2,6 +2,7 @@ import {
   AfterViewInit,
   ChangeDetectionStrategy,
   Component,
+  computed,
   CUSTOM_ELEMENTS_SCHEMA,
   ElementRef,
   inject,
@@ -21,7 +22,7 @@ import { Region } from "@models/Region";
 import { Site } from "@models/Site";
 import { ActivatedRoute, Router } from "@angular/router";
 import { Location } from "@angular/common";
-import { firstValueFrom, map } from "rxjs";
+import { firstValueFrom, map, tap } from "rxjs";
 import { annotationMenuItems } from "@components/annotations/annotation.menu";
 import { Filters, Paging, Sorting } from "@baw-api/baw-api.service";
 import {
@@ -31,7 +32,7 @@ import {
 import { StrongRoute } from "@interfaces/strongRoute";
 import { NgbModal, NgbTooltip } from "@ng-bootstrap/ng-bootstrap";
 import { SearchFiltersModalComponent } from "@components/annotations/components/modals/search-filters/search-filters.component";
-import { UnsavedInputCheckingComponent } from "@guards/input/input.guard";
+import { WithNavigationConfirmation } from "@guards/confirmation/confirmation.guard";
 import { ShallowAudioEventsService } from "@baw-api/audio-event/audio-events.service";
 import { AudioEvent } from "@models/AudioEvent";
 import { AnnotationService } from "@services/models/annotations/annotation.service";
@@ -78,6 +79,11 @@ interface PagingContext extends PageFetcherContext {
   page: number;
 }
 
+const enum NavigationMessages {
+  default = "Are you sure you want to leave this page?",
+  unsavedChanges = "Some changes are still being saved. Please wait one moment.",
+}
+
 const projectKey = "project";
 const regionKey = "region";
 const siteKey = "site";
@@ -108,7 +114,7 @@ const confirmedMapping = {
 })
 class VerificationComponent
   extends PageComponent
-  implements OnInit, AfterViewInit, UnsavedInputCheckingComponent
+  implements OnInit, AfterViewInit, WithNavigationConfirmation
 {
   private readonly audioEventApi = inject(ShallowAudioEventsService);
   private readonly verificationApi = inject(ShallowVerificationService);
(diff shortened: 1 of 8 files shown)


src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -140,7 +146,23 @@ class VerificationComponent
...
-  public readonly hasUnsavedChanges = signal(false);
+  /**
+   * A count of how many requests are currently waiting for a response.
+   * We use this to determine if we need to show an "unsaved changes" navigation
+   * warning if all verification/tagging decisions have not been processed yet.
+   */
+  private readonly pendingRequests = signal(0);
+
+  public confirmNavigation = false;
+  public readonly confirmNavigationMessage = computed<NavigationMessages>(() =>
```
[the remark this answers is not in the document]
hudson-newey: Wrong, `NavigationMessage` should be a subset of `string`, meaning that this **component** type narrows to a more specific discriminated union of strings

hudson-newey:
Something I also changed in this PR is I changed the `DecisionOptions` from an `enum` to `const enum`

`const enum`'s are much better because they don't have the reverse mapping overhead seen in regular enums

image

I wanted to point this out so if I incorrectly use a regular `enum` in the future you can catch / correct me.

https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time

---

A caveat here is that `const enum` doesn't work very nice with esbuild/Vite across modules

https://vite.dev/guide/features.html#isolatedmodules

src/app/guards/confirmation/confirmation.guard.ts
```diff
@@ -0,0 +1,84 @@
...
+   * Whether to hard block navigation without confirmation.
+   *
+   * If set to `true`, navigation will be blocked, and just show an alert
+   * instead of a confirmation dialog.
+   * This should only be used in extreme cases where navigation must not be
+   * allowed or we need to wait for something to complete on the main thread.
+   *
+   * @default false
+   */
+  confirmationHardBlock?: PotentiallySignal<boolean>;
```
atruskie:
```suggestion
  blockNavigation?: PotentiallySignal<boolean>;
```

src/app/guards/confirmation/confirmation.guard.ts
```diff
@@ -0,0 +1,84 @@
...
+  confirmNavigationMessage?: PotentiallySignal<string>;
+
+  /**
+   * Whether to hard block navigation without confirmation.
+   *
+   * If set to `true`, navigation will be blocked, and just show an alert
+   * instead of a confirmation dialog.
+   * This should only be used in extreme cases where navigation must not be
+   * allowed or we need to wait for something to complete on the main thread.
+   *
```
atruskie, line 39: ...because spamming alerts won't work either due to browser restrictions

src/app/guards/confirmation/confirmation.guard.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  confirmNavigation?: PotentiallySignal<boolean>;
+
+  /**
+   * A custom message to show in the confirmation dialog.
+   * If not provided, a default message will be used.
+   *
+   * @default "Are you sure you want to leave this page?"
+   */
+  confirmNavigationMessage?: PotentiallySignal<string>;
```
atruskie: i probably would have opted for a monadic type over an extra signature, but that's just a style thing

src/app/helpers/signals/signals.ts
```diff
@@ -0,0 +1,22 @@
+import { isSignal, Signal } from "@angular/core";
+
+/**
+ * A type that can either be a direct value of type T or a Signal that
+ * produces a value of type T.
+ * This is useful for route guards and directives that want to work in static,
+ * zone.js, and signal-based components/environments.
+ */
+export type PotentiallySignal<T> = T | Signal<T>;
```
atruskie: I'd have chosen `SignalOr<T>`, maybe `SignalOrScalar<T>`.

IDM `PotentialSignal` , though I do think the adverb is a bit much

atruskie (approved):
consider:

```
private trackPendingRequests(observable) {
   this.pendingRequests.update((value) => value + 1);
   
   observable().pipe().finalize(() => {
        this.pendingRequests.update((value) => value - 1);
    })
}
```

Packages all state, wrappers, and side effects into one function

src/app/helpers/signals/signals.ts
hudson-newey: I agree, this also follows the precedent set in the `IdOr<T>` type

https://github.com/QutEcoacoustics/workbench-client/blob/276a57400b1943ba5b3f84f1d6698bc3292a5609/src/app/services/baw-api/api-common.ts#L13

---

I originally chose this name to mimic our existing `PotentiallyAsync` type helper

https://github.com/QutEcoacoustics/workbench-client/blob/276a57400b1943ba5b3f84f1d6698bc3292a5609/src/app/helpers/advancedTypes.ts#L64-L65

---

I do like the prefer verbose `SignalOr<T>` type

src/app/guards/confirmation/confirmation.guard.ts
hudson-newey: image

hudson-newey:
> consider:
>
> private trackPendingRequests(observable) {
>    this.pendingRequests.update((value) => value + 1);
>   
>    observable().pipe().finalize(() => {
>         this.pendingRequests.update((value) => value - 1);
>     })
> }
>
> Packages all state, wrappers, and side effects into one function

image

PR #2520, opened 2025-11-18
Don't emit initial empty filters from date/time filter component

hudson-newey:
# Don't emit initial empty filters from date/time filter component

Do not emit an initial filter update from the date/time filter component if there are no filter conditions.

This was causing some components to update twice because the filter component would cause the filter conditions to update, even when nothing changed.

## Changes

- Fix annotation search page performing initial query twice

## Final Checklist


hudson-newey:
@atruskie I'm going to hold off on merging this until you provide confirmation that this is something that we should fix

atruskie (approved):

PR #2521, opened 2025-11-18
Fix verification parameter model parsing url tags

### LINKED ISSUE #2512: verification `taskTag` incorrectly cached between navigations

hudson-newey:
# Fix verification parameter model parsing url tags

What is the purpose of this PR?

## Changes

- Removes `tags` from verification parameter model so that there is only one source of truth for the filtered tags (now `annotationSearchParameters`)
- Refactors `tagPriority` into a `verification.component` method so that the parameter models can work independently and without knowledge of the other
- Changes verification search form component change detection strategy from `OnPush` to `Default` so that associations can resolve in the template
- Adds `searchFilters` to verification form component so that it can use the `searchFilters().tagModels` as the "tags to verify" placeholder value

## Issues

Fixes: #2512

## Visual Changes

If the PR has any visual changes to the website, post pictures of the new pages and how they look. Label any images if more than one is given.

## Final Checklist

### DIFF
--- a/src/app/components/annotations/components/verification-form/verification-form.component.html
+++ b/src/app/components/annotations/components/verification-form/verification-form.component.html
@@ -3,8 +3,8 @@
     id="task-tag-input"
     label="Tag to Verify"
     [inputPlaceholder]="
-      verificationParameters().tagModels.length > 0
-        ? $any(verificationParameters().tagModels)
+      searchParameters().tagModels.length > 0
+        ? $any(searchParameters().tagModels)
         : 'First Tag'
     "
     [searchCallback]="tagsApi.typeaheadCallback()"
--- a/src/app/components/annotations/pages/verification/verification.component.ts
+++ b/src/app/components/annotations/pages/verification/verification.component.ts
@@ -21,7 +21,6 @@ import { Project } from "@models/Project";
 import { Region } from "@models/Region";
 import { Site } from "@models/Site";
 import { ActivatedRoute, Router } from "@angular/router";
-import { Location } from "@angular/common";
 import { firstValueFrom, map, Observable } from "rxjs";
 import { annotationMenuItems } from "@components/annotations/annotation.menu";
 import { Filters, Paging, Sorting } from "@baw-api/baw-api.service";
@@ -127,7 +126,6 @@ class VerificationComponent
   private readonly modals = inject(NgbModal);
   private readonly route = inject(ActivatedRoute);
   private readonly router = inject(Router);
-  private readonly location = inject(Location);
   private readonly config = inject(ConfigService);
   private readonly injector: AssociationInjector = inject(ASSOCIATION_INJECTOR);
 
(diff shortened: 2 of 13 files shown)


src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -519,8 +530,7 @@ class VerificationComponent
       verificationParameters,
     );
 
-    const urlTree = this.router.createUrlTree([], { queryParams });
-    this.location.replaceState(urlTree.toString());
+    this.router.navigate([], { queryParams });
```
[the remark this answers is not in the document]
hudson-newey: This is exactly what we want. I'll add a comment explaining why I haven't used `replaceUrl`

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -392,6 +386,23 @@ class VerificationComponent
     }
   }
 
+  /**
+   * @description
+   * An ordered array of tag IDs representing what tags should be prioritized
+   * when displaying the tag that is being verified.
```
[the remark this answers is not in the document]
hudson-newey: We are using TypeScript

Using JSDoc return types adds maintenance overhead for little to no benefit over TypeScript return type annotations (which we do have)

I'm not opposed to JSDoc return documentation if it helps documenting the return type more than typing.

However, this suggested comment does not provide any additional insight over the JSDoc description section

src/app/components/annotations/components/verification-form/verification-form.component.html
```diff
@@ -3,8 +3,8 @@
     id="task-tag-input"
     label="Tag to Verify"
     [inputPlaceholder]="
-      verificationParameters().tagModels.length > 0
-        ? $any(verificationParameters().tagModels)
+      searchParameters().tagModels.length > 0
+        ? $any(searchParameters().tagModels)
         : 'First Tag'
```
atruskie, line 8: unrelated: but 'first tag' isn't really accurate is it?

src/app/components/annotations/pages/verification/verification.component.ts
```diff
@@ -392,6 +391,23 @@ class VerificationComponent
     }
   }
 
+  /**
+   * @description
+   * An ordered array of tag IDs representing what tags should be prioritized
```
atruskie, line 396:
```suggestion
   * Returns an ordered array of tag IDs representing what tags should be prioritized
```

src/app/components/annotations/components/verification-form/verification-form.component.html
hudson-newey: If there is no "tag to verify" and there are no "search parameter tags", we select the first tag on the `AudioEvent` model.

I'm very happy to change the copy here because I'll admit it's not very clear

Additionally, the "Tag to verify" label is not very descriptive imo

image

I think it's easy to be confused because in my mind this sounds like the filter for what events will be shown

src/app/components/annotations/pages/verification/verification.component.ts
hudson-newey: @copilot Open a new pull request with this change

src/app/components/annotations/components/verification-form/verification-form.component.html
atruskie: But don't we have a tag priority algorithm?

PR #2526, opened 2025-11-19
Add support for new projection fields

### LINKED ISSUE #2525: Support new projection api fields
The API now supports new `only`, `add`, and `remove` fields in the projection body

https://github.com/QutEcoacoustics/baw-server/blob/master/app/modules/filter/projection.rb#L11

We should adjust our TypeScript typings to support these fields.

Note that the `NEW_KEYS` cannot be used with the `OLD_KEYS`, meaning that this is an XOR relationship.

https://github.com/QutEcoacoustics/baw-server/blob/master/app/modules/filter/projection.rb#L184

hudson-newey:
# Add support for new projection fields

I'm going to wait for a review from @atruskie until I merge this PR so that I can be sure my understanding is correct.

## Changes

- The filter "projection" field can now take either a new projection object or an old projection object (XOR relationship).

## Issues

Fixes: #2525

## Final Checklist

### DIFF
--- a/src/app/services/baw-api/baw-api.service.ts
+++ b/src/app/services/baw-api/baw-api.service.ts
@@ -1099,6 +1099,9 @@ export type InnerFilter<Model = unknown> = Combinations<Writeable<Model>> &
       Subsets;
   };
 
+/**
+ * @see https://github.com/QutEcoacoustics/baw-server/wiki/API:-Spec#projection-for-resource
+ */
 interface NewProjection<Model = unknown, K extends keyof Model = keyof Model> {
   /**
    * The `only` parameter specifies the base set of keys to include.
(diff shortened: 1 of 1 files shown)


src/app/services/baw-api/baw-api.service.ts
```diff
@@ -1099,21 +1099,48 @@ export type InnerFilter<Model = unknown> = Combinations<Writeable<Model>> &
...
+   * This means that `add` and `remove` operations will be performed on top of
+   * the keys specified in `only`.
+   *
+   * If `add` and `remove` are not specified, only the keys in `only` field will
+   * be included in the response.
+   */
+  only?: K[];
+
+  /**
+   * All default keys not filtered out by `remove` will be included in the
```
atruskie: Remove is applied last.

The wording here is strange.


I wouldn't mention remove in the add stage.

It's more like "adds additional fields to the projection, on top of the default/only fields"

atruskie (approved):
I'd harmonize with the docs for your code comments, but essentially this is all correct.

https://github.com/QutEcoacoustics/baw-server/wiki/API:-Spec#projection-for-resource

PR #2533, opened 2025-11-20
Fix region licenses on hidden project environments

### LINKED ISSUE #2421: License not showing up for region cards
**Describe the bug**

On the regions list page the "No License" is being shown even though a license is set for the project.

**To Reproduce**

1. Set a license on a project
2. Go to `/regions`
3. Observe the no license tag

**Expected behavior**

The license tag should show the license from the project.

**Screenshots**

Image

But it is set on the project:

Image

Image

hudson-newey:
# Fix region licenses on hidden project environments

## Changes

- Make the region model `Region.license` getter return an async promises that uses the injected projects service to fetch the project model license (this is not as bad as it sounds because our caching is strong enough so that on single project sites like the A2O that hide projects, there is only 1 request for the project model)
- `cards.component` grid now tracks cards by `{Project|Region}.id` so that if your new search result contains a subset of the previous result page, it doesn't need to emit another filter request to see if the project/region `hasAudio$`
- Migrate `card.component` to signals
- `project.service` `getProjectFor` now uses the `project.service` `show` (GET) requests instead of a filter request for improved caching

## Issues

Fixes: #2421

## Visual Changes

image

_The A2O project on the new client version showing the CC-BY-4.0 license badge_

Note: Images are now showing up because of CORS errors, but the images still do work

## Final Checklist


src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,45 +38,44 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
+  );
 
-  public ngOnInit(): void {
-    this.isOwner = this.model.creatorId === this.session.loggedInUser?.id;
-    this.hasNoAudio$ = this.getRecordings().pipe(
-      map((recordings): boolean => recordings.length === 0),
-    );
-    this.updateLicense();
+  protected readonly hasNoAudio$: Observable<boolean> =
+    this.getRecordings().pipe(map((recordings) => recordings.length === 0));
```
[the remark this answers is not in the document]
hudson-newey: I don't think this is an issue, although, I'll revert the change just to be safe

hudson-newey:
I'm holding off on this PR being merged until the ng 20 upgrade is complete because there's a new `resource` API for async resolving signals meaning that I wouldn't need the `effect` hack for resolving models

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,45 +38,44 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
+
+  public constructor() {
+    effect(async () => {
+      const modelIdentifier = await this.model().license;
+      this.license.set(modelIdentifier);
```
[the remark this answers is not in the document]
hudson-newey: This is fixed with the `resource` signal which we gained access to from the ng20 update

src/app/components/shared/model-cards/card/card.component.html
```diff
@@ -0,0 +1,68 @@
...
+          container="body"
+        >
+          <!--
+            TODO: We can remove this async pipe by making licenseText a resource
+            signal once we upgrade the Angular 20.
+
+            We can't do this yet because the resource signal API in Angular
+            19 does not support the "params" option.
+          -->
+          {{ licenseText() | async }}
```
atruskie: stale comment?

atruskie: and pipe?

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,43 +37,43 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
   private readonly session = inject(BawSessionService);
   private readonly licenseService = inject(LicensesService);
 
-  @Input() public model: Project | Region;
+  public readonly model = input<Project | Region>();
 
-  public hasNoAudio$: Observable<boolean>;
-  protected isOwner: boolean;
-  protected licenseText: string | undefined = undefined;
+  protected readonly license = resource({
```
atruskie: i wonder if we should suffix this. `license` is not a signal or a value... it's a wrapper for fetching the value.

`licenseResource`?

Angular docs do it: https://angular.dev/guide/signals/resource

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,43 +37,43 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
+    params: () => ({ model: this.model() }),
+    loader: async ({ params }) => {
+      return await params.model.license;
```
atruskie: OK this took me a little while to figure out. You added a license getter on the model that returned a promise.

It's a neat trick, but given I don't think we have promise returning values almost anywhere in our models, I think we should avoid this for the moment.


Suggestion: 

Add a `getLicense` method to the `projects` service that accepts a `IdOr<Project>`. You can use a filter request with `only` to literally just get the license.

Or: just get invoke `show` on the project service. The request is likely already cached (because of the association fetcher) so it should be a noop - which is pretty much what you've done anyway (great!) but without introducing new conventions in our models.

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,43 +37,43 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
-      map((recordings): boolean => recordings.length === 0),
-    );
-    this.updateLicense();
-  }
+  protected readonly licenseText = computed(() =>
+    this.licenseService.licenseText(this.license.value()),
+  );
+
+  protected readonly isOwner = computed(
+    () => this.model().creatorId === this.session.loggedInUser?.id,
```
atruskie, line 82: technically this is wrong

src/app/models/Region.ts
```diff
@@ -91,17 +92,24 @@ export class Region
...
+    // TODO: Use the "project" association once we support async associations
+    // see: https://github.com/QutEcoacoustics/workbench-client/issues/2148
+    const projectsApi = this.injector.get(PROJECT.token);
+    const request$ = projectsApi
+      .show(this.projectId)
+      .pipe(map((project: Project) => project.license ?? null));
+
+    return firstValueFrom(request$);
+  }
+
```
atruskie: See comments above. Would suggest moving todo comment into invocation in the resource

atruskie (changes requested):

src/app/components/shared/model-cards/card/card.component.ts
hudson-newey: Good catch

Believe it or not, this is an existing bug that I'd say is out of scope for this PR

https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/shared/model-cards/card/card.component.ts#L106

hudson-newey: Associated issue: https://github.com/QutEcoacoustics/workbench-client/issues/2584

src/app/components/shared/model-cards/card/card.component.ts
hudson-newey: We already have a `getProjectFor` method on the `project.service`

https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/services/baw-api/project/projects.service.ts#L91

So what do you think about using this service in the resource instead?

image

Note: This is not final code, and I'll need to clean it up, but the code above will give you an indication of what I am thinking

hudson-newey: The `getProjectFor` method also brings up a good point that the regions might have multiple projects.

Although, I don't think this should be a realistic concern because if you have `hideProjects` enabled, I think there's an assumption that there will only be 1 project in the entire instance

atruskie: Regions only belong to one project

atruskie: `getProjectFor` seems like the right utility method, but it could use some love.

- it could accept a `Project` model so it just routes back to `show`
- and the Region branch should just be using `show` as well, and should not be using `siteIds`???

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,43 +38,65 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
+      );
+
+      return firstValueFrom(projectLicense);
+    },
+  });
+
+  protected readonly licenseTextResource = resource({
+    params: () => ({ license: this.licenseResource.value() }),
+    loader: async ({ params }) => {
+      return await this.licenseService.licenseText(params.license);
```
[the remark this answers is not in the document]
hudson-newey: This is why we need to enable strict TypeScript type checking...

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,43 +38,65 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
+      // before accessing the license property so that if no project is found,
+      // we do not throw an error.
+      //
+      // I also use nullish coalescing to return null if no license is found
+      // so that we don't mix undefined and null return types when there are no
+      // projects associated with the region.
+      const projectLicense = this.projectService.getProjectFor(cardModel).pipe(
+        map((project) => project[0]?.license ?? null),
+      );
+
```
atruskie, line 72: two points:

- regions can't be orphaned
- and why don't you just pick the license over the project array. That works in the empty array case too, as well as the case where one project may have a license and another may not

src/app/components/shared/model-cards/card/card.component.ts
```diff
@@ -90,43 +38,65 @@ import { WithLoadingPipe } from "../../../../pipes/with-loading/with-loading.pip
...
+      return firstValueFrom(projectLicense);
+    },
+  });
+
+  protected readonly licenseTextResource = resource({
+    params: () => ({ license: this.licenseResource.value() }),
+    loader: async ({ params }) => {
+      return await this.licenseService.licenseText(params.license);
+    },
+  });
```
atruskie: seems like this could be combined with licenseResource.

do async await. No reason to have more signals if we don't need to branch off of them

atruskie (approved):

src/app/components/shared/model-cards/card/card.component.ts
hudson-newey: image

src/app/components/shared/model-cards/card/card.component.ts
hudson-newey: image

PR #2549, opened 2025-11-27
Add verification information to audio event cards

### LINKED ISSUE #2333: Annotation search page should show verification status
The `<audio-event-card>`'s component should show the verification state

### LINKED ISSUE #2518: `PaginationTemplate` should support `projection` input
Similar to how we can include a callback for `filter` and `sorting`

hudson-newey:
# Add verification information to audio event cards

## Changes

- Adds `verificationSummary` and `verificationIds` to `AudioEvent` model
	- Adds `VerificationSummary` sub-model
	- Adds `Consensus` sub-(data)-model
- Adds `defaultProjection` option to `PaginationTemplate` (needed to include `verification_summary` in search page filters)
- Adds `d3-scale` as a dependency

## Issues

Fixes: #2333 (Annotation search page should show verification status)
Fixes: #2518 (PaginationTemplate should support projection input)

## Visual Changes

image

_Annotation event cards showing "correct", "incorrect", and "no verifications"_

---

image

_Tooltip when hovering over a tag_

---

image

image

_A verifications with only 66% consensus ratio, meaning that its color is slightly darker_

## Final Checklist


src/app/components/shared/audio-event-card/annotation-event-card.component.html
```diff
@@ -29,13 +29,70 @@
...
+        @for (tag of tagInfo(); track tag.id; let isFirst = $first) {
+          <div
+            class="tag-item"
+            [ngbTooltip]="tagVerifications"
+            [tooltipContext]="{ verificationSummary: tag.verificationSummary }"
+          >
+            @if (isFirst) {
+              <fa-icon class="me-2" [icon]="['fas', 'tags']" />
+            } @else {
+              <span class="me-4"></span>
```
atruskie: why not use text-indent for this?

src/app/components/shared/audio-event-card/annotation-event-card.component.html
```diff
@@ -29,13 +29,70 @@
...
+              Additionally, if there are 5 verifiers, a 4-1 vote will also show
+              a verified icon (80% consensus), but a 3-2 vote (60% consensus)
+              will not show an icon (which seems reasonable).
+
+              TODO: I'm sure there's some UX research on what consensus ratio
+              is appropriate to show verification ticks, however, I have not
+              looked for this research yet.
+            -->
+            <span class="verified-icon me-1" [style]="{ color: tag.color }">
+              @if (tag.verificationSummary.consensus.ratio > 0.66) {
```
atruskie: pull out the number to the component, and docs

src/app/components/shared/audio-event-card/annotation-event-card.component.html
```diff
@@ -29,13 +29,70 @@
...
+                }
+              } @else {
+                <fa-icon
+                  class="verification-status-icon"
+                  [icon]="['fas', 'circle-question']"
+                />
+              }
+            </span>
+            <a class="tag-link" [bawUrl]="tag.viewUrl">{{ tag.text }}</a>
+          </div>
```
atruskie, line 62: looks neat and clean. I'm worried though that a majority-consensus (where the goal posts kinds move based on which category has the most votes is confusing).

I think we should move to a simpler two vote value we show: $correct / (incorrect + correct)$.
That we always express a consistent value to users. It also makes showing icons a bit easier IMO: `(0, 33]` is a cross, `[0.66, 1)` is a tick, and anything else is a question mark

src/app/components/shared/audio-event-card/annotation-event-card.component.html
```diff
@@ -29,13 +29,70 @@
...
-      />
+              Additionally, if there are 5 verifiers, a 4-1 vote will also show
+              a verified icon (80% consensus), but a 3-2 vote (60% consensus)
+              will not show an icon (which seems reasonable).
+
+              TODO: I'm sure there's some UX research on what consensus ratio
+              is appropriate to show verification ticks, however, I have not
+              looked for this research yet.
+            -->
+            <span class="verified-icon me-1" [style]="{ color: tag.color }">
```
atruskie, line 44: I'd like to see an empty circle (no icon, no fill, with border) that represents no data)

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
```diff
@@ -50,6 +63,42 @@ export class AnnotationEventCardComponent {
...
+  protected readonly undecidedColor = "#555555";
+  protected readonly ConsensusDecision = ConsensusDecision;
+
+  protected readonly tagInfo = computed<TagInfo[]>(() => {
+    return this.annotation().tags.map((tagModel) => {
+      // Audio events without any verifications return "null" instead of an
+      // object.
+      // see: https://github.com/QutEcoacoustics/baw-server/issues/869
+      let verificationSummary = (
+        this.annotation().verificationSummary ?? []
```
atruskie: this normalization can happen in the annotation model right?

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
```diff
@@ -1,25 +1,37 @@
...
+import { isInstantiatedPipe } from "@pipes/is-instantiated/is-instantiated.pipe";
+import { LoadingComponent } from "@shared/loading/loading.component";
+import { scaleLinear } from "d3-scale";
 import { IsUnresolvedPipe } from "../../../pipes/is-unresolved/is-unresolved.pipe";
+import { ZonedDateTimeComponent } from "../datetime-formats/datetime/zoned-datetime/zoned-datetime.component";
+
+interface TagInfo {
+  id?: number;
+  text?: string;
+  viewUrl: string;
```
atruskie: maybe just pass a tag object in rather than destructuring it?

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
```diff
@@ -50,6 +63,42 @@ export class AnnotationEventCardComponent {
...
+          tagId: tagModel.id,
+          count: 0,
+          correct: 0,
+          incorrect: 0,
+          unsure: 0,
+          skip: 0,
+        });
+
+        verificationSummary = noVerificationSummary;
+      }
```
atruskie: i'm of two minds on this: i'd push as much of this as I can to the model

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
```diff
@@ -50,6 +63,42 @@ export class AnnotationEventCardComponent {
   private readonly spectrogram =
     viewChild<ElementRef<SpectrogramComponent>>("spectrogram");
 
+  protected readonly undecidedColor = "#555555";
```
atruskie: why is this the only color out here by itself?

src/app/models/AudioEvent/Consensus.ts
atruskie, line 1: Do you expect to use this in multiple places?

src/app/models/AudioEvent/Consensus.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export class Consensus extends AbstractData<IConsensus> {
+  /** Total number of "correct" verifications */
+  private readonly correct: number;
+  /** Total number of "incorrect" verifications */
+  private readonly incorrect: number;
+
+  // Returns the consensus decision based on the highest vote count
+  // If all counts are 0, returns None
+  public get decision(): ConsensusDecision {
```
atruskie: based on prior comments we can remove the decision getter.

we can add a `correctConsensus` getter.

We can leave room for other forms of consensus later.

src/app/models/AudioEvent.ts
```diff
@@ -79,6 +91,13 @@ export class AudioEvent
   @bawDateTime()
   public readonly deletedAt?: DateTimeTimezone;
 
+  // These fields are not included in the standard response, and must be
+  // explicitly added via the `projection.add` filter.
+  public readonly verificationIds?: CollectionIds;
+
+  @bawSubModelCollection(VerificationSummary)
+  public readonly verificationSummary?: VerificationSummary[];
```
atruskie: we can normalize this on assignment

atruskie (approved):
great work

src/app/components/shared/audio-event-card/annotation-event-card.component.html
hudson-newey: Great idea! For some reason I forgot about text-indent

Thanks for the pointer

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
hudson-newey: Agree, I like the idea

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
hudson-newey: Correct, it shouldn't be a property.

From memory I originally had all of the colors here before I made the `verificationColor` helper because the color logic was split over multiple methods.

Since I refactored all of the color logic into one method, it's no longer needed here

src/app/models/AudioEvent/Consensus.ts
hudson-newey: No, I just like the one model per file layout because it's easier to find and easier to document (imo)

hudson-newey: I also used a class instead of an object constructed by the `VerificationSummary` so that it has ownership over all of its own data and is easier to create getters

atruskie: So I would have inlined it into the summary class.

It doesn't need to be it's own class, unless you're future proofing. I couldn't rule out you weren't, hence the comment

src/app/models/AudioEvent.ts
hudson-newey: I'm not 100% sure about normalizing here.

I'm no opposed to normalizing the `Annotation` model because it's not representing a response / model from the API.

Changing the model returned by the API seems fishy to me, especially when we decided that `null`/`nil` is the correct `verificationSummary` value for an audio event with zero verifications.

src/app/models/AudioEvent/Consensus.ts
hudson-newey: When you say "inlined into the summary class" do you mean "returning an object from the `consensus` getter

or adding a `confirmedConsensus` getter to the `verificationSummary` (I can mostly agree with the second point), especially since it reduces the likelihood of data desync because there's only one source of truth for the `correct` & `incorrect` counts

Edit: Going off what you said in a following comment, I think you just want to have a `correctConsensus` getter on the `AudioEvent`/`Annotation` model

src/app/models/AudioEvent.ts
atruskie: It's correct for the engineering effort I want to put it in. 

Normalising for utility is ok, especially for read only properties

src/app/components/shared/audio-event-card/annotation-event-card.component.html
hudson-newey: image

image

Ignore the 1px offset problem in the image above, I will fix before merging

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
hudson-newey: Here's what I'm thinking

image

image

src/app/models/AudioEvent.ts
hudson-newey: https://github.com/QutEcoacoustics/workbench-client/pull/2549#discussion_r2575538381

src/app/models/AudioEvent/Consensus.ts
hudson-newey: image

src/app/components/shared/audio-event-card/annotation-event-card.component.html
atruskie: looks good

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
atruskie: I was literally just thinking about converting null to an empty array.

Also why can't  we just use a setter? why another decorator?

src/app/components/shared/audio-event-card/annotation-event-card.component.ts
```diff
@@ -50,6 +58,62 @@ export class AnnotationEventCardComponent {
...
+      const color = this.verificationColor(
+        verificationSummary.correctConsensus,
+        verificationSummary.resolvedDecisionCount,
+      );
+
+      return {
+        tag: tagModel,
+        verificationSummary,
+        color,
+      };
```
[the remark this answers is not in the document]
hudson-newey: The `flatMap` is weird, and you should never have a `verificationSummary` that doesn't have a matching `tag`.

If you do, I'm ok hard failing to prevent showing bad data

PR #2562, opened 2025-12-02
Add import filters to annotation search form

### LINKED ISSUE #2302: Users should be able to filter annotation search / verification grid by annotation import
As part of this, we should also add a "verify annotations" button to the batch import screen

hudson-newey:
# Add import filters to annotation search form

## Changes

- Adds annotation import typeahead to annotation search form
- Adds annotation import file typeahead to annotation search form
- Adds "verify" button to annotation import list view
- Adds `ShallowAudioEventImportFile` service (awaiting upstream API backing). At the moment, this service requires you to provide an array of `AudioEventImport` models that it will apply `filter` conditions over
- Add `withTooltip` input to date/time components to optionally disable the tooltip. This was needed because the file tooltip uses the `baw-zoned-datetime` component to show the created date, but we don't want a nested tooltip
- Fixes test modal selectors using `document.querySelector` instead of `spec.query("", { root: true })`
- Re-format service tokens file. (We suddenly started getting "line too long" linting errors in this file. Which is correct. However, this means that I had to re-format a some long token identifiers. We should probably clean up this file, because it's pretty large & unmaintainable)

## Issues

Fixes: #2302

## Visual Changes

image

_Annotation search form with annotation import & annotation import file inputs_

---

image

_Tooltip for annotation import files containing `createdAt` information_

---

image

_"verify" button on annotation import list page_

## Final Checklist


src/app/services/baw-api/ServiceTokens.ts
```diff
@@ -157,14 +160,23 @@ export const TAG = new ServiceToken<TagsService, Tag>("TAG");
...
 export const MEDIA = new ServiceToken<MediaService, never>("MEDIA");
 export const SITE_SETTINGS = new ServiceToken<SiteSettingsService, SiteSetting>("SITE_SETTINGS");
 export const GROUPED_AUDIO_EVENTS = new ServiceToken<GroupedAudioEventsService, AudioEvent>("GROUPED_AUDIO_EVENTS");
+export const AUDIO_EVENT_IMPORT = new ServiceToken<AudioEventImportService, AudioEventImport>("AUDIO_EVENT_IMPORT");
+export const AUDIO_EVENT_IMPORT_FILE = new ServiceToken<AudioEventImportFileService, AudioEventImportFile>(
+  "AUDIO_EVENT_IMPORT_FILE",
+);
+export const SHALLOW_AUDIO_EVENT_IMPORT_FILE = new ServiceToken<ShallowAudioEventImportFileService, AudioEventImportFile>(
+  "SHALLOW_AUDIO_EVENT_IMPORT_FILE",
+)
```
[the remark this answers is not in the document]
hudson-newey: Not a syntax error, but good catch

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -196,3 +231,39 @@
...
+<ng-template #importFileTooltip let-result="result">
+  <div class="import-file-tooltip">
+    <strong>{{ result.name }}</strong>
+
+    <hr />
+
+    <ul class="data-list">
+      <li>
+        Created:
+        <baw-zoned-datetime [value]="result.createdAt" [withTooltip]="false" />
```
atruskie, line 264: this got wrapped in the screenshot. Maybe make the tooltip a bit wider, or fit to content?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
```diff
@@ -196,3 +231,39 @@
...
+  file name.
+-->
+<ng-template #importFileTooltip let-result="result">
+  <div class="import-file-tooltip">
+    <strong>{{ result.name }}</strong>
+
+    <hr />
+
+    <ul class="data-list">
+      <li>
```
atruskie, line 262: we could also add a link to the source object

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.scss
```diff
@@ -9,3 +9,20 @@
...
+  padding-inline: 1rem;
+
+  // The ngb-tooltip adds some bottom padding by default, so we don't need to
+  // add any padding on the bottom here.
+  padding-top: 1rem;
+
+  .data-list {
+    list-style: none;
+    padding: 0;
+  }
```
atruskie: didn't I see you make this adjustment for a tooltip recently?

maybe a common helper class?

src/app/services/baw-api/audio-event-import-file/audio-event-import-file.service.ts
```diff
@@ -94,20 +102,89 @@ export class AudioEventImportFileService
...
+    filters: Filters<AudioEventImportFile>,
+    eventImports: CollectionIds<AudioEventImport>,
+  ): Observable<AudioEventImportFile[]> {
+    const arrayEventImports = Array.from(eventImports);
+
+    return merge(
+      ...arrayEventImports.map((eventImport) =>
+        this.api.filter(filters, eventImport),
+      ),
+    );
```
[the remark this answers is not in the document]
hudson-newey: A file cannot belong to multiple imports

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -400,20 +432,52 @@ export class AnnotationSearchParameters
...
+      // this is an indication that something has gone wrong.
+      // Therefore, if we encounter any import file filters without any import
+      // filters, we log a warning to help with debugging.
+      if (!hasImportFilters) {
+        console.warn(
+          "Annotation import file filters are expected to be a subset of " +
+            "event import filters. However, no event import filters were provided.",
+        );
+      }
+
```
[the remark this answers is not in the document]
hudson-newey: Copilot is having dreams again.

This is not true.

Additionally, this console warn should NEVER be hit in normal circumstances because the parameter model ctor has validation to check against this occurring.

Additionally, any assignments to `annotationImports` or `importFiles` is validated in the annotation search form component where this parameter model gets its values.

This warning only exists as a defensive programming measure and I've never seen it hit in normal use.

However, I have it just in case because I have not enforced this relationship during TypeScript compile time. If I did enforce this dependency requirement in TypeScript, I'd be happy to remove this warning. However, that is a lot of work for not much benefit.

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -400,20 +432,52 @@ export class AnnotationSearchParameters
...
+        },
+      } as InnerFilter<AudioEvent>;
 
-    const importFileFilters = {
-      audioEventImportFileId: {
-        in: Array.from(this.importFiles),
-      },
-    };
+      return filterAnd(initialFilter, importFilters);
+    }
```
atruskie: So given import A with file A1 and import B with file B1, 

If i filter for import A and file A1 and import B (but not file B1!) then i will only see the results of file A1. This is a bug, no?

I'd recommend restructuring this filter to end up with something like:

```
or: [
    {audioEventImportFileId: {eq: A1 },
    {audioEventImport.id: {eq: B },
]
```

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
```diff
@@ -209,6 +219,22 @@ export class AnnotationSearchParameters
...
+    // and annotation import filters upon parameter model construction, we can
+    // ensure that the query string parameters are corrected, meaning that the
+    // user won't be fooled into thinking that their filters are being applied
+    // when they are not.
+    if (hasImportFileFilters && !hasImportFilters) {
+      console.warn(
+        "Annotation import file filters are expected to be a subset of " +
+          "event import filters. However, no event import filters were provided.",
+      );
+      queryStringParameters.importFiles = null;
```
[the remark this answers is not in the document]
hudson-newey: I purposely decided on a soft-warn here because I didn't want to hard fail on something that could be fixed.

Additionally, the only way a user should be able to get into this position is by manually editing the URL, so I hope that it'll be noticeable that the URL parameter was removed because they would have just edited it.

Obviously, this is something that @atruskie might have an opinion on.

src/app/components/shared/datetime-formats/abstract-template.component.ts
```diff
@@ -19,14 +19,13 @@ export abstract class AbstractTemplateComponent<
...
 {
-  public constructor() {}
-
   // we use a transform function on the [input] prop so that we can change the shape of the input
   // to a uniform type. This allows us to take multiple types as input and convert them into a single type
   // eg. the [input] can take a Luxon DateTime, JS Date, or string and convert this.value into a Luxon DateTime
   public readonly value = input.required<NormalizedType, InputType>({
     transform: (newValue) => this.normalizeValue(newValue),
   });
+  public readonly withTooltip = input(true);
```
atruskie, line 38: what is this for?

src/app/services/baw-api/audio-event-import-file/audio-event-import-file.service.ts
```diff
@@ -94,20 +102,89 @@ export class AudioEventImportFileService
...
+  ): Observable<AudioEventImportFile> {
+    throw new Error("Method not implemented.");
+  }
+
+  public destroy(
+    _model: IdOr<AudioEventImportFile>,
+  ): Observable<void | AudioEventImportFile> {
+    throw new Error("Method not implemented.");
+  }
+
```
atruskie: how about we just don't implement StandardAPI if we're not going to use it.

even if we do shallow we're probably not going to impl all the crud methods, just filter

atruskie (changes requested):
I see a verify button on the import list page, what about the import details page?

that would probably be a button in the secondary menu

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: We don't currently have a details page for the annotation import files.

We can easily add a download link?

Or do you think that a rich details page (like the audio recording list page) is helpful?

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.scss
hudson-newey: Name proposal: `ngb-tooltip.data-tooltip`

src/app/components/shared/datetime-formats/abstract-template.component.ts
hudson-newey: I'll add a JSDoc comment

At the moment, the date/time components have additional information that is shown on hover. E.g. timezone information, the ISO8601 format, etc...

However, sometimes we don't want "rich" information.
E.g. for the import file tooltips, we don't want to nest tooltips.

Additionally, this disables the `.has-context` styling that adds an underline to the `<time>` element (because there's no longer any context/tooltip)

src/app/services/baw-api/audio-event-import-file/audio-event-import-file.service.ts
hudson-newey: > even if we do shallow we're probably not going to impl all the crud methods, just filter

I'm sold; sounds good

The client already has a concept of an filter-only endpoint which will make this quite easy

hudson-newey: image

src/app/components/shared/datetime-formats/abstract-template.component.ts
hudson-newey: image

I'll keep this comment "unresolved" in case you want to object to this input

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
hudson-newey: > If i filter for import A and file A1 and import B (but not file B1!) then i will only see the results of file A1

This is how I envisioned the filters to work.

In my mind

- All values within a single input are $$\text{OR}$$'d together
- Conditions between inputs are $$\text{AND}$$'d together

This makes conditions in the form consistent, and without special cases for existing relationships

If you think this is not the expected behavior, we might also want to change the project, region & site filters because this is how they currently work.

https://github.com/QutEcoacoustics/workbench-client/blob/master/src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts#L317-L346

### Project/Region/Site example

For projects P1 and P2
Which have sites P1<-S1 and P2<-S2

If you filter for P1, P2 and P1<-S1

Currently will currently create the filter (P1 $$\lor$$ P2) $$\land$$ P1<-S1

Therefore, we **currently** shake away the (P1 $$\lor$$ P2) condition because it's already covered by the P1<-S1 relationship

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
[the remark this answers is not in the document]
atruskie: I wondered about this too, and eventually realised that this would happen with URL twiddling, and also agreed a soft fail was appropriate

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
atruskie: Oh good point. OK: let's just link to the files tab on the import details page. Bonus points if we can ensure this item is shown in the table.

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.scss
atruskie: I'd prefer something more descriptive, but don't have a good suggestion.

`left-aligned-tooltip-content`?

Maybe `data-tooltip` is sufficient with a descriptive comment

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
atruskie: > In my mind
> All values within a single input are OR'd together
> Conditions between inputs are AND 'd together

This makes sense; however, we're using the `import` filter to hierarchically filter the `file` filter. It should probably behave as one combined input. 

Your counter example is interesting and well reasoned, but not quite valid.

If we were using the `projects` input to scope the `sites` input then I think it should behave like my proposal for the `import`/`import_file` inputs. And it's arguable we should do that.

But the assumption we're making is that site names are descriptive enough by themselves that the input can exist in a standalone fashion.

The difference is we probably only have a few sites per project with similar names, but we could have thousands of identical import files with the same name per project.

atruskie: This reminds me: is import file filtering disabled unless and import has been selected?

src/app/components/shared/datetime-formats/abstract-template.component.ts
atruskie: good comment

src/app/components/annotations/components/annotation-search-form/annotationSearchParameters.ts
hudson-newey: > This reminds me: is import file filtering disabled unless and import has been selected?

Yes, this is how it currently works

It also clears the import file input if you remove all of the imports

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: Discussed offline: Adding a tooltip that stays open on hover is quite hard without some arch changes that are out of scope for this PR

Additionally, we thought of having an "info" template that can be provided to the typeahead input which would show a small "info" icon that could be toggled to show additional info.

Related issue: https://github.com/QutEcoacoustics/workbench-client/issues/2567

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: This is not something that I think we can easily change.

However, I might be able to add a directive to override this width because this is probably something that we want more often.

src/app/components/import-annotations/pages/list/list.component.html
```diff
@@ -44,6 +44,16 @@ <h1>Batch Annotation Import History</h1>
         View
       </a>
 
+      <a
+        name="verify-button"
+        class="btn btn-sm btn-outline-primary ms-1"
+        [strongRoute]="verificationRoute.project"
+        [routeParams]="{ projectId: project.id }"
+        [queryParams]="{ audioEventImports: value.id }"
```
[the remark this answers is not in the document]
hudson-newey: Good catch, I did forget to update this

src/app/components/annotations/components/annotation-search-form/annotation-search-form.component.html
hudson-newey: image

src/app/components/import-annotations/pages/details/details.component.html
```diff
@@ -172,8 +172,7 @@ <h1>Import Annotations: {{ audioEventImport.name }}</h1>
                   [strongRoute]="verificationRoute.project"
                   [routeParams]="{ projectId: project.id }"
                   [queryParams]="{
-                    importFiles: value.id,
-                    audioEventImports: value.audioEventImportId,
+                    imports: value.audioEventImportId + ':' + value.id ,
```
atruskie: This shouldn't be left up to manual construction. Should definitely use a helper function

src/app/components/import-annotations/pages/list/list.component.html
```diff
@@ -49,7 +49,7 @@ <h1>Batch Annotation Import History</h1>
         class="btn btn-sm btn-outline-primary ms-1"
         [strongRoute]="verificationRoute.project"
         [routeParams]="{ projectId: project.id }"
-        [queryParams]="{ audioEventImports: value.id }"
+        [queryParams]="{ imports: value.id + ':' }"
```
atruskie: ditto.

Also, we can probably omit the colon for an empty mapping?

src/app/helpers/query-string-parameters/queryStringParameters.ts
```diff
@@ -4,64 +4,125 @@ import { toNumber } from "@helpers/typing/toNumber";
...
+): SerializationTechnique<T, string> => ({
+  serialize: (value: T): string => {
+    const mapEntries = Array.from(value.entries());
+    if (mapEntries.length === 0) {
+      return null;
+    }
+
+    const entries = mapEntries.map(([keys, values]) => {
+      if (values.size === 0) {
+        return `${keys}:`;
```
atruskie: can we omit the colon for empty mapping?

src/app/helpers/query-string-parameters/queryStringParameters.ts
```diff
@@ -4,64 +4,125 @@ import { toNumber } from "@helpers/typing/toNumber";
...
+      }
+
+      return Array.from(values)
+        .map((fileId) => `${keys}:${fileId}`)
+        .join(",");
+    });
+
+    return entries.join(",");
```
atruskie, line 92: Isn't this an ambiguous grammar? How do you separate the inner list from the outer list?

Is that why we have the colon?

Suggestion, use semi colon for one of the lists. Probably the outer one.

src/app/helpers/query-string-parameters/queryStringParameters.ts
```diff
@@ -4,64 +4,125 @@ import { toNumber } from "@helpers/typing/toNumber";
...
+  },
+  deserialize: (urlString: string): T => {
+    const stagedImports = new Map<Values, Set<Values>>() as T;
+
+    const importPairs = urlString.split(",");
```
atruskie, line 97: yeah I don't understand how this would work in a case such as:

`A:A1,A2,B:B2,B1`

atruskie (changes requested):

src/app/helpers/query-string-parameters/queryStringParameters.ts
hudson-newey: This would be emitted as 

`A:A1,A:A2,B:B2,B:B1`

Each key + value pair is explicitly emitted

This makes it a lot easier to code rather than trying to have lists inside of lists. Instead it's just one flat array

I assume this is why a lot of your other comments are mentioning lists inside of lists

atruskie: Ok, yep, you're right, I understand now.

atruskie (approved):

src/app/helpers/query-string-parameters/queryStringParameters.ts
hudson-newey: > Ok, yep, you're right, I understand now.

Does this effect any of your preivous comments?

atruskie: Only: https://github.com/QutEcoacoustics/workbench-client/pull/2562#discussion_r2596805832

src/app/helpers/query-string-parameters/queryStringParameters.ts
hudson-newey: Yes (I already support omitting it at the moment, it's just not generated like that)

src/app/components/import-annotations/pages/details/details.component.html
hudson-newey: The "correct" solution here would be to use the `jsMap` qsp serializer

image

Note this implementation is a smell because I'm calling a function in a template.

Adding it to the view model would be correct, but would also require a large refactor so that the `baw-datatable` can construct "virtual" fields. I'll opt to not do this refactor at the moment, and just do a hacky call to a function in the template

hudson-newey: image

hudson-newey:
@atruskie Let me know if you have any feedback on changes

https://github.com/QutEcoacoustics/workbench-client/pull/2562/commits/eb038971bdf66c88841bc9c01d3107a4cae43c26

src/app/components/import-annotations/pages/details/details.component.html
atruskie: I mean the model could return a well constructed url? But sure this will do

PR #2696, opened 2026-09-01
chore: take ng-http-caching 22.2.0

nigrosimone:
Hi, I am the author of ng-http-caching. Every now and then I look at the projects that depend on my libraries, to see how they are used, and this is how I found this one.

The comment in `ngHttpCachingConfig.ts` about `@bawReadonlyConvertCase` getting a "status is read-only" error on cached objects was my fault: up to version 21 the library froze the whole cached response, the `HttpResponse` and its `HttpHeaders` included, so `status`, `ok` and `url` became read only, and even reading a header from a cached response could throw. Version 22 freezes only the body, and 22.2.0 adds a `responseSerializer` option that gives every cache hit its own copy of the body, if you ever need to modify one in place.

## Changes

- bump `ng-http-caching` from 20.0.1 to 22.2.0
- the same comment said the lifetime was "temporarily set to 0 seconds", but it is 10 seconds since #2171, so I rewrote it

## Problems

The 22.x public API is a superset of the 20.x one, and the new `clearCacheOnMutation` option defaults to no invalidation, so the behaviour should not change. I could not run `npm run test:all` here, so please let the CI check it.

Not fixed here: `buildCachingOptions` spreads the whole config into `withNgHttpCachingContext`, but the context only honours `getKey`, `isCacheable`, `isExpired`, `isValid`, `clearCacheOnMutation` and `responseSerializer`. So a per request `cacheOptions: { lifetime: ... }` is ignored today.

## Issues

None open for this. Tell me if you want me to create one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
         "just-camel-case": "^6.0.1",
         "just-snake-case": "^3.0.1",
         "luxon": "^3.4.4",
-        "ng-http-caching": "^20.0.1",
+        "ng-http-caching": "^22.2.0",
         "ngx-captcha": "^14.0.0",
         "ngx-clipboard": "^16.0.0",
         "ngx-cookie-service": "^20.1.0",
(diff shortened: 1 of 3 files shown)


atruskie:
Wow, thanks @nigrosimone, much appreciated!

Looks great. I'm going to look further into the problems you mentioned tomorrow and I'll merge this soon after.

PR #2710, opened 2026-09-16
feat: serve the SSR app on fulmine.js, with pre-compressed assets

nigrosimone:
Hi again. After #2696 I looked at how the SSR server is set up, and I think this project can gain from another library of mine, [fulmine.js](https://github.com/nigrosimone/fulmine.js): a drop-in replacement of Express 5 that runs on uWebSockets.js. Same API, same middleware, the Express 5 test suite passes whole, and `@angular/ssr`'s `AngularNodeAppEngine` works on it unchanged.

## Changes

- `src/server.ts`: `import express from "fulmine.js"`, and the catch-all becomes `/{*splat}`, the Express 5 syntax (`/**` is Express 4). So this is also the upgrade from Express 4 to the Express 5 API. If anything goes wrong, going back to plain Express 5 is that one import line.
- `src/server.ts`: `express.compression()` in front, and `express.static(..., { preCompressed: true })`, which serves a `.br` or `.gz` twin when the client accepts it. `scripts/precompress.mjs` writes the twins after `ng build` (`postbuild`): the browser bundle goes from 12.2 MB to 1.9 MB brotli. Today www.ecosounds.org sends `main.js` uncompressed, 3.8 MB, even to a client that asks for `br, gzip`; with this it is 0.67 MB and nothing is compressed at request time.
- `angular.json`: `fulmine.js` and `uWebSockets.js` declared external, so esbuild does not try to bundle the native binary (`npx fulmine.js angular` writes this).
- `Dockerfile`: uWebSockets.js ships a glibc binary, so Alpine cannot run it. Build stage `node:24-trixie` (it has the git that npm needs to fetch uWebSockets.js), runtime `node:24-trixie-slim`, and a small stage that installs only fulmine.js for the runtime image, since it is no longer inside the server bundle. I built and ran the image locally.
- `package.json`: `express` removed (it stays as a transitive dependency of webpack-dev-server), `@types/express` to 5. The lock diff is bigger than that because npm deduplicated some chokidar copies while it was there.

## Measured on this project

Production build, Linux x64, node 24, all servers up at the same time and the load alternated between them, 4 rounds, median. The same run with Express against itself reads 0.95 to 1.02, so that is the noise.

Static files, with `Accept-Encoding: br, gzip` like a browser:

| Request                | Express 4 as deployed | Express 4 + `compression` | this PR                          |
| ---------------------- | --------------------- | ------------------------- | -------------------------------- |
| `main.js` (3.7 MB)     | 133 req/s, 3741 KB    | 26 req/s, 856 KB          | 4014 req/s, 666 KB (30x)         |
| a 940 KB chunk         | 441 req/s, 941 KB     | 110 req/s, 266 KB         | 8934 req/s, 219 KB (20x)         |
| `favicon.ico`          | 2549 req/s            | 1863 req/s                | 4695 req/s (1.8x)                |

The framework alone, no compression on either side:

| Request                                 | Express 4  | fulmine.js  | ratio |
| --------------------------------------- | ---------- | ----------- | ----- |
| `/assets/environment.json` (1.6 KB)     | 7365 req/s | 25745 req/s | 3.50x |
| `/favicon.ico` (5 KB, `express.static`) | 2493 req/s | 13547 req/s | 5.43x |
| `main.js` (3.7 MB, `express.static`)    | 115 req/s  | 180 req/s   | 1.57x |
| SSR page `/`, uncached, one at a time   | 154 ms     | 158 ms      | same  |

The render is the same JavaScript on both and it waits on the API, so a cache miss does not change. Everything around it does, and for an SSR site most requests are the bundle and the assets. If you want the pages cached too, [ng-ssr-caching](https://www.npmjs.com/package/ng-ssr-caching) is one `server.use()` in front of the Angular handler and works on both: `/` served 2743 req/s on Express and 3299 req/s on fulmine with it. Not in this PR, since it needs a decision about cookies and logged-in users.

Public numbers, on other people's hardware: [HttpArena](https://www.http-arena.com/#sort=rps:-1&q=Js) and [web-frameworks](https://web-frameworks-benchmark.netlify.app/result?l=javascript). And if one day the workbench needs live data, `app.ws()` gives native uWebSockets on the same port, with an `upgrade(req, res)` hook for auth.


atruskie:
Hello again, thanks again!

I'll have to consider this carefully. Frankly there are multiple bugs with our SSR implementation, and we've been considering just ditching it completely. So, widening our dependencies for something we don't like that much wouldn't be something we normally consider

Assets _should_ be compressed by the load balancer as well. That's not working, so at the very least, thanks for the heads up.

nigrosimone:
Understood, thanks for looking at it. If SSR goes away this PR is moot, so feel free to close it. The compression part stands on its own: if the load balancer starts compressing, the numbers in the PR are what you get for free. No pressure from my side.

